PR #7, opened 2023-08-16
Start breaking parser into its own pacakge

siefkenj:
This is based off of #6 . It adds a `packages/parser` package and has the start of a conversion to typescript.


dqnykamp (changes requested):
I haven't been able to trace the source of this error, but there are now errors with the integration between the parser and CodeMirror.

Steps to reproduce this error.
1. Create this DoenetML in testCode.doenet: `<codeEditor>`
2. In the test viewer, type `<` into the code editor that appears
3. The console gives the error: `Uncaught TypeError: cursor.childAfter is not a function`
4. Autocomplete no longer functions and the cursor appears to stay at the front while typing.
5. You can still type though, and write the DoenetML `<text>hello`
6. Typing `<` triggers the same error, but folliowing with `/` does invoke the autocomplete to select `</text>`.
7. However, selecting `</text>` from autocomplete does not work and gives the error: `Uncaught TypeError: cursor.childBefore is not a function`

siefkenj:
How do I do that? Do I need to add a `<CodeEditor />` widget to the `test-viewer`?

siefkenj:
Okay. I'm pretty sure this is due to the fact that I updated the `lezer` dependencies and now they're out of sync with the CodeMirror ones. It seems that somewhere along the line `parent.cursor` changed to `parent.cursor()` but the old codemirror doesn't recognize that.

siefkenj:
I have updated CodeMirror and the error goes away, but I don't see any syntax highlighting after the update, so I'm guessing a little more work needs to be done :-/.

dqnykamp:
> How do I do that? Do I need to add a `<CodeEditor />` widget to the `test-viewer`?

You add `<CodeEditor>` to the testCode.doenet file that gets read into the test viewer, as I assume you've figured out by now.  You can also add `<CodeEditor showResults />` to get an attached `<CodeViewer>` (though the styling is off right now -- it looks better with `<CodeEditor showResults resultsLocation="left" />`).

dqnykamp:
We also lost the line numbering and the styling showing which line is active.

siefkenj:
Superseded by #10

PR #8, opened 2023-08-17
Move `Core/` source code into its workspace

siefkenj:
Previously, the `doenetml-worker` workspace linked to code in the `doenetml` workspace. This is sloppy and leaves the code linked. This PR is a start at moving that code. It depends on #7 

Currently it is broken. I've chased dependencies for a while and they just keep popping up...


siefkenj:
Duane is redoing this processes, so this PR is no longer needed.

PR #24, opened 2023-09-03
DAST and XML Compatability

siefkenj:
This PR 
 * updates the Lezer grammar to support all XML features, so valid XML documents should parse as DoenetML documents.
 * Introduces DAST (DoenetML Abstract Syntax Tree) as a format for interacting with DoenetML source code. The old `parser.ts` is still around, but it should be reworked to use DAST.
   - DAST has a `type: "error"` node that contains information about parse errors.
   - the `toXml` function can output proper XML or DoenetML-style markup (e.g., with JSX-style properties). It can also output explicit `<_error>` nodes where errors were encountered, or it can silently omit them.
 * A basic pretty-printer based off of Prettier. Lots more work needs to be done on the pretty printer to make sure it satisfies all the edge cases.
 * Add a `showFormatter` attribute to `CodeEditor` that will allow for pretty printing of code and for converting between XML and DoenetML


packages/parser/test/fragments/README.md
```diff
@@ -0,0 +1,3 @@
+Fragments of DoenetML code to test the parser.
+
+Fragments `xml-*` come from the `lezer-xml` test cases: https://github.com/lezer-parser/xml/tree/main MIT license
```
dqnykamp, line 3: This comment seems incorrect since xml-doenet-* files don't appear to be standard xml.

siefkenj: Good call. Only the `xml-##` ones are from `lezer-xml`

packages/parser/src/lezer-to-dast/lezer-to-dast.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (source == null) {
+        throw new Error(
+            `If you provide a SyntaxNode, you must also provide the source string`,
+        );
+    }
+    return _lezerToDast(node, source);
+}
+
+export function _lezerToDast(node: SyntaxNode, source: string): DastRoot {
+    if (typeof node === "string") {
```
dqnykamp, line 67: Delete this check since already know node is a SyntaxNode? Certainly true if `_lezerToDast` is not exported outside this file.

packages/parser/src/lezer-to-dast/gobble-function-arguments.ts
```diff
@@ -0,0 +1,223 @@
...
+    DastText,
+} from "../types";
+import { createOffsetToPositionMap } from "./lezer-to-dast-utils";
+
+/**
+ * Functions can take arguments consisting of DastNodes. The macro parser works on strings only,
+ * so functions that take DastElement children will not be fully parsed. We manually parse through
+ * and attach the arguments in this special case.
+ *
+ * **Node**: this function may mutate the input.
```
dqnykamp, line 16: **Note**

packages/parser/src/lezer-to-dast/lezer-to-dast-utils.ts
```diff
@@ -0,0 +1,304 @@
...
+    referenceNode: DastNodes,
+) {
+    if (!nodeToUpdate.position || !referenceNode.position) {
+        return;
+    }
+    const offset = referenceNode.position.start.offset;
+    const line = referenceNode.position.start.line - 1;
+    const column = referenceNode.position.start.column - 1;
+    if (!nodeToUpdate.position.start.offset != null && offset != null) {
+        nodeToUpdate.position.start.offset! += offset;
```
dqnykamp, line 300: Remove the `!` after `offset`.

packages/parser/src/lezer-to-dast/lezer-to-dast-utils.ts
```diff
@@ -0,0 +1,304 @@
...
+        i = entity.to - node.from;
+    }
+    ret += rawValue.slice(i);
+    return ret;
+}
+
+export function textNodeToText(node: SyntaxNode, source: string): string {
+    if (node.name !== "Text") {
+        throw new Error(
+            `Can only get the text of an "AttributeValue" type node, not "${node.name}"`,
```
dqnykamp, line 67: Incorrect error message.

packages/parser/test/dast-advanced.test.ts
```diff
@@ -0,0 +1,1017 @@
...
+                  "type": "function",
+                },
+                {
+                  "type": "text",
+                  "value": " bar",
+                },
+              ]
+            `);
+
+        // Argument with a macro in it
```
dqnykamp, line 139: It looks like the argument has a full component in it, not a macro.

dqnykamp (approved):
+1

PR #34, opened 2023-09-28
Static assets

siefkenj:
Make static-assets into a module for tree-shaking purposes.

Directly imported JSON doesn't get tree-shaken by `vite`, but JS code does. By making `static-assets` re-export everything as a module, tree-shaking should work across packages.


dqnykamp (approved):
+1

PR #44, opened 2023-10-19
Add plugin that externalizes @doenet/* deps

siefkenj:
Hopefully this gets the build sizes down a little bit.


siefkenj:
It actually increases the size, so I am closing the PR.

PR #45, opened 2023-10-20
added sample component docs files

virginiaMae:
also created new docs folder in doenetml package


siefkenj:
Can you repoen this with the previous commits? It was a useful reference...

PR #48, opened 2023-10-29
add properties to schema

dqnykamp:


packages/static-assets/scripts/get-schema.ts
```diff
@@ -229,6 +372,131 @@ export function getSchema() {
...
+    let componentType = description.createComponentOfType;
+
+    let prop: PropertyDescription = {
+        name: varName,
+        type: componentType,
+        isArray: description.isArray,
+    };
+
+    if (description.isArray) {
+        let numDimensions = description.numDimensions || 1;
```
siefkenj, line 393: I suppose `description.numDimensions` can never be zero?

dqnykamp: That's right, it needs to be a positive integer.

PR #62, opened 2023-11-30
Simplified Macro AST

siefkenj:
This PR simplifies the parsing of macros, disallowing `/` and only allowing attributes at the end of a macro. The old macro parsing is kept around for use in the LSP to allow auto-updating of the old macro syntax.


packages/parser/test/macro-parse.test.ts
```diff
@@ -105,6 +107,106 @@ describe("Macro parsing", () => {
...
+            $t.x[1].y
+            $t.x[1][2].y
+            $t.x[1].y[2]
+            $t{a}
+            $t{a="b"}
+            $t{a="$b"}
+            $t[1]{a="b"}
+            $t.x[1]{a="b"}.y
+            $(t)
+            $(x-y)
```
dqnykamp, line 139: Could you also add a test with `$x-y` with no parens to make sure that it gives the macro `$x` followed by "-y"?

dqnykamp: * `$(t)[1]`  (can break before `[` if want to )
* `$t[]`  (so eating empty [] is standardized)

packages/parser/test/macro-parse.test.ts
```diff
@@ -105,6 +107,106 @@ describe("Macro parsing", () => {
...
+    }
+    {
+        const validFunctions = `$$f
+            $$f1(y)
+            $$f[1](y)
+            $$f[$x](y)
+            $$f[a][b](y)
+            $$f($x)
+            $$(x-y.z{t})(m)
+            $$f(x, y)`.split(/\n\s+/g);
```
dqnykamp, line 208: Other cases involving functions
* `$$f`
* `$$f()`

Do you have tests of invalid cases like `$$f(`?

dqnykamp (approved):

PR #66, opened 2023-12-10
Switch CodeMirror to get completions via the language server

siefkenj:
This PR adds language server support to codemirror for autocompletion. This will allow completion of macros in the future. It also uses `react-codemirror` to manage the codemirror interface rather than manually managing it through references.


siefkenj:
@dqnykamp This should be ready to merge :-)

packages/lsp/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "exports": {
+        ".": {
+            "import": "./dist/index.js"
+        },
+        "./language-server.js": {
+            "import": "./dist/index.js"
+        }
+    },
+    "scripts": {
+        "dev": "vite",
```
dqnykamp, line 22: When run this `npm run dev` and load web page, get
```Failed to load url /src/dev-site.tsx```

File: packages/lsp-tools/test/doenet-auto-complete.test.ts [hunk 2]
```diff
@@ -166,6 +166,25 @@ describe("AutoCompleter", () => {
             `);
         }
     });
+    it("Can suggest completions after a `<`", () => {
```
dqnykamp, line 169: This tests makes me think that after typing "<" in the editor, I should get some completions. However I don't get completions after typing just "<" inside a `<codeEditor>` tag (in test-viewer dev server) or inside the editor from the dev server in lsp-tools.

dqnykamp (changes requested):

packages/lsp-tools/test/doenet-auto-complete.test.ts
siefkenj: Really? I need to look into that then...It gives suggestions to me when testing out in `packages/codemirror`.

dqnykamp: Tried it out more.  It turns out it doesn't work if my cursor is at the end of the file, but it works otherwise. I was always testing at the end of the file, so it wasn't working.

dqnykamp: Also, if at end of file and have an unclosed open tag, then `<` does not suggest the closing tag.

siefkenj:
Autocomplete should now work for opening tags at the end of a string. Closing tags are still an issue. I added a `.skip` test for closing tags which can be used when this issue is ready to be tackled again.

dqnykamp (approved):
Remaining autocomplete issues for closing tags are given in issue #71.

PR #68, opened 2023-12-11
Switch worker communication to comlink

siefkenj:
[Comlink](https://github.com/GoogleChromeLabs/comlink) is a library that makes it much easier to communicate with webworkers. This PR builds off of #67


packages/doenetml-prototype/src/doenet-applet.tsx
```diff
@@ -0,0 +1,65 @@
...
+    showHints: true,
+    allowLoadState: false,
+    allowSaveState: false,
+    allowLocalState: false,
+    allowSaveSubmissions: false,
+    allowSaveEvents: false,
+    autoSubmit: false,
+};
+
+export function DoenetApplet({
```
dqnykamp, line 39: I don't think of DoenetML documents as applets, so I'd change the wording. Other possible words are documents or activities, though avoiding any such term is OK. Plus, Doenet and DoenetML are distinct, and I think everything here is DoenetML.

siefkenj: Sure, I can change it back if you like.

My thinking was the following: we have `PageViewer` which renders things and assumes the state, etc. has already been defined. It just uses callbacks. Then we have `DoenetApplet` which is the "root" of a self-contained doenet instance. It might split the source into multiple pages and call several instances of `PageViewer` or it might render everything together. If you have multiple instances of `DoenetApplet`, they cannot communicate.

I felt like `DoenetML` was a little confusing becase
  1. You're not just inserting some doenet code, you're creating an isolated context.
  2. The `ML` part stands for Markup Language; I know you're distinguishing the Doenet server/LMS from the Doenet render/interface, but I think `ML` is a strange way to distinguish this...

packages/doenetml-prototype/src/DoenetML.tsx
```diff
@@ -48,124 +36,30 @@ const defaultFlags: DoenetMLFlags = {
...
-
     return (
-        <MathJaxContext
-            version={2}
-            config={mathjaxConfig}
-            onStartup={(mathJax) => (mathJax.Hub.processSectionDelay = 0)}
-        >
-            <ActivityViewer
-                doenetML={doenetML}
+        <Provider store={store}>
```
dqnykamp, line 57: Where does one configure MathJax now?

siefkenj: I removed it temporarily because the import was pulling in lots of stuff it shouldn't have...

packages/doenetml-prototype/src/state/redux-slices/core/slice.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// Define the initial state using that type
+const initialState: CoreState = {
+    initialized: false,
+    launched: false,
+    workerCacheKey: undefined,
+    inErrorState: false,
+};
+
+const dastSlice = createSlice({
```
dqnykamp, line 31: coreSlice?

packages/doenetml-prototype/src/state/redux-slices/dast/utils/normalize-dast.ts
```diff
@@ -0,0 +1,146 @@
...
+ * Plugin that ensure the root of the document is a single `<document>` element
+ * and that all whitespace is trimmed from the content of `<document>`.
+ */
+const pluginEnsureDocumentElement: Plugin<[], DastRoot, DastRoot> = () => {
+    return (tree) => {
+        const existingDocument = tree.children.find(
+            (n) => n.type === "element" && n.name === "document",
+        ) as DastElement | undefined;
+        if (existingDocument) {
+            existingDocument.children = trimWhitespaceNodes(
```
dqnykamp, line 62: filterNonDoenetXml here too?

siefkenj: Ah, I didn't do that because docstrings are only valid at the root of the document. Since we're throwing away the rest of the root, it should be okay.

dqnykamp (approved):

PR #76, opened 2023-12-15
Use wireit to auto-build dependencies

siefkenj:
[wireit](https://github.com/google/wireit) is a script that handles dependency building and caching, configured in `package.json`. This PR changes the build processes for most packages to use `wireit`.

Results:
  * building should be much faster since building of deps can happen in parallel
  * You can now run `npm run build --watch` in a workspace and it will automatically rebuild all deps when a file changes (no need for multiple watched directories)

The top-level `npm run build` was changed to only build `test-viewer`, which skips the building of `test-cypress` and `standalone`. This should speed things up for most devs. The CI should still build all of the packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ cd packages/test-viewer
 npm run dev
 ```
 
-Paste demo code into `src/test/testCode.doenet` and navigate to the URL specified in the console (probably http://localhost:5173 )
+Paste demo code into `src/test/testCode.doenet` and navigate to the URL specified in the console (probably http://localhost:8012 )
 
 ## Demos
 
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -22,15 +22,21 @@ Semantic markup for building interactive web activities.
...
+To test code, you can use the `projects/test-viewer` package.
 
-Paste demo code into `src/test/testCode.doenet`
+```bash
+cd packages/test-viewer
+npm run dev
+```
 
-Navigate to `localhost:5173`
+Paste demo code into `src/test/testCode.doenet` and navigate to the URL specified in the console (probably http://localhost:5173 )
```
dqnykamp, line 39: test-viewer is configured to use port 8012

dqnykamp (approved):
+1

PR #82, opened 2023-12-21
change worker api

dqnykamp:


packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -15,49 +15,204 @@ export interface FlatDastRoot {
...
+    // so that it can attempt to send final state (to allow it to be saved)
+    // before terminating
+
+    worker.doenetCore?.free();
+    worker.doenetCore = undefined;
+
+    // Since we store source and flags in doenetCore,
+    // we lose source and flags when terminating it.
+    worker.source_set = false;
+    worker.flags_set = false;
```
siefkenj, line 215: This all seems quite excessive...can't you do

```typescript
CoreWorker {
   isProcessingPromise = Promise.resolve();
   async doSomething(){
      const isProcessingPromise = this.isProcesssingPromise;
      let resolve: Function;
      this.isProcessingPromise = new Promise((res) => resolve=res);

      await isProcessingPromise
      // do the stuff ...
      resolve()
   }
}
```

Am I missing something?

packages/doenetml-worker-rust/src/lib.rs
```diff
@@ -14,24 +14,58 @@ static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
 #[wasm_bindgen(getter_with_clone)]
 #[derive(Debug)]
-pub struct PublicDoenetMLCore(DoenetMLCore);
+pub struct PublicDoenetMLCore {
+    core: Option<DoenetMLCore>,
+    dast_json: Option<String>,
+    source: String,
+    flags: Option<String>,
```
siefkenj, line 21: `flags_json`?

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -15,49 +15,204 @@ export interface FlatDastRoot {
     warnings: unknown[];
 }
 
+interface SetSourceInstruction {
+    type: "set_source";
+    source: string;
+    dast: DastRoot;
+    core_worker: CoreWorker;
```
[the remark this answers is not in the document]
dqnykamp: I can never figure out what case is supposed to go where. At least in rust, it complains if I get it wrong!

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -15,49 +15,204 @@ export interface FlatDastRoot {
     warnings: unknown[];
 }
 
+interface SetSourceInstruction {
```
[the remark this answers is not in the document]
dqnykamp: cool!

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -15,49 +15,204 @@ export interface FlatDastRoot {
...
+    reject: Function;
+}
+
+let instruction_queue: (
+    | SetSourceInstruction
+    | SetFlagsInstruction
+    | ReturnDastInstruction
+    | TerminateInstruction
+)[] = [];
+let processing_instruction = false;
```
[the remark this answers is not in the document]
dqnykamp: I was just worried that every member of the CoreWorker class was getting exposed by Comlink, so I didn't want the queuing interface exposed to the DOM thread. But, I can put them into the CoreWorker class.

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -15,49 +15,204 @@ export interface FlatDastRoot {
...
+    setSource(args: { source: string; dast: DastRoot }) {
+        return new Promise((resolve, reject) => {
+            instruction_queue.push({
+                type: "set_source",
+                source: args.source,
+                dast: args.dast,
+                core_worker: this,
+                resolve,
+                reject,
+            } as SetSourceInstruction);
```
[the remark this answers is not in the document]
dqnykamp: Yes, apparently Typescript can figure it out without casting

packages/doenetml-worker-rust/src/CoreWorker.ts
dqnykamp: Interesting. I hadn't thought of that. My understanding of Javascript is that nothing could happen in another call to the worker in between setting `const isProcessingPromise` and setting `this.isProcessingPromise` so there's no race condition here. Is that correct?

siefkenj: Yes. The event loop _should_ be uninterruptible...

dqnykamp: It's complaining that `resolve` is used before it is defined. Any better ideas other than putting a @ts-ignore in front of it?

siefkenj: You could try adding a `!` afterwards to assert it's non-null, or you could initialize it with `()=>{}`

siefkenj:
```typescript
function promiseWithResolver() {
    let resolve: (value: unknown) => void;
    let reject: (reason: unknown) => void;
    const promise = new Promise((res, rej) => {
        resolve = res;
        reject = rej;
    });
    return { promise, resolve: resolve!, reject: reject! };
}
```

dqnykamp: Handy util. However, when I try this:
```
  const isProcessingPromise = this.isProcessingPromise;

  let { promise, resolve } = promiseWithResolver();
  this.isProcessingPromise = promise;
```
I get this error on the last line:
```
Type 'Promise<unknown>' is not assignable to type 'Promise<void>'.
```
Note: in one case, for `return_dast`, the promise is supposed to return something. Use generics or just skip the utility function?

siefkenj: Generics would be fine.
```typescript
function promiseWithResolver<T = unknown>){
  let resolve: (value: T) => void;
  ...
}
```
or something like that :-)

siefkenj: Oh, you might also need to type `const promise: Promise<T> = new Promise(...)`

dqnykamp: Got my promises mixed up with how I was using it for the other way. This promise never returns anything, so I just changed the utility function to
```
function promiseWithResolver() {
    let resolve: (value: void) => void;
    let reject: (reason: unknown) => void;
    const promise: Promise<void> = new Promise((res, rej) => {
        resolve = res;
        reject = rej;
    });
    return { promise, resolve: resolve!, reject: reject! };
}
```

packages/doenetml-worker-rust/src/lib.rs
dqnykamp: why not

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -55,46 +54,47 @@ export const coreThunks = {
...
             }
         },
     ),
+
     /**
      * Launch the DoenetCore webworker with the given source string and DAST tree.
      */
-    launchCore: createLoggingAsyncThunk(
-        "core/launchCore",
+    getDast: createLoggingAsyncThunk(
```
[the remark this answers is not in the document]
dqnykamp: I feel like we'll have a different function with updates that will update/refresh the dast with just the new pieces. This returns the entire dast. What about calling it returnDast like it is on the worker side?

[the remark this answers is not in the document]
dqnykamp: Or `retrieveDast`?

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -55,46 +54,47 @@ export const coreThunks = {
...
+        "core/getDast",
         async (_, { dispatch, getState }) => {
-            const { initialized } = getState().core;
-            if (!initialized) {
-                await dispatch(coreThunks.initialize());
+            const { flags } = selfSelector(getState());
+
+            if (!flags) {
```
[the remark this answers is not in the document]
dqnykamp: Yeah, I was wondering in how many places we define default flag. But, I could import them from `DoenetML.tsx`

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -55,46 +54,47 @@ export const coreThunks = {
...
 
-            const flatDast = await worker.createCore({});
-            dispatch(_dastReducerActions._setFlatDastRoot(flatDast));
-            console.log("flatDast", flatDast);
-            console.log("assembledDast", assembleFlatDast(flatDast));
+            try {
+                const flatDast = await worker.returnDast();
+                dispatch(_dastReducerActions._setFlatDastRoot(flatDast));
+                console.log("flatDast", flatDast);
+                console.log("assembledDast", assembleFlatDast(flatDast));
```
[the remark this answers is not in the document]
dqnykamp: Right!

PR #86, opened 2024-01-04
Implement state variables in rust

dqnykamp:


packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -92,6 +92,50 @@ export class CoreWorker {
...
+            // TODO: Do we need to cast flat_dast_element_updates into a TypeScript type
+            // like we did for flat_dast, above?
+
+            let flat_dast_element_updates = JSON.parse(
+                this.doenetCore.handle_action(args),
+            );
+            resolve();
+            return flat_dast_element_updates;
+        } catch (err) {
+            resolve();
```
siefkenj, line 132: Do you want to put `resolve()` in a `finally { ... }` instead? It gets called even if you return inside the `try`.

packages/doenetml-worker-rust/doenetml-core/src/parse_json.rs
```diff
@@ -0,0 +1,76 @@
...
+        component_idx,
+        action_name,
+        args,
+    })
+}
+
+impl From<serde_json::Number> for StateVarValue {
+    /// A json number will be converted to an integer if it is an integer.
+    ///
+    /// For this reason, it's important that we can convert StateVarValue::Integer to f64,
```
siefkenj, line 51: Are there cases where we'd want to use `BigInt` instead?

packages/doenetml-worker-rust/src/CoreWorker.ts
dqnykamp: Sounds good.

packages/doenetml-worker-rust/doenetml-core/src/parse_json.rs
dqnykamp: I haven't needed a BigInt yet, but I could imagine eventually folks will want it. I'm trying to build a structure where we can easily add additional data types.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+                                            component_state: new_component_state,
+                                            instruction_idx: 0,
+                                            val_idx: 0,
+                                        },
+                                    ));
+
+                                    // break out of the inner loop and continue to outer loop,
+                                    // which will result in first popping off the last instruction
+                                    // that we added to the stack
+                                    continue 'stack_loop;
```
siefkenj, line 449: I looks like you're doing a depth-first search for stale deps. If you did a breadth-first search could you avoid needing to break out of this double loop?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+                    }
+
+                    initial_val_idx = 0;
+                }
+
+                // If we have gotten to this part,
+                // then we have freshened all dependencies
+                // for the state variable (component_idx, state_var_idx).
+                // We calculate its value if a dependency has changed,
+                // else restore its previous value.
```
siefkenj, line 465: I must have missed it. Where is the actual calculation to "freshen" a var done?

Oh, I see, it's only the _dependencies_ which we can assume have been freshend at this point. The real freshening is below.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+
+                let component_idx = component_state.component_idx;
+                let state_var_idx = component_state.state_var_idx;
```
siefkenj, line 468: You're saving these vars to use exactly once?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+
+                let mut comp = components[component_idx].borrow_mut();
+                let state_var = &mut comp.get_state_variables_mut()[state_var_idx];
+
+                // Check if any dependency has changed since we last called record_all_dependencies_viewed.
+                if state_var.check_if_any_dependency_changed_since_last_viewed() {
```
siefkenj, line 474: This seems to be some magic that we haven't talked about yet?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+            instruction_idx: 0,
+            val_idx: 0,
+        }),
+    };
+
+    freshen_stack.push(initial_calculation_state);
+
+    'stack_loop: while let Some(calculation_state) = freshen_stack.pop() {
+        match calculation_state {
+            StateVarCalculationState::Unresolved(unresolved_state) => {
```
siefkenj, line 210: I've been trying to read this branch, but it confuses me...Could you instead just create the dep, set it as stale, and repush it to the stack and let the other branch handle it?

packages/doenetml-worker-rust/doenetml-core/src/dependency.rs
```diff
@@ -0,0 +1,379 @@
...
+    essential_state::{
+        create_essential_data_for, EssentialDataOrigin, EssentialStateVar, InitialEssentialData,
+    },
+    state::{StateVarName, StateVarReadOnlyView, StateVarValue},
+    ComponentChild, ComponentIdx, ExtendSource, StateVarIdx,
+};
+
+/// A DependencyInstruction is used to make a Dependency based on the input document structure
+#[derive(Debug)]
+pub enum DependencyInstruction {
```
siefkenj, line 14: `DependencyRequest` seems like a better name.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+                    state_var.restore_previous_value();
+                }
+            }
+        }
+    }
+}
+
+/// Freshen the state variable specified by original_component_state,
+/// then get its fresh value
+pub fn get_state_var_value(
```
siefkenj, line 487: Since this file is long, I think external functions should be moved to their own file.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+        .get_fresh_value()
+}
+
+/// Recurse in the inverse direction along the dependency graph to attempt to satisfy
+/// the requested update of the state variable described in initial_update_request.
+/// The requested value must already be set on that state variable itself.
+///
+/// When we reach the leaves (essential state variables), set them to their requested values
+/// and then mark all their dependencies as stale
+pub fn process_state_variable_update_request(
```
siefkenj, line 520: Put in its own file?

packages/doenetml-worker-rust/doenetml-core/src/component/text_input.rs
```diff
@@ -0,0 +1,516 @@
...
+                    "immediateValue".to_string(),
+                    StateVarValue::String(immediate_value),
+                )])),
+            })
+        } else {
+            None
+        }
+    }
+
+    fn get_action_names(&self) -> Vec<String> {
```
siefkenj, line 102: This is used for type checking, I assume?

I'm still no sure the best way to do this, but if you set up an enum and use

```rust
use strum_macros::{EnumString, EnumVariantNames};
use strum::VariantNames;

#[derive(EnumString, EnumVariantNames)]
#[strum("camelCase")]
enum Actions {
  UpdateImmediateValue,
  UpdateValue
}
```

then you can do `Actions::VARIANTS` to get a `'static` vec of the string names.

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -82,13 +82,55 @@ export class CoreWorker {
...
+    //     args: HashMap<String, ArgValue>,
+    // }
+
+    // enum ArgValue {
+    //     Bool(bool),
+    //     Number(serde_json::Number),
+    //     NumberArray(Vec<serde_json::Number>),
+    //     String(String),
+    // }
+    async handleAction(args: any) {
```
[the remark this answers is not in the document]
dqnykamp: dispatchAction makes sense

packages/doenetml-worker-rust/doenetml-core/src/utils.rs
```diff
@@ -1,7 +1,26 @@
...
+    /// Return the original key-value pair the the HashMap.
+    fn get_key_value_ignore_case<'a>(&'a self, key: &str) -> Option<(&'a K, &'a V)> {
+        let lowercase_to_original: HashMap<String, &K> = self
+            .keys()
+            .map(|k| (k.to_string().to_lowercase(), k))
+            .collect();
+
+        lowercase_to_original
+            .get(&key.to_string().to_lowercase())
+            .and_then(|k| self.get_key_value(k))
```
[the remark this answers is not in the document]
dqnykamp: Yeah, it looks that way. It doesn't cache the hashmap. Maybe I'll just do a linear search. If it turns out we use it in some place where speed matters and where there are many keys, we can revisit.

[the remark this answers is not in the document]
dqnykamp: Hah, it's doing a linear search anyway, so there doesn't seem to be a point in making a hashmap unless one where caching it.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+        Freshness::Stale => StateVarCalculationState::Stale(StaleCalculationState {
+            component_state: original_component_state.clone(),
+            instruction_idx: 0,
+            val_idx: 0,
+        }),
+    };
+
+    freshen_stack.push(initial_calculation_state);
+
+    'stack_loop: while let Some(calculation_state) = freshen_stack.pop() {
```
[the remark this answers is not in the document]
dqnykamp: Since we do processing after the recursion, I don't think we can do any tail call optimization to avoid stack allocation, at least for this function.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+    essential_state::{EssentialDataOrigin, EssentialStateDescription, EssentialStateVar},
+    state::{Freshness, StateVarValue},
+    ComponentChild, ComponentIdx, ComponentStateDescription, ExtendSource,
+};
+
+/// Internal structure used to track changes
+#[derive(Debug, Clone)]
+pub enum StateVariableUpdateRequest {
+    SetEssentialValue(EssentialStateDescription),
+    SetStateVar(ComponentStateDescription),
```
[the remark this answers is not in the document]
dqnykamp: The term `Essential` is intended to convey that these are the only state variables needed to reconstruct the document state (as every other state variable depends on them). When we write state to a database, we need only save the essential state. I think if an `EssentialStateVar` is not referenced, then it isn't even created. (I'd have to review the code more carefully to be certain of this last claim.)

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+}
+
+/// Used to save the state in the middle of a calculation while freshening a state variable.
+/// Placed on a stack in order to implement recursion using a stack on the heap.
+///
+/// (When created documents so that had to recurse thousands of times,
```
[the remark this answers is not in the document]
dqnykamp: Just deleting this comment, as explained more thoroughly in `freshen_state_var`.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+    let mut stale_renderer_idx = 0;
+    while stale_renderer_idx < stale_renderers.len() {
+        let component_idx = stale_renderers[stale_renderer_idx];
+        stale_renderer_idx += 1;
+
+        stale_renderers.extend(get_non_string_rendered_children_including_from_extend(
+            component_idx,
+            components,
+        ));
+    }
```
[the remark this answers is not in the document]
dqnykamp: `get_non_string_rendered_children_including_from_extend` does not recurse to its children, it only recurses to its `extend_source`, i.e., it doesn't get its rendered descendants, but only accounts for the fact that it inherits any children from its `extend_source.`

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+                essential_data,
+                freshen_stack,
+                should_initialize_essential_data,
+            );
+        }
+    }
+
+    let components_freshened = stale_renderers.clone();
+
+    stale_renderers.clear();
```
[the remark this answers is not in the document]
dqnykamp: The only reason for keeping `stale_renderers` around and just clearing it is to keep the space allocated for the next time this function is called. Since this function is going to be called repeated, I was trying to optimize the number of allocations needed. (Though, unlike the `freshen_stack`, each entry is only an integer.)

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+    // Is there a way to specify the behavior in one place so they cannot diverge and cause bugs?
+    // Perhaps, components cannot override the whole .to_flat_dast method, but just pieces
+    // such as determining the data sent.
+    // And, if they override the behavior is should be through methods called in both places,
+    // like the .get_rendered_children method.
+
+    let component = components[component_idx].borrow_mut();
+
+    let mut children = if let Some(&ExtendSource::Component(source_idx)) = component.get_extend() {
+        get_non_string_rendered_children_including_from_extend(source_idx, components)
```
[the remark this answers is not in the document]
dqnykamp: Right, because it only recurses once, to the `extend_source`. This does not recurse to all descendants or all dependencies.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+    } else {
+        Vec::new()
+    };
+
+    let mut children2 = component
```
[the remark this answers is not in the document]
dqnykamp: Right, `children.extend` and avoiding the `collect()` makes much more sense.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+    // using a stack on the heap.
+    // This approach was chosen because the function recursion implementation would overflow
+    // the small WASM stack once it recursed a few thousands of times.
+    // An alternative approach would be to increase the size of the WASM stack.
+    // This approach was chosen because it is relatively easy to produce documents
+    // with thousands of levels in the dependency graph, and it wasn't clear what
+    // size WASM stack would be appropriate.
+
+    let current_freshness = components[original_component_state.component_idx]
+        .borrow()
```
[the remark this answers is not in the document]
dqnykamp: We could do that. Looking through the code, this seems to be the only place where we want to find the freshness by state variable index, so I'm not sure that it is worth it (given that it would involve the derive macro to pipe it through the enum).

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Yes, I theoretically could put it back on the stack as stale and let the other branch handle it. But that would only save me the last two lines (given that the dependencies are fresh at this point):

```
state_var.calculate_state_var_from_dependencies();
state_var.record_all_dependencies_viewed();
```

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,762 @@
...
+                for (instruction_idx, deps) in dependencies_for_state_var
+                    .iter()
+                    .enumerate()
+                    .skip(stale_state.instruction_idx)
+                {
+                    for (val_idx, dep) in deps.iter().enumerate().skip(initial_val_idx) {
+                        if let DependencySource::StateVar {
+                            component_idx,
+                            state_var_idx,
+                        } = &dep.source
```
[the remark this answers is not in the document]
dqnykamp: That's correct. Given that we are stale (not Unresolved), we know the essential variable has been created and there is nothing to do.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Right, the freshening is done by `calculate_state_var_from_dependencies`. Maybe I should rename that to `calculate_state_var_from_dependencies_and_mark_fresh`?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Once these state variables have been freshened, they are cached and can be used multiple times. This code will need to be run again for a state variable only if one of its dependencies has changed, marking it stale.

The intent of this algorithm is to calculate the state variables (and dependencies) lazily, and once they are need, to memoize the state variable values until they change.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: I need to do a depth-first search because I can calculate/freshen a variable only after all its dependencies (the subgraph below it) have been freshened.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Each state variable has a counter that is incremented on `.set_value()`. The call to `state_var.record_all_dependencies_viewed()` records the values of the counters of the dependency state variables (including essential data). The call to `state_var.check_if_any_dependency_changed_since_last_viewed()` simply checks to see if any of the state variable counters has been incremented since the last call to `state_var.record_all_dependencies_viewed()`.

packages/doenetml-worker-rust/doenetml-core/src/dependency.rs
dqnykamp: Except that I've been using "request" to mean requested values for state variables, as in `DependencyUpdatesRequested` and `StateVariableUpdateRequest`. (I think I'll change the former to `DependencyUpdateRequest` for consistency.)

At the moment, I'm not able to come up with a better name that wouldn't be confused with `DependencyUpdateRequest`, but I'm open to suggestions!

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Separated the state variable calculates and the state variable updates into two files.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Moved these to `state_var_updates.rs`

packages/doenetml-worker-rust/doenetml-core/src/lib.rs
```diff
@@ -1,63 +1,138 @@
...
 pub mod utils;
 
 use crate::utils::{log, log_debug, log_json};
 
-#[derive(Debug, Clone)]
-pub struct ComponentState {
-    pub component_ind: ComponentIdx,
-    pub state_var_ind: StateVarIdx,
+#[derive(Debug, Clone, Copy)]
+pub struct ComponentStateDescription {
```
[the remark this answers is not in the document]
dqnykamp: I like `GlobalStateVarPointer`.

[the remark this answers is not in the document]
dqnykamp: Ended up going with `StateVarPointer` for brevity.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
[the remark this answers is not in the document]
siefkenj: If you clear this, I think it should be done at the call point rather than in this function, and it should be given a name that makes it clear that it is a scratch vector. Though, it sounds like a premature optimization to me...

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
siefkenj: I don't understand...If you add each depth-1 state far to the list, next it will pick the last depth-1 state var and all the depth-2 state var deps to the list, etc. All deps should be handled in the correct order.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
siefkenj: No, what I mean is you're aliasing `component_state.component_idx` but you only use the alias one time.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
siefkenj: Perhaps some of this could be refactored into another function, then? The code is quite hard to follow...

packages/doenetml-worker-rust/doenetml-core/src/dependency.rs
siefkenj: I don't think that is such a bad conflict. You have a `DepGetRequest` and `DepSetRequest`. Seems like a common distinction.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
siefkenj: Can these counters roll over? Or are they u64s where that practically shouldn't be an issue...?

packages/doenetml-worker-rust/src/CoreWorker.ts
```diff
@@ -82,13 +92,43 @@ export class CoreWorker {
...
             return flat_dast;
         } catch (err) {
+            console.error(err);
+            throw err;
+        } finally {
             resolve();
+        }
+    }
+
+    // TODO: TypeScript for args. Here's is the rust type for the args
```
siefkenj, line 104: Unneeded comment.

packages/doenetml-worker-rust/doenetml-core/src/utils.rs
```diff
@@ -1,7 +1,27 @@
...
+where
+    K: ToString + std::cmp::Eq + std::hash::Hash,
+{
+    /// Match key to the HashMap keys, ignoring case.
+    /// Return the original key-value pair the HashMap.
+    fn get_key_value_ignore_case<'a>(&'a self, key: &str) -> Option<(&'a K, &'a V)> {
+        let lower_case_key = key.to_lowercase();
+
+        for k in self.keys() {
+            if k.to_string().to_lowercase() == lower_case_key {
```
siefkenj, line 17: You could even use [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) for slight efficiency gains.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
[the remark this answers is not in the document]
dqnykamp: Actually, I had it wrong how this works. The `DoenetMLCore.stale_renderers` is pushed to whenever a rendered state variable is marked stale. So it needs to be cleared at the end to indicate that there are no more stale renderers. (It's not just a temporary data structure like the stacks, like I erroneously implied.)

And, by looking at this I found a flaw where `stale_renderers` gets pushed to whenever any component has a `for_renderer` state variable marked stale. Instead it should be pushed to only when a **rendered** component has a `for_renderer` state variable marked stale. Will see if this is an easy fix or an issue to be created for later. I don't yet have a quick way to tell if a component is rendered.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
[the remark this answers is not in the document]
dqnykamp: Changing to a match

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: I used u32s. I figure 4 billion updates on a single state variable would be unlikely. But, I could change the > comparison to a != comparison just in case it rolls over.

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
```diff
@@ -17,4 +32,264 @@ pub struct Text {
...
+
+    pub rendered_state_variable_indices: Vec<usize>,
+
+    pub public_state_variable_indices: Vec<usize>,
+
+    pub state_variable_name_to_index: HashMap<String, usize>,
+
+    pub value_state_var_view: StateVarReadOnlyViewTyped<String>,
+
+    pub component_profile_state_variables: Vec<ComponentProfileStateVariable>,
```
siefkenj, line 46: There seems to be a lot of boiler plate here. Can a generic struct be created to house all the non-text-specific stuff?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+            &dependencies[state_var_ptr.component_idx][state_var_ptr.state_var_idx];
+
+        let mut found_stale_or_resolved_dependency = false;
+
+        for deps in dependencies_for_state_var.iter() {
+            for dep in deps.iter() {
+                match &dep.source {
+                    DependencySource::StateVar {
+                        component_idx,
+                        state_var_idx,
```
siefkenj, line 226: Since you're only using these to create a new pointer, you could implement `into StateVarPointer` and not de-structure this.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+                    } => {
+                        let new_state_var_ptr = StateVarPointer {
+                            component_idx: *component_idx,
+                            state_var_idx: *state_var_idx,
+                        };
+
+                        let new_current_freshness = dep.value.get_freshness();
+
+                        match new_current_freshness {
```
siefkenj, line 235: Why not match directly on `dep.value.get_freshness()`?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+        let mut found_stale_or_resolved_dependency = false;
+
+        for deps in dependencies_for_state_var.iter() {
+            for dep in deps.iter() {
+                match &dep.source {
+                    DependencySource::StateVar {
+                        component_idx,
+                        state_var_idx,
+                    } => {
+                        let new_state_var_ptr = StateVarPointer {
```
siefkenj, line 228: This is the pointer to the dep state var. Maybe call it `dep_state_var_ptr` instead of `new`?

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+                        };
+
+                        let new_current_freshness = dep.value.get_freshness();
+
+                        match new_current_freshness {
+                            // No need to recurse if the state var of the dependency is already fresh
+                            // so don't do anything to just continue the inner loop
+                            Freshness::Fresh => (),
+                            Freshness::Stale | Freshness::Resolved => {
+                                if !found_stale_or_resolved_dependency {
```
siefkenj, line 240: This is really tracking whether `state_var_ptr` has already been re-added to the stack. Maybe call this variable `state_var_ptr_on_stack`? And set it to false as the first thing you do in the loop.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+                    // as they cannot be stale
+                    DependencySource::Essential { .. } => (),
+                }
+            }
+        }
+
+        if !found_stale_or_resolved_dependency {
+            // All the dependencies of state_var_ptr are fresh.
+            // We calculate its value if a dependency has changed,
+            // else restore its previous value.
```
siefkenj, line 264: I still don't quite get this part. We might need to talk about it.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+    should_initialize_essential_data: bool,
+) {
+    let component_idx = state_var_ptr.component_idx;
+    let state_var_idx = state_var_ptr.state_var_idx;
+
+    let dependency_instructions;
+
+    {
+        let component = components[component_idx].borrow();
+        let state_var = &component.get_state_variables()[state_var_idx];
```
siefkenj, line 301: Can you explicitly type this. rust-analyzer is having trouble...

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
```diff
@@ -0,0 +1,401 @@
...
+    dependent_on_state_var: &mut Vec<Vec<Vec<StateVarPointer>>>,
+    dependent_on_essential: &mut Vec<HashMap<EssentialDataOrigin, Vec<StateVarPointer>>>,
+    essential_data: &mut Vec<HashMap<EssentialDataOrigin, EssentialStateVar>>,
+    // freshen_stack: &mut Vec<StateVarCalculationState>,
+    should_initialize_essential_data: bool,
+) {
+    let component_idx = state_var_ptr.component_idx;
+    let state_var_idx = state_var_ptr.state_var_idx;
+
+    let dependency_instructions;
```
siefkenj, line 297: Can you explicitly type this. rust-analyzer is having trouble...

siefkenj (commented):
The rewritten function looks so much better!

packages/doenetml-worker-rust/doenetml-core/src/utils.rs
dqnykamp: Good idea!

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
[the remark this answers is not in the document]
dqnykamp: Got tailcall to work for at least part of this, the new `resolve_state_var` function.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Have a much simpler version with depth-first now.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: I renamed it in case that helps our future selves.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: I suppose so. This code is now outdated, but I do that elsewhere. I assume it isn't a problem since I imagine the compiler will optimize that out.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: refactored code, so resolving this conversation.

packages/doenetml-worker-rust/doenetml-core/src/dependency.rs
dqnykamp: I tried changing this to `DependencyRequest` and renaming the other to `DependencyValueUpdateRequest` but then ended up reverting that change as it got too confusing due to the `DependencyValueUpdateRequest` then referring to a `request_idx` rather than an `instruction_idx`. With the `DependencyValueUpdateRequest` overlapping with the "instructions", it was too confusing to call the instructions "requests".

I'm happy to revisit, but it became too hard to disambiguate them for now.

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
dqnykamp: I couldn't figure out how to do that in rust. I removed most of the boilerplate code with the derive macros, but couldn't figure out how to remove the boilerplate specification of the fields. The derive macros don't let you change the structure itself; they only let you add code after the struct.

I decided to leave further attempts for removing the boilerplate fields to our future selves, but if you have any ideas on how to fix it now, all the better!

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Strange how my rust-analyzer doesn't have any trouble here or at the later spots.
It should able to infer it from pushing instruct_dependencies.
But, it doesn't hurt to add it.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Something is wrong with your rust-analyzer! All of these types come directly from the assignment...

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: I added this trait. The code is now shorter, but seems less explicit. I added some comments to explain what is going on. It seemed unnecessary to have a match since try_from did everything.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Done

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Done

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Changed the variable name and added more comments. Hope it is clearer now.

packages/doenetml-worker-rust/doenetml-core/src/state_var_calculations.rs
dqnykamp: Changed the variable name and added more comments. Hope it is clearer now.

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
dqnykamp: It looks like one could add common fields with attribute macros. Trying to decide if it is worth it now.

dqnykamp: Added an  `add_standard_component_fields` attribute macro to put all those standard fields in the components.

That gets rid of the boilerplate fields. I hope all these macros aren't too confusing to future developers.

packages/doenetml-worker-rust/doenetml-core/doenetml-derive/src/component_node.rs
```diff
@@ -0,0 +1,245 @@
...
+        },
+        _ => panic!("only structs and enums supported"),
+    };
+    output.into()
+}
+
+pub fn add_standard_component_fields_impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let component_fields_to_add = [
+        quote! { pub idx: ComponentIdx},
+        quote! { pub parent: Option<ComponentIdx>},
```
siefkenj, line 211: Strange spacing

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
```diff
@@ -1,20 +1,277 @@
...
+use crate::utils::KeyValueIgnoreCase;
 use crate::{ComponentChild, ComponentIdx, ExtendSource};
 
-use super::ComponentNode;
+use super::{
+    ComponentEnum, ComponentNode, ComponentNodeStateVariables, ComponentProfileStateVariable,
+    RenderedComponentNode,
+};
 
+#[add_standard_component_fields]
```
siefkenj, line 25: This is nice, but as you say, less explicit. I was thinking something like,
```rust
struct ComponentCommonData {
  idx: ComponentIdx,
  parent: Option<ComponentIdx>,
  ...
}

struct Text {
  common: ComponentCommonData,
  render_data: TextRendererData,
  ...
}
```

Won't that do basically the same thing (possibly affecting the layout of the data structure, but I assume we don't care)?

packages/doenetml-worker-rust/doenetml-core/doenetml-derive/src/component_node.rs
dqnykamp: How would you want it spaced? The formatter is forcing this spacing.

siefkenj: If the formatter is doing that, that's fine. It's just different fromt he lines below.

dqnykamp: Oh, I see what you are commenting on, the spacing inside the `quote!{}`. (I couldn't see the discrepancy as the snippet didn't show the lines below for comparison.) The formatter doesn't work inside the `quote!` macro, leading to such inconsistency if one isn't careful.  I'll fix that, assuming we want to keep this macro.

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
dqnykamp: I could go either way. I feel like the way you suggest will cause fewer headaches for developers, as it is clearer what's going on. It could be confusing if a developer tries to add a field that conflicts with one of the standard fields and wonder where the conflict came from.

It was fun to learn to write an attribute macro, though!

dqnykamp: I removed the attribute macro in favor of this common struct.

packages/doenetml-worker-rust/doenetml-core/src/component/text.rs
dqnykamp: Removed macro

packages/doenetml-worker-rust/doenetml-core/src/component/text_input.rs
dqnykamp: Deferring until another PR.

PR #91, opened 2024-01-11
Rust File Reorg

siefkenj:
This PR reorganizes files in `lib-doenetml-core` with the goal of making things easier to find for future component authors. The strategy was
  * Group files into folders named with their "type"
  * Keep the number of files in a folder down (the only _large_ folder the project should have is the `components/doenet` folder)
  * Keep the root `src/` file especially small. A new author shouldn't have to be blinded by dozens of files and have to guess which are the important entry points.
  * If a module has both a file and a folder, the file is only module declaration/re-exports (e.g., `pub mod XXX` or `pub use XXX`)

The following changes were made:
  * Almost everything was moved to `core/`
  * Core was split into `components`, `state`, and a few other modules (that mostly existed before).
  * `components` was split into `special` (where `_error` and `_external` go) and `doenet`, where all the authored components will go.


packages/doenetml-worker-rust/lib-doenetml-core/src/utils/logging.rs
```diff
@@ -0,0 +1,61 @@
+use std::collections::HashMap;
+
+pub trait KeyValueIgnoreCase<K, V> {
```
dqnykamp, line 3: This doesn't belong in `logging`.

siefkenj: Fixed!

dqnykamp (approved):

PR #95, opened 2024-01-13
Direclty use tsify types

siefkenj:
This PR adds doc strings to many of the Rust types and uses the exported Rust types instead of the hand-crafted typescript types.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast.rs
```diff
@@ -7,6 +7,7 @@ use wasm_bindgen::prelude::*;
 
 use crate::state::StateVarValue;
 
+/// Allowed children of the root node
```
dqnykamp, line 10: This is the root node, not the allowed children of the root node.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast.rs
```diff
@@ -18,6 +19,7 @@ pub struct DastRoot {
     pub position: Option<Position>,
 }
 
+/// Allowed children of an element node
```
dqnykamp, line 22: This is the allowed children of the root node or an element node.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -9,17 +9,6 @@ import type { DastRoot, DastElement, DastError } from "@doenet/parser";
 
 type Flags = Record<string, unknown>;
 
-export interface FlatDastElement extends Omit<DastElement, "children"> {
```
dqnykamp, line 12: Nice that this goes away!

PR #96, opened 2024-01-15
Strongly type Actions

siefkenj:
This PR introduces strong typing for actions sent to components. The Rust types are shared with Typescript.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/actions.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "web", derive(tsify::Tsify))]
+#[cfg_attr(feature = "web", tsify(from_wasm_abi))]
+pub struct ActionBody<T> {
```
dqnykamp, line 7: What function does `ActionBody` perform? Does it help with the types? A simpler alternative would to just put the action args directly in the action `enum` for each component. `ActionBody` does currently appear to do anything in the rust side.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/actions.rs
```diff
@@ -0,0 +1,14 @@
...
+    pub args: T,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "web", derive(tsify::Tsify))]
+#[cfg_attr(feature = "web", tsify(from_wasm_abi))]
+pub struct ActionBodyEmpty {}
```
dqnykamp, line 14: Do we need `ActionBodyEmpty`? It's not currently used, as you just are using unit variants when there is no action body.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
```diff
@@ -243,6 +244,14 @@ pub struct Action {
     pub args: HashMap<String, Vec<StateVarValue>>,
 }
 
+#[derive(Debug, Deserialize, Serialize, derive_more::TryInto)]
+#[cfg_attr(feature = "web", derive(Tsify))]
+#[cfg_attr(feature = "web", serde(rename_all = "camelCase"))]
+#[cfg_attr(feature = "web", tsify(from_wasm_abi))]
+pub enum AllActions {
```
dqnykamp, line 251: I suggest we move this into `common.rs` of `components`. That way, when creating a new component, you only have to add to `enum`s in one file.

I would also give this a singular name, as an instance of the `enum` actually represents just one action. The problem is that `Action` is already used. Maybe something like `ComponentAction`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/parse_action.rs
```diff
@@ -0,0 +1,59 @@
...
+            format!(
+                "Cannot deserialize action `{}` for component `{}`. Got error: {}",
+                action_json,
+                component.borrow(),
+                e
+            )
+        };
+
+        let action = match *component.borrow() {
+            ComponentEnum::TextInput(_) => {
```
dqnykamp, line 44: Can we write this so that we don't have to hard code any particular component type in this file? I'd like to make it easy as easy as possible to add new components, where it is just a matter of creating a new file `components/doenet` and then adding to enums in `cmponents/common.rs`.

I would suggest adding this logic to a trait of `RenderedComponentNode`, so that the implementation could be coded in each component file. The `enum_dispatch` macro will then automatically implement the trait on `ComponentEnum` so that it could be called here as something like `component.deserialize_action(action_json)`. (Not sure if there is common preprocessing of `action_json` before sending to the component's implementation.)

packages/doenetml-prototype/src/renderers/components/text-input.tsx
```diff
@@ -34,13 +34,12 @@ export const TextInput: BasicComponent<TextInputData> = ({ node }) => {
...
-                                actionName: "updateImmediateValue",
-                                componentIdx: id,
-                                args: { text: [e.target.value] },
-                            }),
-                        );
+                        const action: Action = {
+                            actionName: "updateImmediateValue",
+                            componentIdx: id,
+                            args: { text: e.target.value },
+                        };
```
dqnykamp, line 41: Good, I was hoping this array would go away.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -32,7 +32,7 @@ pub type AttributeName = &'static str;
 ///
 /// Each component type added to `ComponentEnum` must implement that component node traits.
 
-#[derive(Debug, EnumString)]
+#[derive(Debug, EnumString, Display)]
```
dqnykamp, line 35: It looks like the `strum::Display` macro is just being used for error messages. We already have the `.get_component_type()` method that gives the component name (as it would be used in `xml`), and that is used in other error messages. We should be consistent.

Maybe implement display for `ComponentEnum` that uses `get_componentType()`? Then, to maintain consistency, we could remove the reference to `.get_component_type()` in `state_var_updates.rs#226`.

It would take more work to get rid of `.get_component_type()` in `components/common.rs#249`, below, as that would require component developers to implement `Display` to call `.get_component_type()` on their struct, which is probably not worth the tiny gain of consistency.

siefkenj: I think we need to discuss this a little more. Since the xml name of the tag is not a one-to-one match of which component is being rendered. E.g., a `text` inside of a graph vs. outside. What do we want displayed in the error message? Something for the user or something for the developer, or both...?

siefkenj: Though, with `tsify` handling the errors, it is now moot.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/parse_action.rs
siefkenj: This has been eliminated by putting a `component` field in an `Action`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
siefkenj: Done!

packages/doenetml-worker-rust/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -1,11 +1,19 @@
 import * as Comlink from "comlink";
 import init, {
-    Action,
+    Action as _Action,
```
dqnykamp, line 3: Delete import? `_Action` doesn't seem to be used anywhere.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
dqnykamp: Right, I was thinking of these errors being displayed in the document, so they would be for the user. Hence, the desire for having the error message show the xml name that the author would have typed in. We can come back to this as we further develop the errors.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/CoreWorker.ts
dqnykamp: It seems to work when I delete it. But I'm OK leaving it in just in case.

dqnykamp (approved):

PR #99, opened 2024-01-17
Add support for attributes

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
```diff
@@ -19,4 +20,18 @@ where
 
         None
     }
+
+    /// Match key to the HashMap keys, ignoring case.
+    /// If found, remove the original value from the HashMap.
+    fn remove_ignore_case(&mut self, key: &str) -> Option<V> {
```
siefkenj, line 26: I don't know how this is used, but it smells of quadratic behavior somewhere...You might consider creating a hashmap of normalized names and a normalization function.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
```diff
@@ -19,4 +20,18 @@ where
...
+        let mut found_key_option = None;
```
siefkenj, line 27: Functional style would look nicer.

```rust
self.keys().find(|&k| k.to_string().eq_ignore_ascii_case(key).and_then(...)
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) {
+        state_var.set_value(self.string_dependency_value.get_fresh_value().clone());
+    }
+
+    fn request_dependencies_to_update_value(
+        &self,
+        state_var: &StateVarReadOnlyViewTyped<String>,
+        _is_direct_change_from_renderer: bool,
+    ) -> Result<Vec<DependencyValueUpdateRequest>, RequestDependencyUpdateError> {
+        let desired_value = state_var.get_requested_value();
```
siefkenj, line 187: Why not call it `requested_value`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+            }])
+        }
+    }
+}
+
+/// A simplified version of GeneralStringStateVarInterface
+/// that is based on a single dependency.
+/// Requires a `dependency_instruction_hint`.
+///
+/// Panics if `create_dependency_from_extend_source` is true,
```
siefkenj, line 129: I think we should shift to returning errors whenever we can.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+/// that doesn't result in at least one string dependency.
+#[derive(Debug, Default)]
+pub struct SingleDependencyStringStateVarInterface {
+    string_dependency_value: StateVarReadOnlyViewTyped<String>,
+}
+
+impl StateVarInterface<String> for SingleDependencyStringStateVarInterface {
+    fn return_dependency_instructions(
+        &self,
+        _extend_source: Option<&ExtendSource>,
```
siefkenj, line 141: I feel like `extending` or `extending_from` would be easier to understand.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+        parameters: &StateVarParameters,
+    ) -> Vec<DependencyInstruction> {
+        if parameters.create_dependency_from_extend_source {
+            panic!("Cannot create dependency from extend source for SingleDependencyStringStateVarInterface");
+        }
+
+        let mut dep_instructs: Vec<DependencyInstruction> = Vec::with_capacity(1);
+
+        if let Some(dependency_instruction) = &parameters.dependency_instruction_hint {
+            dep_instructs.push(dependency_instruction.clone());
```
siefkenj, line 151: This function is getting bigger? Is that why you wrote it this way? Otherwise you could return a `vec![...]` right here.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+        } else {
+            panic!(
+                "SingleDependencyStringStateVarInterface requires a dependency_instruction_hint"
+            );
+        }
+
+        dep_instructs
+    }
+
+    fn save_dependencies_for_value_calculation(&mut self, dependencies: &Vec<Vec<Dependency>>) {
```
siefkenj, line 161: I don't know the full algorithm, but would `save_dependencies` not be a reasonable name? Is there a special way they are saved for value calculations that is different to how they're saved other times?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+use super::common::create_dependency_instruction_from_extend_source;
+
+// use super::{
+//     StateVarInterface, StateVarMutableViewTyped, StateVarParameters, StateVarReadOnlyView,
+//     StateVarReadOnlyViewTyped,
+// };
+
+/// A string state variable interface that concatenates all string dependencies.
+///
+/// If `create_dependency_from_extend_source` is true and have an extend source extending this variable,
```
siefkenj, line 20: `have`->`has`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn return_dependency_instructions(
+        &self,
+        extend_source: Option<&ExtendSource>,
+        parameters: &StateVarParameters,
+    ) -> Vec<DependencyInstruction> {
+        let mut dep_instructs: Vec<DependencyInstruction> = Vec::with_capacity(2);
+
+        if parameters.create_dependency_from_extend_source {
+            if let Some(dep_inst) =
+                create_dependency_instruction_from_extend_source(extend_source, parameters)
```
siefkenj, line 41: I feel like using `and_then` looks better.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+                value: dep_value, ..
+            } in instruction.iter()
+            {
+                if let StateVarReadOnlyView::String(dep_string_value) = dep_value {
+                    string_vals.push(dep_string_value.create_new_read_only_view())
+                } else {
+                    panic!("Got a non-string value for a dependency for a GeneralStringStateVarInterface");
+                }
+            }
+        }
```
siefkenj, line 70: Would this look nicer with a `filter_map`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        self.string_dependency_values = string_vals;
+
+        if num_dependencies == 1 {
+            if let DependencySource::Essential { .. } = dependencies[0][0].source {
```
siefkenj, line 75: I prefer matches, especially if you're ignoring the contents.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        if let Some(dependency_instruction) = &parameters.dependency_instruction_hint {
+            dep_instructs.push(dependency_instruction.clone());
+        }
+
+        dep_instructs
+    }
+
+    fn save_dependencies_for_value_calculation(&mut self, dependencies: &Vec<Vec<Dependency>>) {
+        let num_dependencies = dependencies.iter().fold(0, |a, c| a + c.len());
```
siefkenj, line 55: `dependencies.iter().map(|x| x.len()).sum()`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
```diff
@@ -0,0 +1,197 @@
...
+                .string_dependency_values
+                .iter()
+                .map(|v| v.get_fresh_value().clone())
+                .collect();
+
+            state_var.set_value(value);
+        }
+    }
+
+    fn request_dependencies_to_update_value(
```
siefkenj, line 104: What is `_value` adding to this name?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -348,6 +344,199 @@ pub fn create_dependencies_from_instruction_initialize_essential(
...
+                        {
+                            let child_profile = child_profile_state_var.get_matching_profile();
+
+                            if match_profiles.contains(&child_profile) {
+                                child_matches_with_profile = Some(child_profile_state_var);
+                                break;
+                            }
+                        }
+
+                        if let Some(profile_sv) = child_matches_with_profile {
```
siefkenj, line 433: If you use `find_map` above, you can just continue here with `and_then`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -348,6 +344,199 @@ pub fn create_dependencies_from_instruction_initialize_essential(
...
+
+                            let state_var_dep = Dependency {
+                                source: DependencySource::StateVar {
+                                    component_idx: *child_idx,
+                                    state_var_idx: sv_idx,
+                                },
+                                value: state_var_view,
+                            };
+
+                            dependencies.push(state_var_dep);
```
siefkenj, line 449: Ah, I see. You were trying to do a `flat_map` the whole time!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -348,6 +344,199 @@ pub fn create_dependencies_from_instruction_initialize_essential(
...
+                // Treat the essential data as though it came from the first string attribute child
+                // of the component, except recursing to extend source components
+                // in order to share the essential data with the extend source.
+
+                let source_idx =
+                    get_recursive_extend_source_component_when_exists(components, component_idx);
+
+                let essential_origin = EssentialDataOrigin::AttributeChild(attribute_name, 0);
+
+                let essential_data_view =
```
siefkenj, line 511: I really wish we could rename `essential` to `primitive`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -396,3 +585,35 @@ fn get_children_with_parent_including_from_extend_source(
...
+                if let Some(&ExtendSource::Component(source_idx)) = component.get_extend() {
+                    return get_attribute_children_with_parent_falling_back_to_extend_source(
+                        components, source_idx, attribute,
+                    );
+                }
+            }
+
+            Some((attribute_children.clone(), component_idx))
+        }
+        None => None,
```
siefkenj, line 617: This is what `.map` does.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/special/_external.rs
```diff
@@ -123,4 +128,34 @@ impl ComponentNode for _External {
...
+
+    fn get_unevaluated_attributes(&self) -> &HashMap<String, DastAttribute> {
+        &self.common.unevaluated_attributes
+    }
+
+    fn get_unevaluated_attributes_mut(&mut self) -> &mut HashMap<String, DastAttribute> {
+        &mut self.common.unevaluated_attributes
+    }
+
+    fn get_is_rendered(&self) -> bool {
```
siefkenj, line 154: I'm still confused by `rendered`. I forget what it means, but `rendered` could mean "will be rendered" or could mean "has been rendered"

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input.rs
```diff
@@ -446,14 +542,13 @@ impl StateVarInterface<String> for ImmediateValueStateVarInterface {
...
-
-            // updates.push(DependencyValueUpdateRequest {
-            //     instruction_idx: 2,
-            //     dependency_idx: 0,
-            // });
+            self.bind_value_to
+                .request_change_value_to(desired_value.clone());
+
+            updates.push(DependencyValueUpdateRequest {
+                instruction_idx: 2,
```
siefkenj, line 549: This literal value terrifies me.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input.rs
```diff
@@ -328,94 +420,98 @@ impl StateVarInterface<String> for ValueStateVarInterface {
...
+    fn save_dependencies_for_value_calculation(&mut self, dependencies: &Vec<Vec<Dependency>>) {
         if let StateVarReadOnlyView::String(essential_value) = &dependencies[0][0].value {
             self.essential_value = essential_value.create_new_read_only_view();
         } else {
             panic!("Got a non-string essential value for immediate value of text input.");
         }
 
-        if let StateVarReadOnlyView::Boolean(sync_values) = &dependencies[1][0].value {
-            self.sync_values = sync_values.create_new_read_only_view();
+        if let StateVarReadOnlyView::Boolean(sync_immediate_value) = &dependencies[1][0].value {
```
siefkenj, line 479: We need to figure out an alternative interface here, because indexing arrays with literal values puts me on edge.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -8,7 +8,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::dast::{DastAttribute, Position as DastPosition};
 use crate::state::{
-    StateVar, StateVarName, StateVarReadOnlyView, StateVarReadOnlyViewTyped, StateVarValue,
+    StateVarEnum, StateVarName, StateVarReadOnlyView, StateVarReadOnlyViewEnum, StateVarValueEnum,
```
siefkenj, line 11: While we're renaming things...Is there a different type of view for a state var? Can we call it `StateVarView` instead of `StateVarReadOnlyView`?

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
```diff
@@ -28,7 +28,7 @@ pub fn component_node_derive(input: TokenStream) -> TokenStream {
             syn::Fields::Named(FieldsNamed { .. }) => {
                 // Convert struct name to camel case, preserving any initial '_'
                 let mut component_string = name.to_string();
-                if component_string.starts_with("_") {
+                if component_string.starts_with('_') {
```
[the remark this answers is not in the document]
dqnykamp: done

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
```diff
@@ -149,9 +151,35 @@ pub fn component_node_derive(input: TokenStream) -> TokenStream {
...
-                        //     &self.essential_state_vars
-                        // }
+                        fn set_attribute_children(
+                            &mut self,
+                            attribute_children: HashMap<AttributeName, Vec<ComponentPointerTextOrMacro>>,
+                        ) {
+                            self.common.attribute_children = attribute_children;
+                        }
+
+                        fn get_attribute_children_for_attribute(
```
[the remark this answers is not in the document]
dqnykamp: Yeah, I'm currently working on different types of attributes, though have only one implemented so far.

This is an attribute type where we treat all vector as children and apply the same logic as we do for component children, including expanding macros into the referents.

Another type of attribute will be where it is a references itself, so macros aren't expanded but left as a reference to be used in computations by the component.

In the Javascript version, I also have primitive attributes, where the attribute becomes a primitive value (number, string, etc.,) that can never be changed (and macros aren't allowed). Not yet sure if we need to add these in the new version.

So, the long name is explaining that we are in the first case and we are getting the "attribute children" for that attribute. Could be shortened to `get_children_for_attribute`, but I haven't decided whether the term `attribute children` is going to stick around.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
dqnykamp: This is used for one case only: We start with the original attributes, called `unevaluated_attributes` which is the hashmap that we get from the dast. Then, if the component type specifies that a particular attribute, it is removed from `unevaluated_attributes` with a case-insensitive match (function `evaluate_attributes_replace_macros`  of `component_creation.rs`).

Rust borrowing rules kept all my first algorithms from working, so I ended up with this one. More than happy to replace it with a better algorithm if this one is suspect. I don't care if it is a general purpose function on a HashMap like this or just programmed specifically in `evaluate_attributes_replace_macros`.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
dqnykamp: I tried approaches like this first, but then it got mad at me because `k` is borrowed and `remove` wouldn't work. I finally gave up and did this two-step approach.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: Older vocabulary crept back in!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: This is a programming error if someone made a mistake, so I'd want it to blow under these conditions. The right behavior is that the whole document should disappear and be replaced by an ugly error message. One option is that the worker quits and send a message to the renderer. We can certainly come up with better way to do this with propagating errors. I could create an issue with the task of determining the best approach.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: Yeah, I've been meaning to go find my notes where we discussed better options for this terminology.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: No, the function was getting smaller, as I copied it from `GeneralStringStateVarInterface` and didn't take advantage of the new simplicity.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: I was trying to distinguish between (A) this local saving of dependencies inside the state variable where it is doing the actual calculation and (B) the global saving of the dependencies in the core dependency graph. I could change it to just `save_dependencies` or `save_dependencies_on_state_variable` if the latter is clearer. Maybe I'll just stick with the shorter name.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: I felt like the name `request_dependencies_to_update` made it sound like you are updating the actual dependencies (i.e., changing the dependency graph). The name was trying to convey that we are just requesting to change the value of the dependency (i.e., `dependency.value`). We could simplify it to `request_dependencies_to_update` or other options such as: `request_new_dependency_values`, `request_updated_dependency_values`, `request_dependency_updates`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: Since the only result is a side-effect (appending to a vector), `and_then` feels funny to me. If I try using `and_then` it forces me to return an option. Returning `None` doesn't work, as it cannot infer the type (and there is no type). The best I could do was:
```
.and_then(|dep_inst| {
     dep_instructs.push(dep_inst);
     None::<()>
});
```
But that seemed weird so I reverted it. If I'm thinking about this wrong, let me know.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: That seems more succinct.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: Assuming you mean `flat_map`, yes, I think it looks better especially with `.try_into`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/common.rs
```diff
@@ -0,0 +1,31 @@
+use crate::{
+    components::prelude::{DependencyInstruction, StateVarParameters},
+    ExtendSource,
+};
+
+pub fn create_dependency_instruction_from_extend_source(
+    extend_source: Option<&ExtendSource>,
+    parameters: &StateVarParameters,
+) -> Option<DependencyInstruction> {
+    if let Some(ExtendSource::StateVar(extend_state_var_description)) = extend_source {
```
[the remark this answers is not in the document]
dqnykamp: I couldn't get `map` to work as there is a case even when the original `Option` is `Some` (so we're inside the `map`) where we want it to return `None`. Inside the `map` of an option, you have to return the type that is inside the `Some`.

I could get it to work with `and_then`, but I don't see the advantage. It adds an extra step and then just removes the `Some` on the `if` statement.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -0,0 +1,246 @@
...
+                    state_var.set_value(*boolean_value.get_fresh_value());
+                }
+            }
+            BooleanOrStrings::Strings(string_values) => {
+                let value: String = string_values
+                    .iter()
+                    .map(|v| v.get_fresh_value().clone())
+                    .collect();
+
+                if value.eq_ignore_ascii_case("true") || value.is_empty() {
```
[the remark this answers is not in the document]
dqnykamp: done

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -0,0 +1,246 @@
...
+                    );
+                } else {
+                    state_var.set_value(*boolean_value.get_fresh_value());
+                }
+            }
+            BooleanOrStrings::Strings(string_values) => {
+                let value: String = string_values
+                    .iter()
+                    .map(|v| v.get_fresh_value().clone())
+                    .collect();
```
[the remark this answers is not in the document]
dqnykamp: added comment

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: It looks like our last discussion ended with the proposal:
```
<point name="P>(1,2)</point>
<point extend="$P" x="3" />
<point clone="$P" x="3" />
```
So we'd still be using the word "extend". For "clone", there would be no link so nothing to add to the data structure.

You are suggesting we change the function parameter name. Are you also suggesting we change the name of the `enum` to something other than `ExtendSource`? I haven't come up with a better name for that. I'm OK with changing the function parameter name to either of the suggestions.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -112,28 +126,48 @@ pub trait StateVarInterface<T: Default + Clone>: std::fmt::Debug {
...
+    ///
+    /// May cause a subsequent `return_dependencies()` to panic if the resulting dependency
+    /// returns an incompatible type.
+    pub dependency_instruction_hint: Option<DependencyInstruction>,
+
+    /// Hint telling `return_dependency_instructions()` whether or not to use `extend_source`.
+    ///
+    /// If true and the extend source is a state variable where the shadowing name matches
+    /// the name of this state variable, then create a dependency from the shadowed state variable
+    pub create_dependency_from_extend_source: bool,
```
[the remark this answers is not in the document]
dqnykamp: It doesn't mean is is actually extending; it means that this particular state variable should extend if it were given an extend source. 

The best I could come up with was `should_create_dependency_from_extend_source`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
```diff
@@ -420,6 +412,7 @@ impl DoenetMLCore {
     ///
     /// Include warnings as a separate vector (errors are embedded in the tree as elements).
     pub fn to_flat_dast(&mut self) -> FlatDastRoot {
+        log!("***Calling to flat dast on core****");
```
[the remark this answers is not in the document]
dqnykamp: Yes, but before I remove it, let's figure out how to make sure `to_flat_dast` isn't called twice on initialization.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
```diff
@@ -301,14 +307,10 @@ impl DoenetMLCore {
     /// Returns a vector of the indices of the components reached.
     pub fn freshen_renderer_state(&mut self) -> Vec<ComponentIdx> {
         freshen_all_stale_renderer_states(
-            &mut self.stale_renderers,
+            &mut self.bookkeeping,
```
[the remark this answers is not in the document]
dqnykamp: I don't know. `clippy` make me reduce the number of function parameters, so I had to come up with some grouping to make it happy. `CoreBookkeeping` is a structure that has all the stacks for the recursive functions as well as the list of stale renderers. I'd be happy to switch to another name or to find a different way to get the number of parameters down or to tell clippy to ignore this lint.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -348,6 +344,199 @@ pub fn create_dependencies_from_instruction_initialize_essential(
...
+            if dependencies.is_empty() {
+                // Found no matching children.
+                // Create an essential dependency with the default_value for the state variable
+
+                // Treat the essential data as though it came from the first string child
+                // of the component, except recursing to extend source components
+                // in order to share the essential data with the extend source.
+
+                let source_idx =
+                    get_recursive_extend_source_component_when_exists(components, component_idx);
```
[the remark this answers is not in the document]
dqnykamp: I think in most cases, we just find the next extend source and don't keep recursing until we find the "original" source. Usually, we don't have to recurse because what we need is propagated each time a component extends something. However, we store essential data on the "original" extend source so that it is stored in just one place and then shared by all extenders. This function is just used to find where to create that essential data. (There isn't a one-level version of this exact function, but just similar functions the get the extend source.)

I'm not sure of a better name. Could say `get_original_extend_source_...`, though we haven't used that terminology `original` anywhere. `get_all` doesn't work, as it only gets a single one, i.e., the one that doesn't have another component extend source.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -348,6 +344,199 @@ pub fn create_dependencies_from_instruction_initialize_essential(
...
+
+            for child in attribute_children.iter() {
+                match child {
+                    ComponentPointerTextOrMacro::Component(child_idx) => {
+                        let child = components[*child_idx].borrow();
+
+                        let mut child_matches_with_profile = None;
+                        for child_profile_state_var in
+                            child.get_component_profile_state_variables().iter()
+                        {
```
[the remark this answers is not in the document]
dqnykamp: I rewrote this section with `filter_map` followed by `find_map`. Is it clearer or more idiomatic?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
dqnykamp: I'm still not convinced by `primitive`. Maybe that's because I've used `primitive` in DoenetML more in reference to primitive data types when creating attribute primitives, where these attributes could only be primitive data types without dynamic features. The essential data does currently correspond to primitive data types, but it won't stay that way as we add more complicated features such as arrays.

If the term `essential` is going to be confusing, that's a strike against it. It'll just take more work to dislodge it from the essential place it has in my head when thinking about the DoenetML data structure.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
dqnykamp: I couldn't get `.map` to work because of the recursion that returns an option, but I got `.and_then` to work. I'm slowly figuring out how these functions work!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/special/_external.rs
dqnykamp: It means neither. It specifies a fixed attribute of a component rather than a state. A `rendered` component is a component that will/does appear in the flat dast tree. A `non-rendered` component still appears in the `flat_dast` because the `flat_dast` contains all components even if they aren't in the tree. But, we won't spend the cpu cycles calculating the rendered state for a component it if is not `rendered`.

For state variables, I used the attribute `for_renderer` to distinguish which variables are in the rendered data. Maybe for components, as in here, I'll call it `in_render_tree`. Would that be clearer?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
dqnykamp: We have both a `StateVarMutableView` and a `StateVarReadOnlyView`.

The `StateVarMutableView` is used, for example, in `calculate_state_var_from_dependencies_and_mark_fresh` because in that function, it sets the value of the state variable.

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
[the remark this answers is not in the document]
siefkenj: Would `get_attribute` and `get_attribute_preserve_references` serve those two cases?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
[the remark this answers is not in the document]
siefkenj: We don't need to look into this now, but when we are performance-focused, we may look at various _interned string_ libraries that can prevent excess clones.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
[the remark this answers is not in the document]
siefkenj: Maybe `dependency_stack`? or `current_dependency_stack`? or `_processing_state`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
siefkenj: What about `RootVar`? As long as it doesn't get confused with the DAST root...or `DependencylessVar` or `NoDepData`?

I'm only one sample point, but to me `essential` means important and indispensable. But all the vars that are referenced are indispensable when computing the final state. The `essential` ones here just indicates where a calculation must start.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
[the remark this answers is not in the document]
siefkenj: What about `get_extend_source_origin`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
siefkenj: I like `request_dependency_updates`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
siefkenj: I agree, that doesn't look nice. `.map` would eliminate that, but might be a strange usage...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
siefkenj: I was suggesting just the parameter name.

To me `<point name="Q" extend="$P" />` reads as `the point $Q extends the component $P`. But when you ask a question `what does $Q extend?`, I prefer `$Q is extending $P`, rather than `the source that $Q declares it will extend is $P$`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
siefkenj: An issue sounds good. A common pattern is for there to be a `dev mode` where errors are fatal, but then have a `production mode` where any recoverable errors are delt with. It would be nice for your problem not to dissapear in the middle of a midterm because of one of your mistakes...

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
siefkenj: A clone was missing.
```
        self.keys()
            .find(|k| k.to_string().eq_ignore_ascii_case(key))
            .map(|k| k.clone())
            .and_then(|k| self.remove(&k))
```
works, but we might consider using `.retain`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/common.rs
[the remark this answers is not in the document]
siefkenj: I didn't notice you were double destructuring (which might be a reason not to use the `if let ...`. Here is what the alternative version looks like

```rust
    extending.and_then(|extend_source| match extend_source {
        ExtendSource::StateVar(extend_state_var_description) => extend_state_var_description
            .state_variable_matching
            .iter()
            .find(|state_var_match| {
                state_var_match
                    .shadowing_idx
                    .map(|sv_idx| sv_idx == state_var_idx)
                    .unwrap_or(parameters.is_primary_state_variable_for_shadowing_extend_source)
            })
            // Either
            // 1. shadowing index was supplied and it matches the index of the state variable, or
            // 2. shadowing index was not supplied and this variable is the primary state variable
            //    for use when shadowing extend sources.
            // Therefore, we shadow the extend source.
            .map(|state_var_match| DependencyInstruction::StateVar {
                component_idx: Some(extend_state_var_description.component_idx),
                state_var_idx: state_var_match.shadowed_idx,
            }),
        _ => None,
    })
```

I don't claim it's the most elegant.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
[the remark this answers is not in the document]
dqnykamp: I like it!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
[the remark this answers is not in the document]
dqnykamp: changed to `processing_state`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/text_state_var_interfaces.rs
dqnykamp: OK. I'm working on this function now, so can change the name to `request_dependency_updates`.

PR #102, opened 2024-01-25
Outlining test framework

siefkenj:
This PR adds some examples and utilities for writing Rust tests of core. See `tests/core_basic.rs` for a test that retrieves the `FlatDast` from core and compares it to a JSON string.


packages/doenetml-worker-rust/lib-doenetml-core/src/utils/keyvalue.rs
```diff
@@ -11,17 +11,56 @@ where
...
+    }
+
+    #[test]
+    fn test_remove_ignore_case() {
+        let mut map = HashMap::new();
+        map.insert("Key1".to_string(), "Value1");
+        map.insert("KEY2".to_string(), "Value2");
+
+        assert_eq!(map.remove_ignore_case("key1"), Some("Value1"));
+        assert_eq!(map.remove_ignore_case("key2"), Some("Value2"));
```
dqnykamp, line 63: And then check that these keys are indeed removed from the hashmap?

packages/doenetml-worker-rust/lib-doenetml-core/tests/core_basics.rs
```diff
@@ -0,0 +1,53 @@
...
+                    "type": "element",
+                    "name": "text",
+                    "attributes": {},
+                    "children": [],
+                    "data": {
+                        "id": 1,
+                        "action_names": [],
+                        "state": {
+                            "value": "you",
+                            "text": "you"
```
dqnykamp, line 44: This is a flaw. The `text` state variable should not be marked `for_renderer`.  Should delete the "text" field from "state" and delete line 25 of `text.rs`

packages/doenetml-worker-rust/lib-doenetml-core/tests/test_utils.rs
```diff
@@ -0,0 +1,173 @@
...
+            std::io::ErrorKind::Other,
+            str::from_utf8(&output.stderr).unwrap().to_string(),
+        )),
+    }
+}
+
+/// Expand the string containing DoenetML into a JSON string of the parsed DAST
+/// omitting the position props.
+#[macro_export]
+macro_rules! dast_no_position {
```
dqnykamp, line 44: Just curious: what do these macros do that couldn't be done with a function?

packages/doenetml-worker-rust/lib-doenetml-core/tests/core_basics.rs
siefkenj: Found a bug already! I'd say the tests are worthwhile :-)

packages/doenetml-worker-rust/lib-doenetml-core/tests/test_utils.rs
siefkenj: Good point! Theoretically there could be more flexibility down the line, but I have converted them all to functions.

PR #103, opened 2024-01-25
state var interface with parameters

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -62,30 +65,34 @@ impl TryFrom<&StateVarReadOnlyViewEnum> for BooleanOrString {
...
+impl GeneralBooleanStateVarInterface {
+    pub fn new(base_dependency_instruction: DependencyInstruction) -> Self {
+        GeneralBooleanStateVarInterface {
+            base_dependency_instruction,
+            ..Default::default()
+        }
+    }
+
+    /// Set parameter `is_primary_state_variable` to true
+    pub fn is_primary_state_variable(mut self) -> Self {
```
siefkenj, line 77: I think non-mutable would be better.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -36,8 +39,8 @@ struct GeneralBooleanStateVarDependencies {
 
 #[derive(Debug, Default, StateVariableDependencyInstructions)]
 struct GeneralBooleanStateVarDependencyInstructions {
```
siefkenj, line 41: Docstring?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -21,9 +21,12 @@ use super::util::{create_dependency_instruction_from_extend_source, string_to_bo
 /// then propagate the `used_default` attribute of the essential state variable.
 #[derive(Debug, Default)]
 pub struct GeneralBooleanStateVarInterface {
+    is_primary_state_variable: bool,
```
siefkenj, line 24: Can you add docstrings explaining what these are? Maybe even with sample code?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
siefkenj: You need to recreate the whole object.

dqnykamp: I could clone it and then set the new parameter, but since that wouldn't be a good idea if one had large objects.

Is it better to recreate by specifying all the fields again like below? Seems simpler to mutate the original object, given that you've taken ownership of it anyway.

```
        GeneralStringStateVarInterface {
            is_primary_state_variable: true,
            base_dependency_instruction: self.base_dependency_instruction,
            dependency_instructions: self.dependency_instructions,
            dependency_values: self.dependency_values,
            from_single_essential: self.from_single_essential,
        }
```

siefkenj: Yeah. Recreating it is the proper way. The compiler should optimize an inefficiency away.

dqnykamp: OK. It became a moot point, though, as I eliminated the primary state variable.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
dqnykamp: Added docstring, but didn't add sample code. Simpler now, though, without primary state variable

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/boolean.rs
```diff
@@ -12,35 +12,46 @@ pub struct Boolean {
...
+            BooleanStateVariables::get_value_state_variable_index(),
+        )
+    }
+}
+
 #[derive(Debug, ComponentStateVariables)]
 pub struct BooleanStateVariables {
+    #[is_public]
+    #[for_renderer]
+    #[component_profile_state_variables(Boolean)]
```
siefkenj, line 32: docstring? Hopefully when we get enough of this working, I can look into extracting the docstrings to automatically create documentation :-)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input.rs
```diff
@@ -191,6 +171,14 @@ struct ValueStateVarInterface {
     dependency_values: ValueDependencies,
 }
 
+impl ValueStateVarInterface {
```
siefkenj, line 174: I forget...why is there `Interface` attached to these names?

siefkenj (commented):
Looks good. Still don't understand 100%, but I think it can be merged for more polishing later.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input.rs
dqnykamp: One piece of `StateVar<T>` is the "interface", which is the trait object defining the functions to set dependencies, calculate value, etc. The rest of the state variable is typed. Not at all stuck on the term "interface".

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/boolean.rs
dqnykamp: I cannot figure out how to add a docstring to derive macro help attributes! I could put the documentation on the `ComponentStateVariables` derive macro, but I'd really like to be able to hover over `#[for_renderer]` to see what that means.

siefkenj: I'm confused. I meant to imply that `value` should have a doc string, not the macros themselves.

dqnykamp: OK. I added a docstring to `ComponentStateVariables` given that I misunderstood your request. Now added docstrings to the state variable fields.

PR #104, opened 2024-01-25
Allow use of `log!` in tests

siefkenj:
Add a new feature flag to allow `log!` to be compiled differently when running tests. This allows `log!` statements to be included in code (and have the tests still pass!)


packages/doenetml-worker-rust/lib-doenetml-core/src/core/doenet_ml_core.rs
```diff
@@ -409,7 +409,7 @@ impl DoenetMLCore {
     ///
     /// Include warnings as a separate vector (errors are embedded in the tree as elements).
     pub fn to_flat_dast(&mut self) -> FlatDastRoot {
-        //log!("***Calling to flat dast on core****");
+        log!("***Calling to flat dast on core****");
```
dqnykamp, line 412: Great, I'm glad we can have logs again!

You can delete this `log!` now, if you like, as I verified `to_flat_dast` is not called twice if I run the built version of test-viewer.

siefkenj: Removed!

dqnykamp (approved):

PR #105, opened 2024-01-27
request updates using macros

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/dependency.rs
```diff
@@ -101,11 +102,12 @@ impl Deref for DependenciesCreatedForInstruction {
 }
 
 /// Information which update were requested so that we can recurse
```
siefkenj, line 104: `Information` -> `Inform`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -509,24 +523,34 @@ impl<T: Default + Clone> StateVarReadOnlyView<T> {
...
 
-    /// Set the `requested_value` field to the supplied value
-    pub fn request_change_value_to(&self, requested_val: T) {
-        self.inner
-            .borrow_mut()
-            .request_change_value_to(requested_val);
+    /// Queue an update to the dependency that will
+    /// attempt to set its value to the `requested_value` argument.
+    ///
+    /// To send all queued updates to core, call `return_queued_updates()`
```
siefkenj, line 543: Can we call this `queued_updates`? I don't think `return` adds anything.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -535,17 +559,66 @@ impl<T: Default + Clone> StateVarReadOnlyView<T> {
     }
 }
 
+#[enum_dispatch]
+pub trait QueryUpdateRequests {
+    /// Reset variable that tracks whether or not an update has been requested
```
siefkenj, line 564: The implementation doesn't look like it is resetting anything, since it's just setting it to false...Maybe `clear_queued_update` would be a better name?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -535,17 +559,66 @@ impl<T: Default + Clone> StateVarReadOnlyView<T> {
...
+    fn reset_queued_updates(&mut self);
+
+    fn return_indices_with_queued_updates(&self) -> Vec<usize>;
```
siefkenj, line 567: Strike `return_`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -535,17 +559,66 @@ impl<T: Default + Clone> StateVarReadOnlyView<T> {
...
+}
+
+impl<T: Default + Clone> QueryUpdateRequests for StateVarReadOnlyView<T> {
+    /// Reset 'update_has_been_queued` to false
+    fn reset_queued_updates(&mut self) {
+        self.update_has_been_queued = false;
+    }
+
+    fn return_indices_with_queued_updates(&self) -> Vec<usize> {
```
siefkenj, line 576: Why is this a vec? Will there be times when multiple things are actually returned?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -535,17 +559,66 @@ impl<T: Default + Clone> StateVarReadOnlyView<T> {
...
+            val.reset_queued_updates();
+        }
+    }
+
+    fn return_indices_with_queued_updates(&self) -> Vec<usize> {
+        // Note: this algorithm does not correctly handle Vec<Vec<T: QueryUpdateRequests>>
+        self.iter()
+            .enumerate()
+            .filter_map(|(idx, val)| {
+                if val.return_indices_with_queued_updates().is_empty() {
```
siefkenj, line 600: Seems like you should have returned an option to begin with...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -678,17 +751,21 @@ impl<T: Default + Clone> StateVar<T> {
     /// Convenience function to call `calculate_state_var_from_dependencies` on interface
     /// and then call mark_fresh
     pub fn calculate_state_var_from_dependencies_and_mark_fresh(&self) {
-        self.interface
-            .calculate_state_var_from_dependencies(&self.value);
+        match self.interface.calculate_state_var_from_dependencies() {
+            StateVarCalcResult::Calculated(val) => self.value.set_value(val),
```
siefkenj, line 755: This worries me. Do you know for a fact that `came_from_default == False` on `self.value`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -144,25 +164,26 @@ impl StateVarInterface<bool> for GeneralBooleanStateVarInterface {
...
 
-    fn calculate_state_var_from_dependencies(&self, state_var: &StateVarMutableView<bool>) {
+    fn calculate_state_var_from_dependencies(&self) -> StateVarCalcResult<bool> {
         if self.have_invalid_combination {
-            state_var.set_value(false);
+            StateVarCalcResult::Calculated(false)
         } else if self.dependency_values.booleans_or_strings.len() == 1 {
             match &self.dependency_values.booleans_or_strings[0] {
                 BooleanOrString::Boolean(boolean_value) => {
                     if self.from_single_essential {
```
siefkenj, line 173: How about `if self.from_single_essential && boolean_value.came_from_default()`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -144,25 +164,26 @@ impl StateVarInterface<bool> for GeneralBooleanStateVarInterface {
         }
     }
 
-    fn calculate_state_var_from_dependencies(&self, state_var: &StateVarMutableView<bool>) {
+    fn calculate_state_var_from_dependencies(&self) -> StateVarCalcResult<bool> {
         if self.have_invalid_combination {
-            state_var.set_value(false);
+            StateVarCalcResult::Calculated(false)
```
siefkenj, line 169: You're using a mix of implicit returns and explicit ones. Please be consistent, especially in deeply nested structures.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
```diff
@@ -244,6 +257,12 @@ impl SingleDependencyBooleanStateVarInterface {
     }
 }
 
+impl From<SingleDependencyBooleanStateVarInterface> for StateVar<bool> {
```
siefkenj, line 260: I think we should get rid of all the `SingleDependency` stuff until we know we need it.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -194,23 +200,36 @@ pub trait ComponentNode: ComponentStateVariables {
...
+pub trait RenderedChildren: ComponentNode {
     /// Return the children that will be used in the flat dast sent to the renderer.
     fn get_rendered_children(&self) -> &Vec<ComponentPointerTextOrMacro> {
         self.get_children()
     }
+}
 
+/// The ComponentAttributes trait can be derived for a component,
+/// giving it the default implementation of no attributes.
+/// To add attributes, a component type can implement the trait to override the defaults.
```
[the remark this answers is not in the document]
dqnykamp: Sure. I was writing this from the point of the view of the worker, for which there are not attributes. Maybe the phrasing of "ignoring all attributes" is clearer.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -194,23 +200,36 @@ pub trait ComponentNode: ComponentStateVariables {
     fn set_is_in_render_tree(&mut self, is_in_render_tree: bool);
 }
 
-/// The RenderedComponentNode trait can be derived for a component, giving it the default implementations.
-/// To add actions or what information is sent when rendering, a component type can implement
+/// The RenderedChildren trait can be derived for a component, giving it the default implementation
+/// of the rendered children being the same as the children.
+/// To specify other rendered children, a component type can implement
```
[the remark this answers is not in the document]
dqnykamp: I could remove the default implementation as specified by the trait and have the derive macro implement the trait where it pass through all the children. Then it wouldn't need `: ComponentNode`, as the derive macro could grab `.common.children` directly.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/p.rs
```diff
@@ -1,11 +1,19 @@
 use crate::components::prelude::*;
 
-#[derive(Debug, Default, ComponentNode, RenderedComponentNode, ComponentStateVariables)]
+#[derive(
+    Debug,
+    Default,
+    ComponentNode,
+    RenderedChildren,
```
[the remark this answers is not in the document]
dqnykamp: I could name the derive macro `PassThroughChildren`, which would derive an implementation of the `RenderedChildren` trait that gets the children. I feel, though, that it would be confusing and deviating from convention if the name of the derive macro differed from the name of the trait it implements.

I could have the `RenderedChildren` derive macro define to helper attributes to specify different options for the derive behavior. For example: `#[pass_through_children]` and `#[no_rendered_children]`.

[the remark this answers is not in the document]
dqnykamp: And, if one didn't specify an attribute, it would pass through the children.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
dqnykamp: Yes, for the case when on vectors, below.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
dqnykamp: Not sure how `Option<Vec<T>>` is better than `Vec<T>`. In the former case, both `None` and `Some(vec![])` would represent the state of having nothing.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
dqnykamp: If the `calculate` function returns the `Calculated` variant of the enum, that means we are to set `came_from_default` to false. If the `calculate` function returns the `FromDefault` variant of the enum, that means we are to set `came_from_default` to true. We don't have a way to know that the `calculate` function didn't lie when chosing the variant to return, but we act on what it outputs.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
dqnykamp: done

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state_var_interfaces/boolean_state_var_interfaces.rs
dqnykamp: I prefer explicit return in this case, but clippy is weird and randomly chooses some to say they should be implicit. Turning off this clippy warning for this function.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
siefkenj: What I was asking is if it was initialized correctly. I would rather see `self.value.set_value_and_set_came_from_default(val, false)`

packages/doenetml-worker-rust/lib-rust-macros/src/component_state_methods.rs
```diff
@@ -137,16 +137,16 @@ pub fn component_state_derive(input: TokenStream) -> TokenStream {
                         //     x if x.eq_ignore_ascii_case(#field_camel_case) => Some(#sv_idx),
                         // });
 
-                        if check_if_field_has_attribute(&named[sv_idx], "is_public") {
+                        if check_if_have_attribute(&named[sv_idx].attrs, "is_public") {
```
siefkenj, line 140: Such a function is normally called `hasAttribute(...)`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/state_vars/immediate_value.rs
```diff
@@ -69,12 +51,10 @@ impl StateVarUpdater<String> for ImmediateValueStateVar {
     }
 
     fn invert(
-        &mut self,
+        data: &mut RequiredData,
```
siefkenj, line 54: It would be great if we could find a way to make this immutable. Then we wouldn't have to think about whether the parent caller was doing something else with the data. Alternatively, can we consume `RequiredData`? Or does Core need to hold onto that?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.rs
```diff
@@ -72,13 +45,13 @@ enum BooleanOrString {
     String(StateVarView<String>),
 }
 
-// We implement TryFrom `StateVarViewEnum`
-// so that we can `try_into` `RequiredData`
-// from the vector of dependencies.
-impl TryFrom<&StateVarViewEnum> for BooleanOrString {
+// We implement TryFromState
+// because all RequiredData must implement this trait.
+// (Needed to create the create the RequiredData from the information sent the state variable)
```
siefkenj, line 50: `create the create`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/state_vars/immediate_value.rs
dqnykamp: It needs to be mutable because the `StateVarView<T>` stores the requested value on itself (even though it is "read only" as far as core is concerned, it is a mutable variable as far as rust is concerned).

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.rs
```diff
@@ -183,12 +142,18 @@ impl StateVarUpdater<bool> for BooleanStateVar {
...
+        } else if data
+            .base
+            .iter()
+            .any(|dep_value| matches!(dep_value, BooleanOrString::Boolean(_)))
+        {
+            // invalid combination. Haven't implemented boolean dependency with other
+            return StateVarCalcResult::Calculated(false);
         } else {
             // concatenate the string values into a single string
             // TODO: can we do this without cloning?
```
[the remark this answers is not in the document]
dqnykamp: Is that better than `collect()`? I figured you'd like that less because it would require `value` to be mutable!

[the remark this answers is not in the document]
siefkenj: I read a clever trick on a blog yesterday.

```rust
let value = {
  let mut value = String::new();
  value.extend(...);
  value
}
```

Then mutability is constrained to exactly where it needs to be.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/state_vars/immediate_value.rs
siefkenj: But there could be another structure that we create, mutate, and then return. Isn't the only thing that's needed here is that the data is being saved just long enough to do `data.queued_update()`? Or will we access the `is_queued`/`requested_value` state from some other function? If the latter is the case, the doc string needs some bold text pointing out that the mutation has big consequences.

siefkenj: A possible solution is if `RequiredData` had an `update_queue() -> RequiredDataUpdateQueue` function. Then we just make a mutable update queue in this function, use it to generate our `queued_updates()` and then throw it away.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
siefkenj: Makes sense! How about `set_value_from_default` for the function name?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/state_vars/immediate_value.rs
dqnykamp: No, the way we communicate the value through the cached data is to directly set the requested value on the cached data. All we send to core is a list of dependencies to recurse to, not their desired values. (I knew you weren't going to like this once you discovered what was going on!)

dqnykamp: The reason for doing this is the same for using the typed, cached data: to avoid the need to pass data through core, where we'd need to wrap in enums and unwrap on the other end. We use the cached data (with its built in Rcs) to communicate in both directions, relying on the structure of `StateVarView<T>`, including its private fields and limited exposed API, to ensure data consistency.

dqnykamp: Wait a minute, the only thing that isn't behind a `RefCell` is the `update_has_been_queued` flag, the info of which does get thrown away after we generate our `queued_updates`, so maybe there is a way to avoid the `mut` as you suggested.

dqnykamp: I'm not immediately seeing a way to do this, but it seems like there should be a way to get rid of the mutability of `RequiredData`. Unless you see a quick way do to this, maybe make an issue rather than hold us this PR?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
dqnykamp: Nice. Glad to have help coming up with less awkward names.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/state_vars/immediate_value.rs
siefkenj: Another PR sounds good. This one is getting big :-P

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var/views.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Equivalent to calling `get_requested_value()` and then calling `set_value()` with that value,
+    /// except more efficient due to only one borrow of the inner value.
+    pub fn set_value_to_requested_value(&self) {
+        let mut inner = self.inner.borrow_mut();
+        let new_value = inner.get_requested_value().clone();
+        inner.set_value(new_value);
+    }
+}
+
+impl<T: Default + Clone> StateVarView<T> {
```
siefkenj, line 201: If they're in the same file, can we group the struct def and the impl together?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var/views.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Structure containing the value of the variable its meta data.
+    /// Since inner is in an Rc<RefCell>, it is shared with other views and could be changed by them.
+    inner: Rc<RefCell<StateVarInner<T>>>,
+
+    /// A change counter that can be compared to the change counter of inner
+    /// in order to determine if the state variable has changed since last viewed.
+    change_counter_when_last_viewed: u32,
+}
+
+/// A read-only view of the value of the state variable.
```
siefkenj, line 22: I'm confused by the use of the word `value`. Doesn't a view have more functionality than just seeing the value?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.rs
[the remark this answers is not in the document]
dqnykamp: Redid with the extend. I didn't add that extra trick as `value` falls out of scope after just one more line.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var/updater.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// for this state variable. These queries may be based on structure of the document,
+    /// e.g., the children, attributes, or other state variables
+    /// of the component of this state variable.
+    fn return_data_queries(
+        &self,
+        extending: Option<ExtendSource>,
+        state_var_idx: StateVarIdx,
+    ) -> Vec<Option<DataQuery>>;
+
+    /// Calculate the value of the state variable from the currently cached query results.
```
siefkenj, line 49: No longer cached. Now it's the passed in `data`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var.rs
```diff
@@ -239,20 +148,20 @@ impl<T: Default + Clone> StateVarInner<T> {
     }
 
     /// Set the value of the state variable to `new_val`, mark it as Fresh,
-    /// set 'used_default` to false, and increment the change counter.
+    /// set `came_from_default` to false, and increment the change counter.
```
siefkenj, line 151: backticks around `false`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
```diff
@@ -1,12 +1,12 @@
 use crate::{
-    components::prelude::{DependencyInstruction, StateVarIdx},
+    components::prelude::{DataQuery, StateVarIdx},
     ExtendSource,
 };
 
-pub fn create_dependency_instruction_if_match_extend_source(
+pub fn create_data_query_if_match_extend_source(
```
siefkenj, line 6: Docstring?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -0,0 +1,158 @@
...
+                // multiple string variables, so concatenate
+                // TODO: can we implement this without cloning the inner value?
+                let value: String = strings.iter().map(|v| v.get().clone()).collect();
+
+                StateVarCalcResult::Calculated(value)
+            }
+        }
+    }
+
+    fn invert(
```
siefkenj, line 126: Should probably explain the algorithm in a docstring.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -0,0 +1,158 @@
...
+        // concatenate the strings from extending and the base data query
+        let mut strings = if let Some(extending) = data.extending.as_ref() {
+            vec![extending]
+        } else {
+            vec![]
+        };
+
+        strings.extend(data.base.iter());
+
+        match strings.len() {
```
siefkenj, line 97: I like the look of this :-)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var/views.rs
dqnykamp: Done

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/state_var/views.rs
dqnykamp: I think the term `value` must have come from the definition of `StateVar<T>` struct with its `value` field. This docstring is definitely outdated.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
dqnykamp: You can thank `clippy` for suggesting it.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.rs
```diff
@@ -158,15 +158,13 @@ impl StateVarUpdater<bool, BooleanRequiredData> for BooleanStateVar {
...
-                .iter()
-                .map(|v| match v {
-                    BooleanOrString::Boolean(boolean_val) => boolean_val.get().to_string(),
-                    BooleanOrString::String(string_value) => string_value.get().to_string(),
-                })
-                .collect();
+
+            let mut value = String::new();
+            value.extend(data.base.iter().map(|v| match v {
+                BooleanOrString::Boolean(boolean_val) => boolean_val.get().to_string(),
```
siefkenj, line 164: Does it work without the `to_string` now? That was supposed to be the whole point of switching to this method (because `extend` can work on references).

dqnykamp: No, it doesn't currently work if I remove the `.to_string()` on the strings. In this case with the `Boolean`, it gets mad about incompatible match arm types. In `string_state_var.rs`, it gets made because the output of `.get()` is currently typed as `impl Deref<Target = String> + '_`, which doesn't implement `Expand`.

PR #109, opened 2024-01-31
State variable refactoring and tests

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -158,3 +158,273 @@ impl StateVarUpdater<String, StringRequiredData> for StringStateVar {
...
+            StringStateVar::calculate(&string_data),
+            StateVarCalcResult::Calculated("hello".to_string())
+        );
+
+        // if change the value of essential as though it were the default,
+        // get the new default value
+        essential.set_value_from_default("bye".to_string());
+        assert_eq!(
+            StringStateVar::calculate(&string_data),
+            StateVarCalcResult::FromDefault("bye".to_string())
```
siefkenj, line 194: Maybe it's not good for our specific case, but could we use something like https://stackoverflow.com/questions/55079070/how-to-accept-str-string-and-string-in-a-single-function to remove all these `.to_string()`s?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -158,3 +158,273 @@ impl StateVarUpdater<String, StringRequiredData> for StringStateVar {
...
+
+    #[test]
+    fn string_state_var_with_no_dependencies() {
+        // set up data that consists of just the essential variable
+        let mut string_data = StringRequiredData::default();
+        let essential = StateVarMutableView::new_with_value("".to_string(), true);
+        string_data.essential = essential.create_new_read_only_view();
+
+        // add meta data needed for invert to work. Use arbitrary values just to test.
+        string_data._data_query_mapping_data.essential.push((5, 7));
```
siefkenj, line 174: Why is this set here and not right before the `invert()` call?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -158,3 +158,273 @@ impl StateVarUpdater<String, StringRequiredData> for StringStateVar {
...
+
+        let extending = StateVarMutableView::new_with_value("Hello".to_string(), false);
+        string_data.extending = Some(extending.create_new_read_only_view());
+
+        let base = StateVarMutableView::new_with_value("World".to_string(), false);
+        string_data.base = vec![base.create_new_read_only_view()];
+
+        // add meta data needed for invert to work. Use arbitrary values just to test.
+        string_data._data_query_mapping_data.extending.push((3, 2));
+        string_data._data_query_mapping_data.base.push((5, 1));
```
siefkenj, line 399: It would be nice to move these setups to their own function. Any function you define inside of `mod test {}` but without `#[test]` will just be a regular function that you can call, but will not be compiled during regular builds.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
```diff
@@ -158,3 +158,273 @@ impl StateVarUpdater<String, StringRequiredData> for StringStateVar {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn string_state_var_with_no_dependencies() {
```
siefkenj, line 167: It would be nice to have some docstrings that had some example doenetml where such a state variable would arise.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
dqnykamp: I don't think it helps as most of these are cases where the function needs to take ownership of the `String`. Could use `String::from("hi")` is that looks more pleasing!

siefkenj: What if a `to_owned()` is called afterwards? Can we eat the reference?

dqnykamp: If one used the trait bound `AsRef<str>`, I don't think you can call `.to_owned()`, as I can't imagine `AsRef<str>` would imply an implementation of `.to_owned()`. My guess is the trait bound `AsRef<str>` is good only if all you need is a reference.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{
+    components::prelude::{Dependency, StateVarMutableView, StateVarView, StateVarViewEnum},
+    dependency::DependencySource,
+    state::essential_state::EssentialDataOrigin,
+};
+
+/// Create a mutual view of a state variable along with a read only view
```
siefkenj, line 7: `mutual` -> `mutable`

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
```diff
@@ -0,0 +1,57 @@
...
+pub fn create_view_pair<T: Default + Clone>(
```
siefkenj, line 8: -> `create_state_var_view`
The return type gives the fact that it's a pair.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
```diff
@@ -0,0 +1,57 @@
...
+        component_idx: 0,
+        origin: EssentialDataOrigin::StateVar(0),
+    };
+
+    let dependency = Dependency {
+        source: dependency_source,
+        value: essential_view.into(),
+    };
+
+    (dependency, essential)
```
siefkenj, line 34: The word `essential` still drives me nuts. This var isn't "an essential". It's a state var, so let's call it `state_var`.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
```diff
@@ -0,0 +1,57 @@
...
+    initial_value: T,
+) -> (Dependency, StateVarMutableView<T>)
+where
+    // make sure T is one of the types that we can convert from StateVarView<T> into StateVarViewEnum
+    StateVarViewEnum: From<StateVarView<T>>,
+{
+    let (state_var, state_var_view) = create_view_pair(initial_value, false);
+
+    let dependency_source = DependencySource::StateVar {
+        component_idx: 0,
```
siefkenj, line 47: Literal indices scare me.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
```diff
@@ -29,13 +29,26 @@ pub fn create_data_query_if_match_extend_source(
...
+pub fn string_to_boolean(s: &str, empty_is_true: bool) -> bool {
+    s.eq_ignore_ascii_case("true") || (empty_is_true && s.is_empty())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_string_to_boolean() {
```
siefkenj, line 44: The test names should be a little more descriptive. I would probably break this into a couple of tests: `fn string_to_boolean_ignores_case` and `fn string_to_boolean_accepts_empty_strings`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
```diff
@@ -29,13 +29,26 @@ pub fn create_data_query_if_match_extend_source(
...
-/// TODO: the empty string is try so that attribute="" sets the boolean attribute to true.
-/// Are there other cases where we'd want the empty string to be false?
-/// For example, an empty `<textInput>` becomes true with this algorithm,
-/// which may be confusing for authors. Not sure how to treat that case differently.
-pub fn string_to_boolean(s: &str) -> bool {
-    s.eq_ignore_ascii_case("true") || s.is_empty()
+/// The word "true" (case-insensitive) becomes `true`.
+/// If `empty_is_true` is `true`, then an empty string also becomes `true`.
+/// All other strings become `false`.
+pub fn string_to_boolean(s: &str, empty_is_true: bool) -> bool {
```
siefkenj, line 35: Haven't seen the use case yet, but I don't like the `empty_is_true` parameter.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
```diff
@@ -0,0 +1,545 @@
...
+        ]
+    );
+
+    // Note: from this point on, at the level of these tests where dependencies are specified manually,
+    // there is no difference in this state variable as the above ones with children,
+    // so we don't repeat the same steps
+}
+
+/// Testing the case of string state variable
+/// where we pass in a particular data query
```
siefkenj, line 515: This isn't a good description. A good description will tell the future person who breaks the test why that test was there (Since future refactors may mean that some tests are not longer relevant and can be deleted).

Since I don't understand what this functionality is, I cannot suggest a good docstring, but maybe something like
```
When created, `StringStateVar`s specify where their values come from.
One possibility is that they come from another state var, which is sepcified
via a data query. 
...
```
Of course, some of this may be already written in the docstring of `StringStateVar`, so you could leave out much of the explanation and reference the corresponding doc. See https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
dqnykamp: I like the derived versus essential/root/primitive distinction when dealing with the DAG of dependencies in core, as it is fundamental to how the DAG updates work. However, since we are trying to make the component development isolated from these complications, I see how it doesn't make sense to introduce such concepts in this domain.

There are two reasons why one would get "essential" variables in a state variable updater.
1. You specifically want to make the state variable depend on a local variable that doesn't depend on anything. Maybe for this case, you ask for a `DataQuery::LocalVariable` or something like that. (Currently, you are allowed only one of these per state variable and it will always have the type of your state variable, so not sure if there is a better name to communicate that.)
2. You ask for `DataQuery::Child` or `DataQuery::AttributeChild` and you get no results. So, you need somewhere to store your value if it were changed via `invert`. I had previously make these data queries return an single "essential" variable behind the scenes when there were no children in order to hide this detail from component authors. I recently removed that feature as it ended up getting confusing when to you had multiple data queries so that you had to ignore one of these automatically added variables. I'll think how I can bring that feature back while addressing that confusion, as it'd be nice to have this case work automatically without having component authors to code special cases.

Both of these cases were further complicated by the possibility of an `ExtendSource` that could create another dependency and require component developers to make a special case with handling these "essential" variable. I'll see if I can come up with a way to handle that automatically inside these other data queries so component authors don't have to deal with it, such as how you were suggesting earlier. (It will be a challenge to know what the automatic answer should be if, for example, there is both a `DataQuery::Child` and a `DataQuery::AttributeChild` in a single state variable.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
dqnykamp: This is vague because I don't yet understand what we'd use `StringStateVar::new()` for either! We aren't using it anywhere, but I didn't feel ready to delete it in case we come with use for it soon.

Maybe rather than making this vague placeholder test, I'll just put a `TODO` in the source to either use the feature (and make a test that illustrates how it is used) or delete the feature. (If we want to make a string variable from a state variable, maybe we'll want a `StringStateVar::new_from_state_variable()` rather than this? It's not yet clear if allowing a generic data query is useful.)

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
siefkenj: I think I understand about 50%. Would `DataQuery::LocalVariable` be the same as `DataQuery::Literal(foo)`, where that would always just return the value `foo`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
dqnykamp: For an attribute, like `<textInput disabled>`, we want the empty string that ends up getting created for `disabled`  to become `true`. For `<boolean>$t</boolean>`, if `$t` resolves to an empty string, we want the result to be false. So, I made a flag to determine if we are in an attribute, which changes the behavior of `string_to_boolean`.

The question is: do we want `<textInput disabled="" />` and `<textInput disabled="$t">`, where `$t` resolves to an empty string to make disabled be `true`? It is currently `false` in version 0.6. I was going to allow it to be `true` given our approach for handling empty attributes. I guess I could shop that around. If we want to maintain the behavior of version 0.6, then we need a different solution which I think will involve parsing empty attribute differently.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
dqnykamp: If you want a test at this level without `core`, we need to specify values for these literals, as the interface between `core` and state variables is defined by these indices. I did end up removing the literal indices in the code that component developers write, but core's representation of components and state variables is in terms of such indices. Integration tests involving code from `core` could presumably avoid such indices.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
siefkenj: If we don't see a use case and it may be deleted, there probably shouldn't be a test for it.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
siefkenj: I suggest a different function if you want the other behavior. But, really, you have a `attr_to_boolean` function, which treats the empty attr differently than other cases.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
dqnykamp: Not exactly, it's more like a "precursor" variable that a state variable developer can modify before calculating the actual value of a state variable. And then, the state variable developer would need to invert those calculations in `invert`.

In simple cases, such as for `sync_immediate_value` for `text_input`, no processing is done and the value is just passed through. (I need to create a general state variable case for that given that there is no logic to program there.)

But, in more complicated cases, such as `immediate_value`, that local value is just one piece of information needed to calculate the value. For example, the default from the local value is overridden by `prefill`.

I'm not really happy about the name. I'm not sure if `DataQuery::LocalValue` is better. It's certainly not static as it will typically be changed in `invert`. Although there's no way to enforce any behavior, the idea is that it is closely related to the value of `foo`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
dqnykamp: When you do those links to other docs, does it work for you in rust-analyzer? Mine just wants to send me to URLs.

packages/doenetml-worker-rust/lib-doenetml-core/src/utils/test_utils.rs
dqnykamp: Decided to call it `DataQuery::PreliminaryValue` for now.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.rs
dqnykamp: reworked this

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
dqnykamp: broke into multiple tests

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/util.rs
dqnykamp: have separate functions now

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -254,74 +254,30 @@ pub trait ComponentActions: ComponentNode {
...
-    /// Used to create the dependency matching ComponentProfile of a child data query.
-    ///
-    /// In this way, the state variable depending on the children can calculate its value
-    /// from the state variable value of the ComponentProfileStateVariable.
-    pub fn into_state_variable_view_enum_and_idx(self) -> (StateVarViewEnum, StateVarIdx) {
+// TODO: implement with macro?
+impl ComponentProfile {
+    /// Return the default value that is associated with the type of state variable
+    /// represented by the component profile.
+    pub fn get_default_value(&self) -> StateVarValue {
```
siefkenj, line 274: Could call it `default`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/component.rs
```diff
@@ -37,7 +37,7 @@ pub struct TextInput {
 
 impl ComponentAttributes for TextInput {
     fn get_attribute_names(&self) -> Vec<AttributeName> {
-        vec!["bindValueTo", "hide", "disabled", "prefill"]
+        vec!["hide", "disabled", "prefill"]
```
siefkenj, line 40: Yay. Confusing attr removed :-D

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.test.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    // should return a query for text or boolean children
+    assert_eq!(
+        queries,
+        vec![DataQuery::Child {
+            match_profiles: vec![ComponentProfile::Text, ComponentProfile::Boolean],
+            exclude_if_prefer_profiles: vec![],
+            always_return_value: true,
+        },]
+    );
```
siefkenj, line 105: This is the same as above, right? We could split this off into its own test, and then only do it once.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.test.rs
```diff
@@ -0,0 +1,423 @@
...
+    state_var.save_dependencies(&dependencies_created_for_data_queries);
+
+    state_var.calculate_and_mark_fresh();
+
+    assert_eq!(*state_var.get(), false);
+}
+
+/// Testing the case of boolean state variable is case insensitive when has string children.
+#[test]
+fn case_insensitive() {
```
siefkenj, line 185: -> `can_convert_string_to_boolean_in_case_insensitive_way()`

Tests are a place where you have have long function names ;-). More specifically, it's helpful when you see which tests failed, if you know what might be going wrong, rather than having to refer to the source of the test.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    // now the value should be "Bye Earth"
+    assert_eq!(state_var.get().clone(), "Bye Earth");
+}
+
+/// Testing the case of string state variable based on an attribute
+/// For example, if we have a `<textInput prefill/>`
+/// its `prefill` state variable would follow this pattern.
+#[test]
+fn string_state_var_with_attribute() {
```
siefkenj, line 160: I don't understand what this is supposed to test. Is it that correct data queries are returned for a string state var?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
```diff
@@ -0,0 +1,209 @@
...
+    let dependencies_created_for_data_queries =
+        vec![DependenciesCreatedForDataQuery(vec![child_dependency])];
+
+    state_var.save_dependencies(&dependencies_created_for_data_queries);
+
+    state_var.calculate_and_mark_fresh();
+
+    // we expect a value of "hello" and came_from_default to be false,
+    // since that the variable is coming from the child dependency, which we initialized to "hello".
+    assert_eq!(state_var.get().clone(), "");
```
siefkenj, line 207: Why the clone? Can you use a `&` instead?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
dqnykamp: That'd make it rustier!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.test.rs
dqnykamp: Yes, I suppose that would make sense. I'm still thinking in terms of integration or end-to-end tests where a test does a whole sequence of things.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
dqnykamp: It seems like we can use `*` instead.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/boolean_state_var.test.rs
dqnykamp: tried to do that!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/string_state_var.test.rs
dqnykamp: Hope it is clearer now

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/mirror_state_var.rs
```diff
@@ -0,0 +1,71 @@
...
+/// The state variable has a single data query, which is for the value of another state variable.
+/// It takes on the value of this other state variable,
+/// and when inverting, requests that the value of this other variable be changed.
+#[derive(Debug, Default)]
+pub struct MirrorStateVar {
+    mirrored_state_var_idx: StateVarIdx,
+}
+
+impl MirrorStateVar {
+    pub fn new(mirrored_state_var_idx: StateVarIdx) -> Self {
```
siefkenj, line 28: I'm torn here between following Rust's convention and our own. Rust convention is to have `new` take no arguments. So this would be called something like `with_index` or `from_index`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/mirror_state_var.rs
```diff
@@ -0,0 +1,71 @@
...
+where
+    T: Default + Clone + std::fmt::Debug,
+    StateVarView<T>: TryFromState<StateVarViewEnum>,
+    <StateVarView<T> as TryFromState<StateVarViewEnum>>::Error: std::fmt::Debug,
+{
+    fn return_data_queries(&self) -> Vec<Option<DataQuery>> {
+        MirrorRequiredDataQueries {
+            mirrored_value: Some(DataQuery::StateVar {
+                component_idx: None,
+                state_var_idx: self.mirrored_state_var_idx,
```
siefkenj, line 45: Is this always mirroring a state var on the same component? Or can this work cross-component?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/independent_state_var.rs
```diff
@@ -0,0 +1,70 @@
...
+    T: Default + Clone,
+{
+    preliminary_value: StateVarView<T>,
+}
+
+/// A state variable that doesn't depend on any data queries,
+/// other that its own preliminary value data query.
+///
+/// The state variable takes on its preliminary value,
+/// and when inverting, sets its preliminary value.
```
siefkenj, line 21: Why isn't it's `preliminary_value` called `value`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/independent_state_var.rs
```diff
@@ -0,0 +1,70 @@
...
+        IndependentRequiredDataQueries {
+            preliminary_value: Some(DataQuery::PreliminaryValue),
+        }
+        .into()
+    }
+
+    fn calculate(&self, data: &IndependentRequiredData<T>) -> StateVarCalcResult<T> {
+        // take on the value from `preliminary_value`, propagating `came_from_default`.
+        if data.preliminary_value.came_from_default() {
+            StateVarCalcResult::FromDefault(data.preliminary_value.get().clone())
```
siefkenj, line 53: This code comes up a lot. Can we add a `StateVarCalcResult::from` that just does this?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/mirror_state_var.rs
```diff
@@ -0,0 +1,71 @@
+#[cfg(test)]
+#[path = "mirror_state_var.test.rs"]
+mod tests;
```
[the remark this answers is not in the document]
dqnykamp: sounds good

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/mirror_state_var.rs
dqnykamp: I hope that eventually we could make this cross-component. We just need to determine an approach for getting the `component_idx` of the desired component into the state variable updater. Since it wasn't obvious to me how this will work yet, I left it intra-component until we find a use-case where we need the cross-component feature.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/mirror_state_var.rs
dqnykamp: It seems like `new` with arguments is OK:

From https://rust-lang.github.io/api-guidelines/predictability.html:
> The name new should generally be used for the primary method of instantiating a type. Sometimes it takes no arguments, as in the examples above. Sometimes it does take arguments, like [Box::new](https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.new) which is passed the value to place in the Box.

PR #111, opened 2024-02-04
Normalization of references

siefkenj:
This PR adds a normalization function that matches all macros with their referents and then replaces them with `Element` nodes.

This PR also introduces an "untagged" `FlatDast`. This format is more uniform in that attributes and elements can have the same types of children. This means macros in attributes can be expanded in the same way as macro children. However, it means that re-serialization isn't so simple (but we probably won't be reserializing this format except in tests).

This PR currently:
 - Expands macros `$f` to `<foo ... />` where `foo` is the element they refer to, or to an error node if there is no unique referent.
 - Expands function macros `$$f(x)` to `<evaluate ...><ol><li>x</li></ol></evaluate>` where the inputs to the function are put into an ordered list.

In both of the above cases, a special prop `extending` contains information about the referent.

This PR also:
 - Consumes any `extend="$foo"` attribute and and sets the `extending` prop to the referent (if the referent can be resolved). If the referent cannot be resolved, or if the `extend` attribute is invalid, an error node is inserted as a child of the element.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.rs
```diff
@@ -0,0 +1,153 @@
...
+impl Expander {
+    /// Expand all macros and function macros into their "xml" form. After expansion,
+    /// the resulting tree may not be serializable as XML since element attributes may contain
+    /// other elements.
+    pub fn expand(flat_root: &mut FlatRoot) {
+        let resolver = Resolver::from_flat_root(flat_root);
+
+        for idx in 0..flat_root.nodes.len() {
+            // The original `nodes[idx]` node is being completely replaced, so we are free to take its value,
+            // which will prevent the borrow checker form complaining if we mutate `flat_root` during processing.
```
dqnykamp, line 34: form ->  from

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
```diff
@@ -0,0 +1,240 @@
...
+                $x.y
+            </b>
+            <f>
+                <e name="w" />
+            </f>
+        </a>
+        $w
+        <d name="y" />"#
+            .replace("\n", "")
+            .replace("  ", ""),
```
dqnykamp, line 69: I'm curious why you put these replaces in here. Can't the `dast_root_no_position()` function handle white space?

dqnykamp: Oh, it's probably for the comparison to the `.to_xml()` output, below.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
```diff
@@ -0,0 +1,240 @@
...
+    );
+    let mut flat_root = FlatRoot::from_dast(&dast_root);
+    Expander::expand(&mut flat_root);
+    assert_eq!(
+        r#"<document><a name="x"><b name="y"><c name="z" /><b /></b><f><e name="w" /></f></a><e /><d name="y" /></document>"#,
+        flat_root.to_xml()
```
dqnykamp, line 75: What is the desired behavior of the `.to_xml()` function? It is supposed to encode what core will try to do? Because, core copies the children when extending a component (as opposed to a state variable). Should the `.to_xml()` indicate that the second `<b />` actually will end up getting children?

(In this example, core will run into a circular dependence, as the children of the `<b />` created from the reference will contain a copy of itself. It currently hangs the browser as we don't have circular dependence checks.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
```diff
@@ -0,0 +1,240 @@
+use assert_json_diff::assert_json_eq;
+
+use super::*;
+use crate::test_utils::*;
+
+#[test]
+fn macros_get_expanded_to_their_referents() {
```
dqnykamp, line 7: I'd like to write a bunch more tests nailing down the edge cases of this expander to make sure we have the behavior well specified. Would you want them in this PR? I can try to think through different edge cases.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
dqnykamp: Oh, it looks like you have more macro resolve tests, below.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
dqnykamp: Oh, it looks like you have more macro resolve tests, below.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
siefkenj: `to_xml` should only be used for testing. It's easier to see what some tests are trying to do in XML form rather than looking at bit data structures. I moved it to `test_utils` so that it cannot be invoked from core.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.test.rs
siefkenj: It's easier to see the layout indented, but those extra text nodes clutter up the output, so I preemtively removed them.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
```diff
@@ -0,0 +1,230 @@
+//! Macros use a `path`, which consists of _names_ separated by dots (e.g. `a.b.c`), possibly followed by
+//! state variable names separated by dots (e.g. `a.b.c.x.y.z`). A `Resolver` searches for a matching node
+//! from a given starting position. Because a `Resolver` does not know whether part of a path corresponds to a
+//! name or state variable name, the longest valid partial math is used and unmatched portions of a `path` are
```
dqnykamp, line 4: math -> match

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// Information about the referent that this element extends (e.g., as specified by the `extend` attribute).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extending: Option<RefResolution>,
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct FlatAttribute {
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub parent: Option<Index>,
```
dqnykamp, line 44: In what situation does a `FlatAttribute` not have a parent? Or is the `Option` there just to make processing easier by sharing the same format with other objects?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub path: Vec<PathPart>,
+    pub input: Option<Vec<Vec<UntaggedContent>>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub position: Option<Position>,
+    pub idx: Index,
+}
+
+/// Objects that can be store in the main `nodes` array of a `FlatRoot`.
```
dqnykamp, line 92: store -> stored

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// This function recursively adds all children etc. of the `DastRoot`.
+    pub fn merge_dast_root(&mut self, dast: &DastRoot) {
+        for child in dast.children.iter() {
+            self.merge_content(child, None);
+        }
+    }
+
+    /// Update the the node at `idx` to be `node`. Any children of `node` are added to the `FlatRoot` as well,
+    /// but children of the old node are not removed. This function does no consistency checking, so it is
+    /// the _caller should be extra careful_.
```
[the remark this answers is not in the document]
dqnykamp: Also, you don't check that a component at idx already exists, but that will just lead to a panic when trying to modify the entry in the nodes vector.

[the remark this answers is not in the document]
dqnykamp: It seems if the parent passed in does not match the parent of the component it is replacing, then we're in trouble.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
```diff
@@ -0,0 +1,383 @@
...
+                    })
+                    .collect();
+                self.set_children(idx, children);
+                self.set_attributes(idx, attributes);
+                ret
+            }
+            DastElementContent::Error(err) => {
+                let ret = UntaggedContent::Ref(self.set_error(err, idx, parent));
+                if parent.is_none() {
+                    self.children.push(ret.clone());
```
dqnykamp, line 231: This line should lead to duplication of a reference to the node. There must have been an element at `idx` already, or `set_error` would panic. Which means, if everything were consistent beforehand (and the replaced node also had no parent, as should be the case), the root should already have a reference to the index of this node. With this push, it should now have two references (and this new one is at the end of the document.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
```diff
@@ -0,0 +1,383 @@
...
+    pub fn parent_iter(&self, start_idx: Index) -> ParentIterator {
+        let start = &self.nodes[start_idx];
+        ParentIterator::new(start, self)
+    }
+
+    /// Merge the content of a `DastRoot` into `FlatRoot`.
+    /// This function recursively adds all children etc. of the `DastRoot`.
+    pub fn merge_dast_root(&mut self, dast: &DastRoot) {
+        for child in dast.children.iter() {
+            self.merge_content(child, None);
```
dqnykamp, line 172: If you collect these results into the `self.children` array here, you wouldn't have to special case `parent.is_none()` in `merge_content()`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
siefkenj: I put it there to be like the others. If you're creating a the data structure outside of a tree, it is also nice to be able to put `None` in there instead of make up some value.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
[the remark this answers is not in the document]
siefkenj: Yes. I've added notes about this. These are pretty low level functions and shouldn't be used often.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
siefkenj: Good catch!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast.rs
siefkenj: That's true, but if I set `flat_root.children` here, I wouldn't be able to do a `merge_content` with parent set to `None` and have them show up in Root. I don't know if I ever need that functionality though...(maybe it's better to not have it since passing an otherwise invalid parent does nothing?)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast_compactify.rs
```diff
@@ -0,0 +1,184 @@
...
+            .iter()
+            .filter_map(|node| match node {
+                UntaggedContent::Ref(idx) => Some(*idx),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+        while let Some(idx) = to_visit.pop() {
+            // Don't visit a node's children more than once.
+            if is_referenced[idx] {
+                continue;
```
dqnykamp, line 106: What would cause this algorithm to visit a node more than once? Are you attempting to be able to handle loops in the graph? It might be better to panic in that situation.

At any rate, if I put a panic in here instead, the tests still pass. So if there is a reason for this test, it's not covered by the tests.

siefkenj: Yes. The idea was to prevent loops from breaking the code. I would rather have this code as loop-tolerant even if it's functionality we don't need.

dqnykamp: Sounds good. Maybe add a comment to this effect?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast_merge.rs
```diff
@@ -0,0 +1,229 @@
...
+        match &mut self.nodes[idx] {
+            FlatNode::FunctionMacro(func_macro) => {
+                func_macro.input = Some(args);
+            }
+            _ => panic!("set_function_macro_args called on non-function-macro node"),
+        }
+    }
+
+    /// Set the node at position `idx` to be the specified element.
+    /// The element will be initialized with empty children and attributes.
```
dqnykamp, line 176: Maybe, since all you are using is the name and position:
/// Set the node at index `idx` to be an element with the name and position of the specified element.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/flat_dast/untagged_flat_dast_merge.rs
```diff
@@ -0,0 +1,229 @@
...
+            parent,
+            idx,
+            // It is impossible to directly set `extending` in DAST; it is computed later.
+            extending: None,
+        });
+        idx
+    }
+
+    /// Set the node at position `idx` to be the specified error.
+    fn set_error(&mut self, node: &DastError, idx: Index, parent: Option<Index>) -> usize {
```
dqnykamp, line 192: It's a little confusing to use "position" to describe `idx`, when "position" means something else in the function.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// - Matching `x` from anywhere returns the index of `<a />`.
+    /// - Matching `y` from `<c />` returns the index of `<b />`.
+    /// - Matching `y` from `<d />` returns an error because there are multiple `y`s accessible from that position.
+    /// - Matching `y.z.w` from `<b />` returns the index of `<c />` along with `.w` as the unresolved path.
+    ///
+    /// If an index appears in the path (e.g. `a[2].b`), the search stops and the remaining path is returned.
+    /// If there was a partial match of the indexed item, the unresolved path will list `name` as an empty string.
+    /// E.g., matching `y.w[2]` from `<b />` returns the index of `<d />` along with `.w[2]` as the unresolved path
+    /// and matching `y[2]` from `<b />` returns the index of `<d />` along with `.[2]` as the unresolved path.
+    pub fn resolve<T: AsRef<[PathPart]>>(
```
dqnykamp, line 81: What are we gaining by making the argument `AsRef<[PathPart]>` rather than `&[PathPart]`? I read some of the documentation `AsRef`, but couldn't figure out the benefit. It looks like it is current called on `&Vec<PathPath>`, for which the reference to the slice would work.

siefkenj: `AsRef` allows passing anything that can be converted into a reference, so it allows both `Vec` and `&Vec`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
```diff
@@ -0,0 +1,230 @@
...
+impl Resolver {
+    pub fn from_flat_root(flat_root: &FlatRoot) -> Self {
+        Resolver {
+            node_parent: flat_root.nodes.iter().map(|node| node.parent()).collect(),
+            name_map: Self::build_name_map(flat_root),
+        }
+    }
+
+    /// Search for a node specified by `path` starting from `origin`. This algorithm searches first
+    /// for the nearest parent that has an element with `path[0].name` as its name. Then it tries
```
dqnykamp, line 59: I was confused by the phrase "has an element with". I first read that the parent element had to have that name, which I know isn't correct. Maybe "has a descendant element with"?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
dqnykamp: OK, like the example with a function taking both`String` and `&str`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
```diff
@@ -0,0 +1,230 @@
...
+            return Ok(RefResolution {
+                node_idx: current_idx,
+                unresolved_path: Some(remaining_path),
+            });
+        }
+
+        while let Some(part) = path.next() {
+            // current_idx specifies the "root" of the search. We try to resolve
+            // children based on the path part, breaking if any of them are not found or if
+            // there is an `index` specified.
```
dqnykamp, line 112: and returning an error if, at some point, there is ambiguity in the possible referent of a path part.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.rs
```diff
@@ -0,0 +1,230 @@
...
+                    }
+                    Ref::Ambiguous(_) => {
+                        return Err(ResolutionError::NonUniqueReferent);
+                    }
+                }
+                // If there index specified, we immediately stop since component information is needed
+                // to resolve all remaining path parts.
+                if !part.index.is_empty() {
+                    let remaining_path: Vec<PathPart> =
+                        iter::once(part.clone()).chain(path.cloned()).collect();
```
dqnykamp, line 126: So we have different behavior for `$x[0]` and `$x.y[0]`? For `$x[0]`, we match `x` and have remaining path be `""[0]`.  For `$x.y[0]`, it seems we don't match `y` but just match `x` and have the remaining path be `y[0]`. Is there a reason we don't match `x.y` and have the remaining path part be `""[0]`. That seems more consistent to me, but I'm not confident that I'm not missing something.

siefkenj: The reason is me not thinking clearly :-P. I even hardcoded this in a test...but I think the behavior should be consistent (especially if we allow indexing of items directly rather than just indexing of state variables)

siefkenj: Fixed, and a test is added.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.test.rs
```diff
@@ -0,0 +1,247 @@
+use super::*;
+use crate::{
+    dast::{DastIndex, DastText, DastTextMacroContent},
+    test_utils::*,
+};
+
+fn find_node_index_by_name(flat_root: &FlatRoot, name: &str) -> Option<Index> {
```
dqnykamp, line 7: Move these utility functions to the bottom of the file? :)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_resolve.test.rs
```diff
@@ -0,0 +1,247 @@
...
+    let dast_root = dast_root_no_position(
+        r#"<a name="x">
+            <b name="y">
+                <c name="z" />
+            </b>
+        </a>
+        <d name="y" />"#,
+    );
+    let flat_root = FlatRoot::from_dast(&dast_root);
+    let b_idx = find_node_index_by_name(&flat_root, "b").unwrap();
```
dqnykamp, line 18: The terminology clashes for "name" and "type" in DoenetML versus the dast is hurting here. This line is confusing since the above DoenetML make it seem that the names are "x", "y" and "z". In DoenetML, "name" become "component type". Since the DoenetML terminology is more front and center in the context of this test, maybe `find_node_index_by_component_type()`?

siefkenj: How about I just change it to `find`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.rs
```diff
@@ -0,0 +1,255 @@
...
+                        Err(err) => FlatNode::Error(FlatError {
+                            idx: macro_.idx,
+                            parent: macro_.parent,
+                            message: format!("Macro resolution error: {}", err),
+                            position: macro_.position.clone(),
+                        }),
+                    }
+                }
+                FlatNode::FunctionMacro(function_macro) => {
+                    // A function ref `$$f(x)` becomes `<evaluate extend="f"><ol><li>x</li></ol></evaluate>`
```
dqnykamp, line 73: Currently, the evaluate syntax is 
`<evaluate function="$f" input="x" />`
or
`<evaluate function="$f" input="x y" />`
where the `input` attribute gets expanded into a list. I don't know to what extent people use the `<evaluate>` tag, but if they did, I think it would be unintuitive to put the lists as children.

It's true that one cannot enter a list directly into an attribute, but we have sugar that translates a space separated list implemented on the list component (at least on the `<mathList>` component, not currently on the `<li>` component).

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/macro_expand.rs
```diff
@@ -0,0 +1,255 @@
...
+                                        .collect(),
+                                    position: function_macro.position.clone(),
+                                    data: None,
+                                };
+                                // Insert the `ol` into `flat_root`
+                                let ol = flat_root.merge_content(
+                                    &DastElementContent::Element(dast_ol),
+                                    Some(idx),
+                                );
+                                // The `<li>` tags the exclusive children of the `<ol>` tag.
```
dqnykamp, line 108: "tags the exclusive" -> "tags are the exclusive"

PR #113, opened 2024-02-08
Redo Component Creation

siefkenj:
This PR uses the newly-merged normalized DAST for component creation.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,95 @@
...
+                            .get_attribute_names()
+                            .iter()
+                            .map(|&name| (name, Vec::new())),
+                    );
+                    component.initialize(
+                        elm.idx,
+                        elm.parent,
+                        elm.extending
+                            .clone()
+                            .map(|ref_resolution| ExtendSource::Component(ref_resolution.idx())),
```
dqnykamp, line 59: `ExtendSource::Component` is the correct thing to create if there is no remaining path. However, if there is a remaining path, then you'll need to create an `ExtendSource::StateVar`, assuming there is a match to the remaining path. If there isn't a match, then the component should be replaced with an error.

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
```diff
@@ -4,14 +4,28 @@ use proc_macro2::{Ident, Span};
...
+    pub ref_transmutes_to: Option<String>,
+}
+
 /// Implement the ComponentNode trait structs
 /// assuming they have have a common field that is `ComponentCommonData`
 pub fn component_node_derive(input: TokenStream) -> TokenStream {
     let ast: syn::DeriveInput = syn::parse(input).unwrap();
     let name = &ast.ident;
     let data = &ast.data;
+    let options: ComponentOptions = FromDeriveInput::from_derive_input(&ast).unwrap();
```
dqnykamp, line 28: Nice, you found a much better way to get the attributes for a derive macro!

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
```diff
@@ -22,8 +36,7 @@ pub fn component_node_derive(input: TokenStream) -> TokenStream {
                     n => n.to_case(Case::Camel),
                 };
 
-                quote! {
-                    impl ComponentNode for #name {
+                let mut component_node_impl_body = quote! {
                         fn get_idx(&self) -> ComponentIdx {
```
dqnykamp, line 40: These lines got indented twice, and the formatter doesn't work inside `quote!` to fix it.

packages/doenetml-worker-rust/lib-rust-macros/src/component_node.rs
siefkenj: Yeah. The `darling` package looks pretty cool :-)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+                        // should become `<textInput name="i" /><text extend="$i` />`.
+                        // This information is stored in `ref_transmutes_to()`.
+                        if let Some(name) = component.ref_transmutes_to() {
+                            // It is forbidden to expand to an invalid component type, so we do not
+                            // have a fallback to `_External` here.
+                            component = ComponentEnum::from_str(name).unwrap();
+                        }
+                    }
+                    // Some references
+                    component.ref_transmutes_to();
```
dqnykamp, line 61: delete

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+                            name: elm.name.clone(),
+                            ..Default::default()
+                        })
+                    });
+                    if let Some(Source::Macro(_)) = &elm.extending {
+                        // Some components specify that when they are referenced with the `$foo` syntax,
+                        // a different component should be created in their place. E.g., `<textInput name="i" />$i`
+                        // should become `<textInput name="i" /><text extend="$i` />`.
+                        // This information is stored in `ref_transmutes_to()`.
+                        if let Some(name) = component.ref_transmutes_to() {
```
dqnykamp, line 54: Only do this is there is no remaining path. `$foo.x` should not transmogrify.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+                            Some(ref_resolution) => {
+                                let ref_resolution = ref_resolution.get_resolution();
+                                // If there is no remaining path, we are extending a component directly. Otherwise,
+                                // we look up the state variable on that component and extend it.
+                                if let Some(unresolved_path) = &ref_resolution.unresolved_path {
+                                    if unresolved_path.len() != 1 {
+                                        Err(anyhow!("Handle nested state variables"))
+                                    } else {
+                                        // We only know how to handle single state variable access at the moment.
+                                        if let Some(state_var) = component
```
dqnykamp, line 82: Use `.get_public_state_variable_index_from_name_case_insensitive()` as we can match only public state variables and paths are matched in a case-insensitive manner.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+                                            .get_state_variable_index_from_name(
+                                                &unresolved_path[0].name,
+                                            )
+                                        {
+                                            Ok(Some(ExtendSource::StateVar(
+                                                ExtendStateVariableDescription {
+                                                    component_idx: ref_resolution.node_idx,
+                                                    state_variable_matching: vec![
+                                                        StateVariableShadowingMatch {
```
dqnykamp, line 91: See function `match_public_state_variable()`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+                                                            // XXX: THIS 0 IS MADE UP AND IS A PLACEHOLDER UNTIL WE HAVE AN ACTUAL METHOD
+                                                            shadowed_state_var_idx: 0,
+                                                            shadowing_state_var_idx: state_var,
+                                                        },
+                                                    ],
+                                                },
+                                            )))
+                                        } else {
+                                            Err(anyhow!(
+                                                "State variable {} not found on component {}",
```
dqnykamp, line 101: The name could be "" if we stopped due to a path index. Probably want a different message in that case.

dqnykamp: Also, if there is a path index after a matching state variable, do we want to `Err` for now?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -76,11 +82,11 @@ pub struct ComponentCommonData {
     ///
     /// Macros remain in this vector only if they couldn't be expanded.
```
dqnykamp, line 83: Delete obsolete comment

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -76,11 +82,11 @@ pub struct ComponentCommonData {
...
     /// TODO: implement function macros so they don't stay in this vector
-    pub children: Vec<ComponentPointerTextOrMacro>,
+    pub children: Vec<UntaggedContent>,
```
dqnykamp, line 85: It doesn't need to change for this PR, but I think `UntaggedContent` is a confusing type, as it isn't clear how it used to be tagged and how this is different. I'm not sure of a better word.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -118,14 +124,11 @@ pub trait ComponentNode: ComponentState {
...
+    fn get_children(&self) -> &Vec<UntaggedContent>;
     /// Set the vector containing the indices of all child component nodes and the literal string children.
-    fn set_children(&mut self, children: Vec<ComponentPointerTextOrMacro>);
+    fn set_children(&mut self, children: Vec<UntaggedContent>);
     /// Replace with new values the vector containing the indices of all child component nodes and the literal string children.
-    fn replace_children(
-        &mut self,
-        new_children: Vec<ComponentPointerTextOrMacro>,
-    ) -> Vec<ComponentPointerTextOrMacro>;
+    fn replace_children(&mut self, new_children: Vec<UntaggedContent>) -> Vec<UntaggedContent>;
```
dqnykamp, line 131: Could delete `replace_children` from trait as it is no longer needed with the `ComponentBuilder`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,361 @@
...
+            Creating,
+            /// A component has been created at this index.
+            Created,
+        }
+        let mut component_status: Vec<Status> = iter::repeat(Status::Empty)
+            .take(normalized_root.nodes.len())
+            .collect();
+
+        // Creating the nodes lowest-index first should lead to less queueing than the other way around.
+        let mut queue: Vec<usize> = Vec::from_iter((0..components.len()).rev());
```
dqnykamp, line 96: This queue currently isn't doing anything.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        while let Some(idx) = queue.pop() {
+            if components[idx].is_some() {
+                // No need to remake a component.
+                continue;
+            }
+            if matches!(component_status[idx], Status::Creating) {
```
dqnykamp, line 103: The component status currently isn't doing anything. The only check is here where it checks for `Creating`, but it can never happen given that the status is always changed to `Created` or `Empty` before continuing the loop. We can just remove it.

dqnykamp: For this reason, this check does not catch circular dependencies. We get "maximum call stack size" errors with either of these two DoenetMLs:
```
<text name="t1" extend="$t1" />
```
or
```
<text name="t1" extend="$t2" />
<text name="t2" extend="$t1" />
```

dqnykamp: Of course, the response to this circular dependency is arguably worse:
```
<text name="t1" extend="$t1.value" />
```
We don't run into trouble with this code, but the code for freshening dependencies gets into infinite recursion. Since we designed that code to handle a large stack, it grinds on for a long time before finally panicking.

siefkenj: Hmmm...okay. I'll check another way.

dqnykamp: Even though I can't trigger a circular dependence in this code, the check seems reasonable for when we add features that could lead to a circular dependence here.

PR #116, opened 2024-02-13
Typed Attributes

siefkenj:
Add a `AttributeStateVar` macro that creates appropriate state var creators for attributes.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -225,6 +225,42 @@ pub trait ComponentAttributes: ComponentNode {
...
+/// impl AttributeStateVar<bool> for Attrs {
+///   fn state_var(&self) -> StateVar<bool> {
+///     match self {
+///       Attrs::Disabled => BooleanStateVar::new_from_attribute("disabled", false).into_state_var(),
+///       _ => panic!("This attribute does not have a boolean state variable."),
+///     }
+///   }
+/// }
+/// ```
+pub trait AttributeStateVar<T: Default + Clone> {
```
dqnykamp, line 259: I'd like to see the `default` more obviously specified in the interface. Whenever you implement an attribute state variable, you have to decide on a default in order to handle the case when no attribute was specified on the component.

Maybe the method should be
```
  fn state_var(&self, default: T) -> StateVar<T>;
```
Then, the example implementations would use `default` rather than `""` and `false`.

dqnykamp: And alternative is that you specify the default on the attribute `enum`, and then it doesn't need to be specified when you create the state variable definition. (I'm on the fence when is the best time to specify defaults.) It looks you are specifying defaults in our `TextInputAttribute`, so I'm not yet sure how they are working together.

siefkenj: I'm not understanding your comment...Are you commenting on the docstring? The default is speficied on `TextInputAttribute` fields and nowhere else. A macro copies it to the correct location.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/component.rs
```diff
@@ -2,19 +2,23 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, AttributeStateVar)]
 pub enum TextInputAttribute {
     /// Whether the `<textInput>` should be hidden.
+    #[component_attribute(state_var = BooleanStateVar, default = false, explicit_type = bool)]
     Hide,
     /// Whether the `<textInput>` should be editable.
+    #[component_attribute(state_var = BooleanStateVar, default = false, explicit_type = bool)]
     Disabled,
     /// The content that should prefill the `<textInput>`, giving it a default value before a user has interacted with the input.
+    #[component_attribute(state_var = StringStateVar, default = "\"\"", explicit_type = String)]
```
dqnykamp, line 21: I'm confused why you are using the string containing quotes as the default. I'd assume we want the default to be a blank string.

siefkenj: The string containing quotes is the empty string...One layer of quotes is peeled off if present. I don't know why or how to change it...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
dqnykamp: That's right. I was thinking about implementing this manually. Now, I see you that you are implementing it via the macro and using the defaults from the macro attributes.

In this case, since you have the defaults defined in the attribute itself, I see why you don't want to have the default as an argument to the `state_var` method.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/component.rs
siefkenj: Would `String::new()` be easier to understand?

packages/doenetml-worker-rust/lib-rust-macros/src/component_attributes.rs
```diff
@@ -0,0 +1,173 @@
...
+    default: syn::Expr,
+    state_var: Ident,
+}
+
+impl VariantImpl {
+    /// Create a match case for this variant that looks like
+    /// ```ignore
+    /// TextInputAttribute::Hide => BooleanStateVar::new_from_attribute("hide", false).into_state_var(),
+    /// ```
+    fn to_match_case(&self, enum_name: &Ident) -> proc_macro2::TokenStream {
```
dqnykamp, line 29: I found the name of this method misleading, as the name made me think about matching the case of a string. Maybe `to_match_arm`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/text_input/component.rs
dqnykamp: Sure, I like `String::new()`, as it removes the confusion. Strange how it peels off quotes.

dqnykamp:
I'm liking this!

PR #125, opened 2024-02-14
Add math support

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -97,7 +97,7 @@ impl DoenetMLCore {
         Ok(result.into())
     }
     #[cfg(feature = "testing")]
-    pub fn parse_latex_into_math<'a, TXT: AsRef<str>, FnSymbol: AsRef<str>>(
+    pub fn parse_latex_into_math<TXT: AsRef<str>, FnSymbol: AsRef<str>>(
```
siefkenj, line 100: `TXT` is a weird choice...`Latex` perhaps? or `Text`?

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+    "circ",
+    "$",
+    "%",
+];
+
+export var textToAst = new me.converters.textToAstObj({
+    appliedFunctionSymbols: appliedFunctionSymbolsDefault,
+});
+
+export function getTextToMathConverter({
```
siefkenj, line 191: The common name of the types of functions are `factories`. Perhaps `textToMathFactory`?

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+        return (x: string) =>
+            me.fromAst(
+                new me.converters.latexToAstObj({
+                    appliedFunctionSymbols,
+                    functionSymbols,
+                    allowedLatexSymbols,
+                    parseScientificNotation,
+                }).convert(wrapWordWithVar(x, parseScientificNotation)),
+            );
+    }
```
siefkenj, line 251: `let wrapper = splitSymbols ? wrapWordIncludingNumberWithVar : wrapWordWithVar` and then you don't need to duplicate all this code.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+        ["\u03C9", "\\omega "], // 'ω'
+        ["\u2212", "-"], // minus sign
+        ["\u22C5", " \\cdot "], // dot operator
+        ["\u00B7", " \\cdot "], // middle dot
+        ["\u222A", " \\cup "], // ∪
+        ["\u2229", " \\cap "], // ∩
+        ["\u221E", " \\infty "], // ∞
+        ["\u2205", " \\emptyset "], // ∅
+        ["\u2032", "'"], // ′
+    ];
```
siefkenj, line 310: We already have a `static assets` folder with fancy mechanisms to compress large JSON structures. If you're going to need big lists of symbols, I suggest you move them there and then import them.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+
+    for (let sub of substitutions) {
+        latexString = latexString.replace(new RegExp(sub[0], "g"), sub[1]);
```
siefkenj, line 313: Yikes. Scary loop. https://stackoverflow.com/questions/15604140/replace-multiple-strings-with-multiple-other-strings might help. You can also precompile this regex as a `const` so it doesn't need to be generated every time.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    let startLdotsMatch = latexString.match(
+        /^(\\ )*(\\ldots|\.(\\ )*\.(\\ )*\.)(\\ )*(.*)$/,
```
siefkenj, line 317: For complicated regexes, please annotate with some examples of what they're supposed to match.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+    // replace [space]or/and/U[space]
+    // with latex
+    latexString = latexString.replace(/(\b|\\ )or(\b|\\ )/g, "$1\\lor$2");
+    latexString = latexString.replace(/(\b|\\ )and(\b|\\ )/g, "$1\\land$2");
+
+    if (unionFromU) {
+        latexString = latexString.replace(/(\b|\\ )U(\b|\\ )/g, "$1\\cup$2");
+    }
+
+    return latexString;
```
siefkenj, line 351: I do have a whole library, `unified-latex` that does latex parsing. It might make the bundle size bigger, but it would certainly be more robust than regexes.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
```diff
@@ -0,0 +1,521 @@
...
+}
+
+function wrapWordWithVar(string: string, parseScientificNotation: boolean) {
```
siefkenj, line 354: Docstrings, please

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
```diff
@@ -18,9 +24,125 @@ export function evalWithMathExpressionsInScope(source: string) {
...
+        splitSymbols: splitSymbols,
+        parseScientificNotation: parseScientificNotation,
+        functionSymbols,
+    });
+    let expression;
+    try {
+        expression = fromText(text);
+    } catch (e) {
+        expression = me.fromAst("\uFF3F"); // long underscore
+        console.log("Invalid value for a math of text format: " + text);
```
siefkenj, line 56: `console.warn`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,273 @@
...
+    ///   will be interpreted as apply that function to the arguments (rather than multiplication)
+    pub fn from_latex<TXT: AsRef<str>, FnSymbol: AsRef<str>>(
+        latex: TXT,
+        split_symbols: bool,
+        function_symbols: &[FnSymbol],
+    ) -> Self {
+        let s = latex.as_ref();
+
+        let math_object =
+            match DoenetMLCore::parse_latex_into_math(s, split_symbols, function_symbols) {
```
siefkenj, line 79: Is there a reason we're putting these functions on Core?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,273 @@
...
+    Integer(i64),
+    String(String),
+}
+
+/// Parameters for creating a latex string from a `Math`:
+#[derive(Debug)]
+pub struct ToLatexParams {
+    /// If present, then pad numbers with zeros so they have at least
+    /// this many decimal places after the decimal point displayed.
+    pub pad_to_decimals: Option<i32>,
```
siefkenj, line 204: Negatives are allowed? `u32`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,273 @@
...
+}
+/// Levels of simplification of mathematical expressions.
+#[derive(Debug, Display, Default)]
+#[strum(serialize_all = "camelCase")]
+pub enum MathSimplify {
+    /// No simplification is performed.
+    #[default]
+    None,
+    /// Simplify numbers, such as simplify `1+1` to 2,
+    /// except do not change the order between numerical and non-numerical operands.
```
siefkenj, line 230: What is a non-numerical operand? Example, please.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
dqnykamp: I wish you told that to the version of me from years ago when I wrote this. I'll put in a TODO, as I don't currently remember. Had to do with something about enabling syntax like 1,2,3..., would be my guess.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
dqnykamp: Done

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
dqnykamp: used strategy from that article

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
dqnykamp: moved

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/doenet/math/state_vars.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// queries for children with the `Math` component profile.
+    #[is_public]
+    #[component_profile_state_variable]
+    value: StateVar<MathExpr>,
+
+    /// An alias to the `value` state variable.
+    ///
+    /// It is marked `is_public` so that it can be referenced in DoenetML via `.math`.
+    #[is_public]
+    math: StateVar<MathExpr>,
```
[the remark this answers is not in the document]
dqnykamp: It's a pattern I developed from the untyped JS code, where I'd put props like `text` and `math` on components as a way for other components to find the `text` or `math` representations of the a component. I guess it is obsolete with the `ComponentProfile` system we have now. Removing them might break existing DoenetML if they are public props, but I'll remove it and think about compatibility as we get further along. (And, actually, I can't find a `math` prop on the old `<math>` anyway.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_state_var/math_state_var.rs
```diff
@@ -1 +1,167 @@
...
+    state::types::math_expr::{MathExpr, MathParser},
+};
 
+use super::util::MathOrString;
+
+#[derive(Debug, Default)]
+pub struct MathStateVar {
+    /// The data query that indicates how the dependencies of this state variable will be created.
+    data_query: DataQuery,
+
```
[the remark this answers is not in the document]
dqnykamp: This section is still under construction.  I was imagining I'd put docstring, so the blank lines are spots for docstrings.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/math.ts
siefkenj: Okay. Maybe put a big message in this file that states that these functions were copied from Doenet0.6 and should be reworked. That way I know during review I can at least skip over them.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Not sure. Was just blindly following the prototype again! I'll move them off Core.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,224 @@
...
+                .to_string(),
+        )
+    }
+
+    /// Parsing a string into math using the `math-expressions` latex parser.
+    ///
+    /// Parameters:
+    /// - `latex`: the string to be parsed
+    /// - `split_symbols`: if true, split multi-character symbols that do not contain an number
+    ///   into the product of the characters.
```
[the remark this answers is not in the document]
dqnykamp: Yes, here's where I really wanted to put in a doctest, if only they worked with `wasm_bindgen`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// - `function_symbols`: a list of the symbols that will be treated as a function,
+    ///   i.e., one of these symbols followed by arguments in parentheses
+    ///   will be interpreted as apply that function to the arguments (rather than multiplication)
+    #[cfg(not(feature = "testing"))]
+    pub fn parse_text_into_math<TXT: AsRef<str>, FnSymbol: AsRef<str>>(
+        text: TXT,
+        split_symbols: bool,
+        function_symbols: &[FnSymbol],
+    ) -> Result<String, String> {
+        let js_function_symbols = JsValue::from(
```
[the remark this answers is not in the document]
dqnykamp: Yes, referencing crates that weren't activated from running those tests broke it. I finally got them working again addressing both ways in which we know the web features aren't available.
```
#[cfg(all(not(feature = "testing"), feature = "web"))]
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Right, we can round to a negative number of decimals, but we can't pad to a negative number.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Wasn't sure how to word this. I hope the example helps.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
```diff
@@ -18,9 +24,128 @@ export function evalWithMathExpressionsInScope(source: string) {
...
+        console.warn("Invalid value for a math of text format: " + text);
+    }
+
+    return JSON.stringify(expression, serializedComponentsReplacer);
+}
+
+/**
+ * Parse the string `latex` into math using the `math-expressions` latex parser.
+ * Return the `JSON.stringify`ed results.
+ */
```
siefkenj, line 65: Can you explain what the arguments are? You can make a list with `@argName` syntax. E.g.


```typescript
/**
 * @latex - the source string
 * @splitSymbols - ???
 ...
 */
```

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
```diff
@@ -18,9 +24,128 @@ export function evalWithMathExpressionsInScope(source: string) {
...
+        padToDigits,
+        showBlanks,
+    });
+}
+
+/**
+ * Create a new mathematical expression formed by substituting variables with new expressions
+ */
+export function substituteIntoMath(
+    mathObject: string,
```
siefkenj, line 110: It's not an object. Maybe `source`?

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
```diff
@@ -18,9 +24,128 @@ export function evalWithMathExpressionsInScope(source: string) {
...
+    for (const [key, value] of substitutions.entries()) {
+        subs_object[key] = JSON.parse(value, serializedComponentsReviver);
+    }
+
+    let newExpr = mathExpr.substitute(subs_object);
+
+    return JSON.stringify(newExpr, serializedComponentsReplacer);
+}
+
+enum Simplify {
```
siefkenj, line 126: Don't use typescript enums. Declare these as consts or as a const object (e.g., `const SIMPLIFY = {...} as const;`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+/// ```
+
+// TODO: add examples with createVectors and createIntervals
+// once have a way to demonstrate their effect
+#[derive(Debug, Default, Clone, Copy)]
+pub struct NormalizeParams {
+    /// We currently support four options for `simplify`:
+    /// - `MathSimplify::None`: no simplification
+    /// - `MathSimplify::NumbersPreserveOrder`: simplify numbers within the expression, such as simplify `1+1` to 2,
+    ///   except do not change the order between numerical and non-numerical operands (such as a variable).
```
siefkenj, line 406: Example, please?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// - `MathSimplify::Numbers`: simplify numbers within the expression, such as simplify `1+1` to 2,
+    /// - `MathSimplify::Full`: simplify the mathematical expression using the currently
+    ///   implemented features. These features are limited and subject to change.
+    ///   For example, simplification of ratio expressions is essentially non-existent.
```
siefkenj, line 410: I think it's okay to have a `see [MathSimplify]` here and consolidate the docs.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+            Err(_) => BLANK_MATH_OBJECT.to_owned(),
+        };
+
+        MathExpr { math_object }
+    }
+}
+
+/// An enum of different types of values that can be interpreted as mathematical expressions.
+#[derive(Debug)]
+pub enum MathOrPrimitive {
```
siefkenj, line 277: I don't like names that spell out the type, in general. Ideally the name should reflect the objects use and properties, since we can always figure out the type via a mouse hover.

What about `Math`? You already have `MathExpr` for something else, otherwise that might be a good name...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
+use std::collections::HashMap;
+use strum_macros::Display;
+
+use crate::math_via_wasm::{
+    eval_js, math_to_latex, normalize_math, parse_latex_into_math, parse_text_into_math,
+    substitute_into_math,
+};
+
+const BLANK_MATH_OBJECT: &str = "{\"objectType\":\"math-expression\",\"tree\":\"\u{ff3f}\"}";
+/// A mathematical expression represented as a serialized `math-expressions` object.
```
siefkenj, line 10: Blank line above.

This seems like the most important item in this module. Could the docstring reflect that?

Like: `A mathematical symbolic mathematical expression allowing variables etc.. Operations on this object are carried out by the `math-expressions` JavaScript library, which allows for symbolic calculations like simplification. ... Data in this object is stored as a serialized string which is deserialized to preform operations. This may have performance implications and may be replaced with a rust-based symbolic calculator in the future`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+        serializer.serialize_str(&self.math_object)
+    }
+}
+
+impl MathExpr {
+    /// Create a `Math` by parsing the string `text` using the `math-expressions` text parser.
+    ///
+    /// Parameters:
+    /// - `text`: the string to be parsed
+    /// - `split_symbols`: if true, split multi-character symbols that do not contain an number
```
siefkenj, line 44: Finally, a description :-D. I think `single_char_symbols` would be a better name.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+    ///   into the product of the characters.
+    /// - `function_symbols`: a list of the symbols that will be treated as a function,
+    ///   i.e., one of these symbols followed by arguments in parentheses
+    ///   will be interpreted as apply that function to the arguments (rather than multiplication)
+    ///
+    /// Examples:
+    /// - `MathExpr::from_text("xy", true, &["f"])` is `x` times `y`.
```
siefkenj, line 51: All your examples pass `&["f"]`, which makes it look mandatory.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,429 @@
...
+
+        MathExpr { math_object }
+    }
+
+    /// Create a new mathematical expression by adding `term` to the current expression.
+    pub fn add(&self, term: MathOrPrimitive) -> MathExpr {
+        let term = serde_json::to_string(&term).unwrap();
+
+        let js_source = format!(
+            r#"JSON.parse('{}', serializedComponentsReviver).add(JSON.parse('{}', serializedComponentsReviver));"#,
```
siefkenj, line 211: If someone uses a `'` in a math expression, does that break this component? This type of code makes me nervous about code injections. We should have a `sanatize` function to make sure that we're not going to break things with this string manipulation.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/views.rs
```diff
@@ -221,6 +221,31 @@ impl<T: Default + Clone> PropView<T> {
         Ref::map(inner, |v| v.get())
     }
 
+    pub fn prop_calc_result_calculated(&mut self) -> PropCalcResult<T> {
```
siefkenj, line 224: What do these functions do?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/updater.rs
```diff
@@ -91,7 +93,7 @@ pub trait PropUpdaterWithCache<T: Default + Clone>: std::fmt::Debug {
 
     /// Calculate the value of the prop from the currently cached query results.
     /// Results of this function will be cached, so local caching is not needed.
-    fn calculate(&self) -> PropCalcResult<T>;
+    fn calculate(&mut self) -> PropCalcResult<T>;
```
siefkenj, line 96: This `mut` does not make me happy...can we do this with interior mutability instead? (I could be talked out of this opinion, though...)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/updater.rs
```diff
@@ -11,14 +11,16 @@ use crate::{
 use super::PropView;
 
 /// The possible results of a call to `calculate`:
-/// - Calculated(T): the value was calculated to be T
-/// - FromDefault(T): the value T was determined from the default value
-/// - From(&PropView<T>): set both `value` and `came_from_default` from `PropView<T>`
+/// - `Calculated(val)`: the value was calculated to be `val`
+/// - `FromDefault(val)`: the value `val` was determined from the default value
+/// - `From(val, came_from_default)`: the value is `val`, `came_from_default` it is was from the default value
```
siefkenj, line 16: Why do we need this when we already hve `Calculated` and `FromDefault`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop.rs
```diff
@@ -335,19 +335,30 @@ impl<T: Default + Clone> Prop<T> {
 
     /// Convenience function to call `calculate` on interface
```
siefkenj, line 336: -`on interface`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+/// Return a normalize mathematical expression from `math_object` as specified in `params`.
+///
+/// Example:
+///
+/// ```
+/// let expr = parse_text_into_math("(x+x+1)(2y+1-y)", true, &["f"]).unwrap();
+///
+/// let lp = ToLatexParams::default();
```
siefkenj, line 247: Why do all your examples store `lp` to a variable instead of just putting `ToLatexParams::default()` into where `lp` is used? In any case, this should be defined much closer to its use.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,327 @@
...
+/// );
+/// ```
+#[cfg(all(not(feature = "testing"), feature = "web"))]
+pub fn normalize_math<MathObj: AsRef<str>>(
+    math_object: MathObj,
+    params: NormalizeParams,
+) -> Result<String, anyhow::Error> {
+    let result: JsString = normalizeMath(
+        JsString::from(math_object.as_ref()),
+        JsString::from(params.simplify.to_string().to_lowercase()), // TODO: remove .to_lowercase()
```
siefkenj, line 267: More explanation about the `.to_lowercase` removal? Why is it there now. Why do we not want it later?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,327 @@
...
+#[cfg(all(not(feature = "testing"), feature = "web"))]
+pub fn math_to_latex<MathObj: AsRef<str>>(
+    math_object: MathObj,
+    params: ToLatexParams,
+) -> Result<String, anyhow::Error> {
+    let result: JsString = toLatex(
+        JsString::from(math_object.as_ref()),
+        params.pad_to_decimals.map(Number::from),
+        params.pad_to_digits.map(Number::from),
+        Boolean::from(params.show_blanks),
```
siefkenj, line 184: We could make this function more ergonomic to use if we grouped all the 2,3,4th arguments together as a `JsValue` (an object) and then defined `.into()` for `params`.
It culd look like `toLatex(JsString::from(math_object.as_ref()), params.into())`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
```diff
@@ -0,0 +1,327 @@
...
+///   i.e., one of these symbols followed by arguments in parentheses
+///   will be interpreted as apply that function to the arguments (rather than multiplication)
+///
+/// Examples:
+/// - `parse_latex_into_math(r#"\frac{xy}{z}"#, true, &["g"])` is `x` times `y` divided by `z`.
+/// - `parse_latex_into_math(r#"\frac{xy}{z}"#, false, &["g"])` is the multi-character symbol `xy` divided by `z`.
+/// - `parse_latex_into_math("g(x)", true, &["f", "g"])` is the function `g` evaluated at `x`.
+/// - `parse_latex_into_math("g(x)", true, &["f"])` is `g` times `x`.
+#[cfg(all(not(feature = "testing"), feature = "web"))]
+pub fn parse_latex_into_math<Latex: AsRef<str>, FnSymbol: AsRef<str>>(
```
siefkenj, line 99: This function is `str -> str`, so it's a little weird to have `parse_` in its name since we don't get the parse tree as a result. `latex_to_math` perhaps?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
```diff
@@ -1 +1,385 @@
...
+    // parse this string into a `MathExpr`, which is the expression template
+    let expression_template = match parser {
+        MathParser::Latex => MathExpr::from_latex(template_string, split_symbols, function_symbols),
+        MathParser::Text => MathExpr::from_text(template_string, split_symbols, function_symbols),
+    };
+
+    (expression_template, math_codes)
+}
+
+/// Concatenate all string and math values, only replacing the math values with generated codes.
```
siefkenj, line 336: Example, please

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
```diff
@@ -1 +1,385 @@
...
+/// Parameters:
+/// - `maths_and_strings`: the math and string values forming the expression
+/// - `code_prefix`: the beginning of each generated code used to represent the math values
+/// - `parser`: an enum specifying whether we will use the latex or text parser to create the `MathExpr`
+///
+/// Returns a tuple of:
+/// - the template string that will be parsed into the expression template
+/// - the math codes used to represent each math value
+fn create_template_string(
+    maths_and_strings: &mut [MathOrString],
```
siefkenj, line 349: Mutable values for functions with a return value make me unhappy. Why do you need this `mut`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
```diff
@@ -1 +1,385 @@
...
+    /// when parsing the string to a math expression
+    split_symbols: DataQuery,
+
+    // TODO: this should be based on a data query for a prop/attribute once we implement array props or attributes
+    /// A vector of the symbols that should be treated as functions if they are followed by parentheses.  
+    function_symbols: Vec<String>,
+
+    /// A cached value of the expression template used to calculate the final mathematical expression,
+    /// saved here in order to prevent the need for its recalculation if only math values change
+    expression_template_cache: Option<MathExpr>,
```
siefkenj, line 31: Perhaps this is why you needed things to be `mut`? If we're going to have various bits that need caching on different components, perhaps we should set up a general caching layer in Core. A cache could be returned along with the `DataQuery` results. That way the functions could focus on being pure functions instead of worrying about caching stuff. (Caching will also make `cloning` a bigger operation)

packages/static-assets/scripts/generate-math-assets.ts
```diff
@@ -0,0 +1,249 @@
+import * as fs from "node:fs/promises";
```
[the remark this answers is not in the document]
dqnykamp: The reason I did this was because I wanted the comments, which I couldn't add to the JSON file.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Right, because if you pass a `&[]`, it gets mad at you for not being able to infer the type. So, I gave up and just passed in `&["f"]` instead. It seemed that adding type annotations to `&[]` made it look too complicated. Not sure the best approach here.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
dqnykamp: This isn't the bad reason for all the `mut`. (It's the tracking of what's changed that is the really annoying part that should be removed with refactoring.)

We do have a caching layer between this and Core that caches the `RequiredData`, which I toyed with extending for this case. But, it seemed like too heavy of a lift for this first pass, as it would require macros to deal with keeping everything properly typed.

I feel like I don't yet have a clear enough picture of how this should work to implement the caching. I'm going to have to revisit all this when implementing `invert` for math, the complexity of which will shed more light on needed functionality.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
dqnykamp: I think the standard simplify variant should be `numbersPreserveOrder`, not `numberspreserveorder`, but this code is currently calling the codebase from `@doenet/utils` that is also called from the Javascript implementation of core, so I can't change it now without a lot of refactoring of the Javascript code that I don't want to do.

siefkenj: That can go in the comment then. (I also think we can jettison the JS Core...)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
siefkenj: That's an excellent point. There should be an opaque type `struct JsMathExpr(String)` or `struct MathObject(String)` or something like that. Then this type can be passed around and we don't have to know what's inside unless we are directly manipulating it.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Maybe, but `splitSymbols` is the current DoenetML word used for the attribute. We could have the source code deviate from the DoenetML term if there is a strong enough reason, but I try to minimize that. I've generally gotten resistance at making names longer, so I'm not sure if `singleCharSymbols` would be approved for DoenetML.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
siefkenj: Maybe make the last argument an option, then? Or perhaps the builder patter would be better here. You could add a `finalize()` option that would initiate the parsing and return the actual object. It would look like

```
MathExpr::from_text("xy").split_symbols().finalize()`
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
siefkenj: I didn't know it was a `DoenetML` keyword.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: I don't think the ast of math-expressions uses `'`, but one could put a `'` inside a `MathOrPrimitive::VariableName`, which will break this. But, `VariableName` isn't supposed to handle that case. (I'm not sure what to call it. Maybe I'll call it `Symbol`.)

I'm not comfortable with any of these `eval_js` uses, but I got tired of making an extra Javascript function for each of these, so put these in for now. My plan is to eventually remove them and create separate Javascript functions like I did for `normalize`, etc.

I want to get a better handle on what api we actually need to create on `MathExpr` and then come up with the best approach to communicating between Rust and JS. Does what I did for `normalize_math`, etc., seem reasonable and should be reproduced for other functions like these? Or is there a better way?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Making the last argument an option would be the simplest approach.

I thought about the builder pattern, but then it wasn't clear how to minimize the number of calls Javascript. You could imagine wanting to do something like
```
MathExpr::from_text("xy(1+xy)").split_symbols().simplify().expand()
```
and have that communicate once to Javascript and then come back. But that seemed too complicated for now so I stuck with the separate functions, where each function is a call to Javascript. Maybe your `.finalize()` idea takes care of that, but it'd be nice if that weren't required.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: `MathOrPrimitive` doesn't work by itself. It is designed for arguments of `MathExpr` functions like `.add()` to give a shortcut so that you can, for example, add `1` without first creating a `MathExpr` for `1`.

OK, just writing out this reply helped me think through it. I'm thinking I'll call it `MathArg` and document it with the usage I just described.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop.rs
dqnykamp: old terminology. It's now called the `updater`.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
siefkenj:
```typescript
const SIMPLIFY = {
    none: "none",
    numbersPreserveOrder: "numberspreserveorder",
    numbers: "numbers",
    full: "full"
} as const;
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
dqnykamp: OK, I at least got rid of this `mut` by moving the recording of data viewing outside of the `calculate` function.

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
dqnykamp: Sure, but `simplify` is a `string` that can take on only one of those four values. (It came from a Rust enum.) Just making sure that assigning the const `SIMPLIFY` as the type for `simplify` has that meaning. It has a different feel to me, but I haven't taken the time to learn TypeScript in any depth.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/views.rs
dqnykamp: removed

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: redid this

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
siefkenj: Oh, I see what you're saying. Just make the type in the simplify function `option: "none" | "numberspreserveorder" | "numbers" | "full"` That will prevent a typescript user from mistyping it :-P

packages/doenetml-worker-rust/lib-js-wasm-binding/src/eval-math.ts
dqnykamp: done

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
dqnykamp: Ugh, making `function_symbols` an `Option<&[FnSymbol]>` didn't help, as passing it `None` for `function_symbols` still makes it mad because it cannot infer the type of the parameter `FnSymbol`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
dqnykamp: It was there to make the examples easier to read! But, since apparently this failed, I got rid of the `lp`s.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/math_via_wasm.rs
dqnykamp: We could do that, but I'm reluctant to add it to these intermediate "throw away" functions that won't be used directly, given that it will add more complexity with more implementations depending on this conditional compilation nonsense.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/updater.rs
dqnykamp: I wasn't able to get rid of this `mut`, but I did get rid of others. We can revisit mutability patterns after we finish retooling this part of core.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
```diff
@@ -80,20 +79,29 @@ impl PropUpdater<String, RequiredData> for StringProp {
         .into()
     }
 
-    fn calculate<'a>(&self, data: &'a RequiredData) -> PropCalcResult<'a, String> {
+    fn calculate(&mut self, data: &RequiredData) -> PropCalcResult<String> {
```
siefkenj, line 82: You weren't able to get rid of this `mut` after all?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -0,0 +1,462 @@
...
+/// and may be replaced by a Rust-based symbolic tool in the future.
+#[derive(Debug, Clone, PartialEq, serde::Deserialize)]
+pub struct MathExpr {
+    pub math_object: JsMathExpr,
+}
+
+impl MathExpr {
+    /// Produce a string representation of the `MathExpr`
+    /// that will be revived into the `math-expression` object when eval'ed in Javascript.
+    pub fn to_reviver_string(&self) -> String {
```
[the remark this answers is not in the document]
dqnykamp: It turns into a math expression here, but on line 339, it turns in different objects (math expression, number, string). I wanted to call the function the same thing in all cases. Plus, "math expression string" makes me think of the `JsMathExpr` object, which is the the string storing the `MathExpr` in the first place. I was trying to find a term that distinguishes those case. The best I've come up with so far is: `.to_js_string()` and `to_reviver_string()`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/views.rs
```diff
@@ -326,4 +334,8 @@ impl<T: Default + Clone> QueryUpdateRequests for PropView<T> {
             vec![]
         }
     }
+
+    fn record_data_viewed(&mut self) {
+        self.record_viewed();
```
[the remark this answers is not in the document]
dqnykamp: No, it's just recording the fact that it was viewed. `mark_viewed` is fine.

PR #128, opened 2024-02-16
Add General Dependency Graph

siefkenj:
This PR adds a `Graph` object for keeping track of dependencies. This PR also does some minor renaming.

Currently `Graph` is not used anywhere.


packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+            None
+        }
+    }
+}
+
+/// `Taggable` items can have data attached to them. This provides hash-map-like functionality,
+/// but may be optimized for faster lookup.
+pub trait Taggable<Node, T> {
+    /// Get the index of `node`
+    fn get_tag(&self, node: Node) -> &T;
```
dqnykamp, line 222: Is the reason for not addressing the case where `Node` isn't in the set a matter of efficiency and that your code never tries to get something that doesn't exist?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+/// graph.add_node("b".into());
+/// graph.add_node("c".into());
+/// graph.add_edge("a".into(), "b".into());
+/// graph.add_edge("a".into(), "c".into());
+/// graph.add_edge("b".into(), "c".into());
+/// assert_eq!(graph.walk_descendants("a".into()).collect::<Vec<_>>(), vec!["b", "c"]);
+/// ```
+#[derive(Debug)]
+pub struct Graph<Node: Clone + Eq + Hash + Debug, Index: Taggable<Node, usize>> {
+    /// An `Taggable` that allows for looking up the index of a node in the graph.
```
dqnykamp, line 24: An -> A

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// I.e., used for `Node -> usize` lookups.
+    index: Index,
+    /// The nodes in the graph. Internally nodes are represented as `usize`. But `Graph` stores nodes
```
dqnykamp, line 27: The use of `index` to mean two different things in the span of a few lines is confusing. On line `24` it means the `usize`, but on lines 23 and 26, in means the structure storing these indices. When thinking about code, I think the use of index on lines 23 and 26 is less obvious. Maybe something with `map` in it, or you trying to emphasize that it doesn't need to be a `HashMap`?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// for the users to make its internal structure opaque.
+    nodes: Vec<Node>,
+    /// Directed edges between nodes.
+    edges: Vec<Vec<usize>>,
```
dqnykamp, line 31: Additional explanation of the data structure would be helpful here. My first thought when seeing the nested vectors was that `edges[i][j]` would have something to do with an edge from `i` to `j` rather than the `j`th edge that has been created coming from node `i`.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+                self.remaining_indices
+                    .extend(self.edges[index].iter().filter(|&&to| !self.visited[to]));
+            }
+            None
+        }
+    }
+}
+
+/// `Taggable` items can have data attached to them. This provides hash-map-like functionality,
+/// but may be optimized for faster lookup.
```
dqnykamp, line 219: In our potential, optimized use case, `Node` would be `usize`, i.e., the index corresponding to a bigger object stored outside of `Graph`, and then the `Taggable` would be a vector with `Node` being the index in the vector? In this case, we'd just pad the vector if we add an index out of order (which will definitely happen when adding edges or reverse edges)?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+    edges: Vec<Vec<usize>>,
+    /// Reverse directed version of every edge.
+    reverse_edges: Vec<Vec<usize>>,
+}
+
+impl<Node: Clone + Eq + Hash + Debug, Index: Taggable<Node, usize>> Graph<Node, Index> {
+    /// Add a node to the graph.
+    pub fn add_node(&mut self, node: Node) {
+        let index = self.nodes.len();
+        self.index.set_tag(node.clone(), index);
```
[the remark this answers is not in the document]
dqnykamp: Alternatively, it could panic if a node is duplicated. That might make it easier to debug uses of `Graph`.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+        // There should always be the same number of elements in the `nodes`, `edges`, and `reverse_edges`
+        // vectors.
+        self.nodes.push(node);
+        self.edges.push(Vec::new());
+        self.reverse_edges.push(Vec::new());
+    }
+    /// Set an edge between two nodes.
+    pub fn add_edge(&mut self, from: Node, to: Node) {
+        let &from_index = self.index.get_tag(from);
```
dqnykamp, line 49: The data structure could handle duplicate edges between nodes. I'm not sure a reason for supporting such a "multigraph", as it would seem easier to reason about the behavior of `Graph` if we allowed only one edge between each pair of nodes. In that case, maybe check for a duplicate and panic if a duplicate edge is created?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
+use std::{collections::HashMap, fmt::Debug, hash::Hash};
+
+/// An abstract graph. The speed of lookups is determined by the `Taggable` type,
```
dqnykamp, line 3: Somewhere in the comments or even the name of the data structure, it should make it clear we are taking about a directed graph.

dqnykamp: Maybe called it a "directed acyclic graph" or "DAG"

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: Especially if we call it a "directed acyclic graph", then we should make sure we don't have multiple edges.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+    /// Walk through all nodes that have `node` as a descendant. Nodes are walked in _topological_ order.
+    /// Panics if a cycle is detected.
+    pub fn walk_ancestors(&self, node: Node) -> DescendantTopologicalIterator<Node> {
+        let &start_index = self.index.get_tag(node);
+        DescendantTopologicalIterator::new(&self.nodes, &self.reverse_edges, start_index)
+    }
+
+    /// Iterate through all nodes that have `node` as an ancestor. This iterator is meant to be fast.
+    /// The order of the nodes is not guaranteed, and the same node may be yielded multiple times.
```
dqnykamp, line 68: Maybe include the comment that this possibility of a descendant will be yielded twice if there is a undirected cycle.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+/// graph.add_node("a".into());
+/// graph.add_node("b".into());
+/// graph.add_node("c".into());
+/// graph.add_edge("a".into(), "b".into());
+/// graph.add_edge("a".into(), "c".into());
+/// graph.add_edge("b".into(), "c".into());
+/// assert_eq!(graph.walk_descendants("a".into()).collect::<Vec<_>>(), vec!["b", "c"]);
+/// ```
+#[derive(Debug)]
+pub struct Graph<Node: Clone + Eq + Hash + Debug, Index: Taggable<Node, usize>> {
```
dqnykamp, line 23: Just for completeness, I'd either add a constructor where one can supply the initialized `Index` or just add the `Default` trait bound to `Index` here.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
```diff
@@ -0,0 +1,307 @@
...
+use iterators::*;
+pub mod iterators {
+    //! Iterators for the `Graph` type.
+
+    /// Iterate through the descendants of a node. This iterator is
+    /// meant to be fast, does not guarantee any particular order, and
+    /// may yield the same node multiple times.
+    pub struct DescendantIterator<'a, Node> {
+        nodes: &'a [Node],
+        graph: &'a [Vec<usize>],
```
dqnykamp, line 103: Probably call this `edges` for consistency with the rest of the code.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp, line 23: Why does a `Node` have to implement `Hash`?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/prop_graph.rs
```diff
@@ -0,0 +1,171 @@
...
+    //                self.graph.add_node(PropPointer::new(origin, prop_ptr.prop_idx));
+    //            },
+    //            _ => panic!("Not implemented"),
+    //        }
+    //    }
+}
+
+/// Pointer to a prop on a specific component.
+#[derive(PartialEq, Eq, Clone, Copy, Hash)]
+pub struct PropPointer {
```
dqnykamp, line 75: Interesting abstraction of how it looks like you are hashing `PropPointer` but you're instead doing something more efficient. I just added a comment `graph.rs` about this, but it's not clear to me why `PropPointer` has to implement `Hash`.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: OK, I see how are you doing it in `prop_graph.rs`, where `Node` is a multi-index.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
siefkenj: Could be cyclic, so I went with `DirectedGraph`

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
siefkenj: Could you explain more?

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
siefkenj: Yeah. This is meant to be implemented as an efficient sparse data structure for the `(comp_id, prop_id)` pairs. It's generalized so that it can store any information. I assume it is going to be used to store chache data/etc. as well, but the `Graph` object only needs it for `usize`s.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
siefkenj: I left out a check for efficiency. I could put one in, but the naive implementation could lead to quadratic insertion behavior, and putting a hashset in is probably overkill...

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: Is it possible to create a `Graph` where `Taggable` doesn't implement `Default`? I don't think so, because the fields are private and your constructors require that it implements `Default`. So, it just seemed strange to define a type (`Graph` where `Taggable` doesn't implement `Default`) that couldn't be constructed.

There's no practical purpose to my comment, and I don't really care whether you add the `Default` trait bound. It's more just a fun exercise in trying to reason with Rust trait bounds.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: Well, it can't be cyclic, or your walk functions panic!

siefkenj: But, on purpose, no checks are done when edges are inserted, so you could make a cycle, just not query a lot from it.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: As long it doesn't have strange behavior with multiple edges, it's probably not worth doing anything about it.

packages/doenetml-worker-rust/lib-doenetml-core/src/dep_graph/graph.rs
dqnykamp: Fair enough!

packages/doenetml-worker-rust/lib-doenetml-core/src/graph/directed_graph.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Panics if a cycle is detected.
+    pub fn walk_ancestors(&self, node: &Node) -> DescendantTopologicalIterator<Node> {
+        let &start_index = self.index_lookup.get_tag(node).unwrap();
+        DescendantTopologicalIterator::new(&self.nodes, &self.reverse_edges, start_index)
+    }
+
+    /// Iterate through all nodes that have `node` as an ancestor. This iterator is meant to be fast.
+    /// The order of the nodes is not guaranteed, and the same node may be yielded multiple times (e.g., if
+    /// the graph is not a tree).
+    pub fn descendants_quick(&self, node: &Node) -> DescendantIterator<Node> {
```
dqnykamp, line 77: Add the comment `Panics if a cycle is detected` here, too.

dqnykamp (approved):

PR #129, opened 2024-02-20
Refine text input

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/text_input/with_references.rs
```diff
@@ -40,6 +40,47 @@ fn text_input_reference_child_is_changed_when_update_value() {
     assert_eq!(get_text_value(text_idx, &mut core), "bye");
 }
 
+/// A text component referenced as a child of a text input
+/// will be changed with `UpdateValue`.
+/// This test is checking to make sure the text is changed by the actions
+/// even if it starts blank, i.e., with its default value.
```
siefkenj, line 46: I would love for this comment to include
```
For example in `<textInput>$t</textInput><text name="t" />`, 
the text input's value can still change even though it has 
a child consisting of an empty text.
```

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/text_input/basic_actions.rs
```diff
@@ -122,6 +124,46 @@ fn value_and_immediate_value_with_text_child_respond_to_actions() {
...
+    let press_enter_action = update_value_action(text_input_idx);
+
+    let _ = core.dispatch_action(press_enter_action);
+
+    assert_eq!(get_immediate_value(text_input_idx, &mut core), "bye");
+    assert_eq!(get_value(text_input_idx, &mut core), "bye");
+    assert_eq!(get_text_value(text_idx, &mut core), "bye");
+}
+
+/// The UpdateImmediateValue and UpdateValue actions work the same when have a blank text child,
```
siefkenj, line 140: `when have`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/prop/updater.rs
```diff
@@ -199,22 +186,18 @@ where
...
             cache: Default::default(),
-            queries_used: Default::default(),
         };
         Prop::new(Box::new(updater_with_cache), default_value)
     }
 }
 
 pub trait DependenciesToData<RequiredData> {
-    fn to_data(&self, queries_used: &[usize]) -> RequiredData;
+    fn to_data(&self) -> RequiredData;
```
siefkenj, line 195: Good to see params eliminated :-)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/boolean_prop.rs
```diff
@@ -23,38 +23,69 @@ pub struct BooleanProp {
...
+    /// - If `propagate_came_from_default` is `false`, then `came_from_default`
+    ///   will always be `false` if there are any matching children.
+    ///
+    /// Regardless of the `propagate_came_from_default` argument,
+    /// - If there are two or more matching children, `came_from_default` is always `false`.
+    /// - If there are no matching children, then `came_from_default` will initially be true
+    ///   (and the value of the prop itself will be set to `default_value`).
+    ///   As soon as the value is changed (by a call to `invert()`),
+    ///   then `came_from_default` will be set to `false`.
+    pub fn new_from_children(default_value: bool, propagate_came_from_default: bool) -> Self {
```
siefkenj, line 81: Not thrilled about this extra param. Seems like you almost always set it as `true`?

I suggest leaving it off of the `new_from_children` and creating an additional `new_from_children_no_propagate_came_from_default`. Let the ugly name be used in the few cases its needed.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
```diff
@@ -115,6 +185,20 @@ impl PropUpdater<String, RequiredData> for StringProp {
         _is_direct_change_from_action: bool,
     ) -> Result<Vec<DependencyValueUpdateRequest>, InvertError> {
         match data.strings.len() {
+            0 => {
+                if self.propagate_came_from_default {
+                    // if propagate_came_from_default is true,
+                    // then always_return_value is true on the string data query,
+                    // so we should never reach this
```
siefkenj, line 192: I don't understand why, but I believe you...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
```diff
@@ -17,52 +17,103 @@ pub struct StringProp {
...
+    /// Regardless of the `propagate_came_from_default` argument,
+    /// - If there are two or more matching attribute components, `came_from_default` is always `false`.
+    /// - If there are no matching attribute components, then `came_from_default` will initially be true
+    ///   (and the value of the prop itself will be set to `default_value`).
+    ///   As soon as the value is changed (by a call to `invert()`),
+    ///   then `came_from_default` will be set to `false`.
+    pub fn new_from_attribute<S: Into<String>>(
+        attr_name: AttributeName,
+        default_value: S,
+        propagate_came_from_default: bool,
```
siefkenj, line 107: Why don't we have a consuming `.dont_propagate_came_from_default()` method here that returns a version that has `propagate_came_from_default` as `false`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
```diff
@@ -72,20 +123,39 @@ impl PropUpdater<String, RequiredData> for StringProp {
         self.default_value.clone()
     }
 
-    fn return_data_queries(&self) -> Vec<Option<DataQuery>> {
+    fn return_data_queries(&self) -> Vec<DataQuery> {
         RequiredDataQueries {
-            strings: Some(self.data_query.clone()),
+            independent_state: DataQuery::State,
```
siefkenj, line 128: I hope we can figure out another way to do this...

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
```diff
@@ -72,20 +123,39 @@ impl PropUpdater<String, RequiredData> for StringProp {
...
+                    // if propagate_came_from_default is true,
+                    // then always_return_value is true on the string data query,
+                    // so we should never reach this
+                    unreachable!()
+                }
+
+                // If propagate_came_from_default is false and there were no dependencies returned,
+                // then use the independent state, propagating its came_from_default as well as its value.
+                // In this way, came_from_default will be true only if no dependencies were returned.
+                data.independent_state.prop_calc_result()
```
siefkenj, line 147: What's the difference between `prop_calc_result()` and `calc()`? Can the functions have the same name?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
```diff
@@ -39,6 +54,11 @@ pub struct MathProp {
 #[add_dependency_data]
 #[derive(Debug, Default, PropDependencies, PropDataQueries)]
 pub struct RequiredData {
+    /// An independent state variable (that doesn't have any dependencies)
```
siefkenj, line 57: `state variable` -> `prop` or `piece of data`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/independent_prop.test.rs
```diff
@@ -125,16 +125,16 @@ mod setup_functions {
 
         // Note: the default_value specified in the creation of prop above
         // isn't used at this level of testing, so we supply it directly here to match
-        let (preliminary_value_dependency, preliminary_value_var) =
+        let (independent_state_dependency, independent_state_var) =
```
siefkenj, line 128: `state_var` -> `prop`

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/text_input/with_references.rs
dqnykamp: The key point of this test isn't exactly that. It is the motivation for the additional parameter in the props, below. Without that parameter, the text input would still change, but it wouldn't recognize that it had a text child, and so the text child would still stay blank as the text input changes.

I added more text to elaborate further.

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/text_input/basic_actions.rs
dqnykamp: fixed

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: I'm not sure, as we don't have a `calc`. `prop_calc_result()` propagates both the value and the `came_from_default` of the prop.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/boolean_prop.rs
dqnykamp: Right, I wasn't sure whether to duplicate code and make a separate function or add another parameter. I guess could make the current function be a non-public one that is called from `new_from_children` and `new_from_children_no_propagate_came_from_default`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
dqnykamp: This is actually a `StateProp`, the thing that used to be called "essential data". Since both become a `PropView<T>`, the distinction get muddied. It's hard to figure out what is the best way to communicate what's happening. I can try on "piece of data".

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: We need a place to store changes when there aren't dependencies. One way is with the "state props" that get snuck into children/attributes when there are no components (if `always_return_value` is `true`). If we don't have that turned on (because we want to treat the case with no components differently), then we have to create another "state prop" to store changes.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: Not sure the best name for it. It returns a `PropCalcResult` based on the prop. If it creates confusion, I could just remove it and write the `if` statement manually each time.

At one point it was instead
`PropCalcResult::From(data.indepenent_state)`
which sent a reference to `data.independent_state`. But sending a reference is no longer an option due to Rust borrowing rules and the fact that I now mutate `data.independent_state` immediately after this function returns to mark it viewed and before processing the `PropCalcResult`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: I like this way of getting around Rust's refusal to provide default params!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: To make the propagation of `came_from_default` more explicit, I'm removing this function and writing the `if` statement manually each time.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/independent_prop.test.rs
dqnykamp: Trying to distinguish between the "independent state prop" that is the leaf of the dependency graph and the "independent prop" that is the actual prop that mirrors the state.  I'm trying "independent_state" for the former.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/string_prop.rs
dqnykamp: I tried making the wording more explicit.

PR #130, opened 2024-02-23
Simplify the algorithm for extending components

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/common.rs
```diff
@@ -179,14 +182,8 @@ pub trait ComponentNode: ComponentState {
...
 
-    /// When this component has `extend="$ref"`, depending on the different
-    /// `ComponentProfiles` `$ref` may present itself as, the component might want
-    /// to set different prop values. This function returns a vector of
-    /// possible pairings of the `ComponentProfile` that `$ref` may provide and
-    /// the index of the prop that should be set if `$ref` provides that
-    /// `ComponentProfile`.
-    fn accepted_profiles(&self) -> Vec<(ComponentProfile, PropIdx)> {
-        vec![]
+    fn extend_via_default_prop(&self) -> bool {
```
siefkenj, line 185: Docstring?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -48,17 +48,32 @@ impl ComponentBuilder {
...
+            let ref_source = elm.extending.clone().unwrap();
             let component = &builder.components[idx];
-            let referent = &builder.components[ref_resolution.node_idx];
+            let referent = &builder.components[ref_source.idx()];
 
-            match Self::determine_extending(&ref_resolution, component, referent) {
+            match Self::determine_extending(ref_source, component, referent) {
                 Ok(extending) => {
-                    builder.components[idx].set_extending(extending);
+                    let extending_component = matches!(extending, Extending::Component(..));
```
siefkenj, line 60: -> `is_extending_component`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -48,17 +48,32 @@ impl ComponentBuilder {
...
+                    let component_type_changed =
+                        component.get_component_type() != referent.get_component_type();
```
siefkenj, line 62: If you don't need these for other things, define them just before the if statement where they're used. This lessens the cognitive burden of the reader.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -48,17 +48,32 @@ impl ComponentBuilder {
...
+                        // Since we are extending a component where the component type changed,
+                        // the reference was inside an extend attribute and
+                        // `Expander::consume_extend_attributes` added an extra child
+                        // in case the referent was marked `extend_via_default_prop`
+                        // which would created an `Extending::Prop`.
+                        // The fact that we still have an `Extending::Component`
+                        // means that this extra child was not needed, so we delete it.
+                        let mut children = builder.components[idx].take_children();
+                        children.remove(0);
+                        builder.components[idx].set_children(children);
```
siefkenj, line 75: I'm not thrilled with adding children to then be conditionally removed later. We'll discuss if there is another way.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
dqnykamp: I agree. You just need to convince the Rust borrow-checker that I need to be able to move them down.

The `set_extending` is a mutable borrow, and if I move these references definitions below that line, then immutable borrow is sustained past the mutable borrow, violating the rules! I could redefine `component` and `referent` so they don't overlap if we really want them adjacent.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
dqnykamp: We could look up to see if the referent has `extend_via_default_prop` back when we attach the file in `Expander::consume_extend_attributes`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -242,6 +229,58 @@ impl ComponentBuilder {
         }
     }
 
+    /// If component `component_idx` extended a prop using the `extend` attribute,
+    /// then add a child to the component corresponding to that prop.
+    ///
+    /// For example, if `<text name="$i"/>`, then we add a child corresponding to `$i.value`
+    /// in the case of `<p extend="$i.value" />` but not to the text component coming from a `$i.value`
+    /// that occurs outside the `extend` attribute.
```
siefkenj, line 237: `name="$i"` -> `name="i"`. I do not understand this example. Did you mean `<textInput name="i" />` and `<text extend="$i" />`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/dast/ref_expand.rs
```diff
@@ -233,32 +228,7 @@ impl Expander {
...
-            // corresponding to that prop.
-            // If the component type did not set `extend_via_default_prop`,
-            // then we will delete this child as we are extending an entire component.
-
-            if source.get_resolution().unresolved_path.is_some() || ref_name != element.name {
-                element.children.insert(0, UntaggedContent::Ref(ref_idx));
-            }
-
-            element.extending = Some(source.as_attribute());
+            element.extending = extend_referent.map(|source| source.as_attribute());
```
siefkenj, line 231: Comment explaining why this cast is okay?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -242,6 +229,58 @@ impl ComponentBuilder {
...
+    fn add_child_if_prop_from_extend_attribute(&mut self, component_idx: ComponentIdx) {
```
siefkenj, line 238: I don't like the `mut` here. I would much rather this function compute something and return an `Option<child>` and let the caller do the insertion.

It could be called `get_implicit_child` or something like that.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
```diff
@@ -242,6 +229,58 @@ impl ComponentBuilder {
...
+            None
+        };
+
+        if let Some(new_child) = new_child {
+            // Add the child created to the component as well as the vector of all components
+            let component = &mut self.components[component_idx];
+            let mut children = component.take_children();
+            children.insert(0, UntaggedContent::Ref(new_child.get_idx()));
+            component.set_children(children);
+            self.components.push(new_child);
```
siefkenj, line 280: This process is complicated. Perhaps a `prepend_child()` utility function could be created?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
dqnykamp: Since neither `<text>` nor `<textInput>` have rendered children, I was trying to use an example `<p>` that has rendered children (given that the need for rendered children was the reason we did it this way). I'll make the example have both `<text>` and `<p>` extending from a `<textInput>`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
dqnykamp: Hah, but that would be a `mut` function! It'll try breaking it into two functions, one `mut` and one not.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/components/component_builder.rs
dqnykamp: I couldn't get myself to make the name that short, but I tried.

PR #131, opened 2024-02-23
[WIP] Internals Rework

siefkenj:
Rework Core to use `DirectedGraph`


packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/component_builder.rs
```diff
@@ -0,0 +1,691 @@
...
+            let referent_prop_node = self.structure_graph.get_component_props(referent_node)
+                [prop_source.prop_pointer.prop_idx];
+
+            self.structure_graph
+                .prepend_edge(&component_children_virtual_node, &referent_prop_node);
+        }
+    }
+}
+
+#[cfg(test)]
```
dqnykamp, line 689: Delete, since duplicate.

packages/doenetml-worker-rust/lib-doenetml-core/src/graph/directed_graph.rs
```diff
@@ -168,28 +175,96 @@ use iterators::*;
...
+            let (from, to) = self.descendant_edge_iterator_raw.next()?;
+            Some((&self.nodes[from], &self.nodes[to]))
+        }
+    }
+
+    /// Iterate through the descendants of a node but return
+    /// the `DirectedGraph`'s internal representation of each node.
+    ///
+    /// **For internal use**. Only use this function if you know what you're doing.
+    pub struct DescendantIteratorRaw<'a> {
```
dqnykamp, line 248: Mark this as private?

siefkenj: It has to be public since it is used by a parent..., but I can certainly make it `pub(crate)`

dqnykamp: The current code still complies for me if I get rid of `pub` on the two `Raw` structs.

PR #133, opened 2024-02-29
Replace `core` with `new_core`

siefkenj:
This PR breaks a lot, but things should be fixed eventually...


packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/render/render_children.rs
```diff
@@ -0,0 +1,73 @@
...
+                "Can only get the string of a GraphNode::String, not {:?}",
+                node
+            ),
+        }
+    }
+}
+
+/// All components that can be rendered must implement this trait.
+/// It instructs Core how to render the children of a component.
+pub trait ComponentChildren<'a> {
```
dqnykamp, line 64: Could we provide implementations of the two simple cases via a derive macro, based on the `#[pass_though_children]` and `#[no_rendered_children]` attributes? Then, component authors would need to create their implementation of the trait only if they want more complicated behavior.

We could also further lower the bar for component authors by changing the return type to a vector, which I think will be easier to understand and implement than an iterator object. My guess is that we don't need the efficiency of an iterator for the return type, given that we will be doing something with the children whenever the function is called (though we do want the `child_query_object` argument to be an iterator so it could be lazy).

siefkenj: Yeah. That can be changed. I didn't want to do a derive macro until we knew the final form, since rust analyzer is much more helpful at finding issues when no macros are involved.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/render/render_children.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// Get an iterator that will iterate through your children in order.
+    pub fn child_iter(&self) -> impl Iterator<Item = GraphNode> + 'a {
+        self.structure_graph.content_children(self.self_graph_node)
+    }
+
+    /// Get the type of the component specified by `node`. `node` must
+    /// be a `GraphNode::Component`.
+    pub fn get_component_type(&self, node: GraphNode) -> &str {
```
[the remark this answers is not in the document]
siefkenj: That was my though as well. And with this abstraction, it should be easy to add new methods without breaking existing code.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+        *self.value.borrow_mut() = Some(value);
+        self.meta.borrow_mut().status = PropStatus::Fresh;
+        let change_counter = self.meta.borrow().change_counter;
+        // A little bit of safety in case someone wiggles their mouse 4 billion times
+        self.meta.borrow_mut().change_counter = change_counter.wrapping_add(1);
+    }
+
+    /// Sets the prop as stale and clears the cached value using interior mutability.
+    /// **For internal use only.**
+    fn _set_stale(&self) {
```
dqnykamp, line 65: When mark stale, don't remove the value. That way, if the result of a `calculate()` is `NoChange`, you can just mark the cached value as `Fresh` and the old value is restored. We can prevent anyone from accessing the obsolete value as it is private.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+    ) -> Ref<PropValue> {
+        match self.get_status() {
+            PropStatus::Fresh => self.get_cached_value().expect(
+                "Prop is marked as Fresh but no value is cached. This state should be unreachable",
+            ),
+            PropStatus::Stale | PropStatus::Resolved => {
+                let result = calculate();
+                match result {
+                    PropCalcResult::Calculated(value) => {
+                        self._set_value(value);
```
dqnykamp, line 98: You need to set `came_from_default` to be `false` here.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Set `came_from_default` using interior mutability.
+    /// **For internal use only.**
+    fn _set_came_from_default(&self, came_from_default: bool) {
+        self.meta.borrow_mut().came_from_default = came_from_default;
+    }
+
+    /// Set `status` using interior mutability.
+    /// **For internal use only.**
+    fn set_status(&self, status: PropStatus) {
```
dqnykamp, line 78: If this is also internal, is there a reason is doesn't start with a `_`?

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+                match result {
+                    PropCalcResult::Calculated(value) => {
+                        self._set_value(value);
+                    }
+                    PropCalcResult::FromDefault(value) => {
+                        self._set_value(value);
+                        self._set_came_from_default(true);
+                    }
+                    PropCalcResult::NoChange => {
+                        self.set_status(PropStatus::Fresh);
```
dqnykamp, line 105: For this `set_status` to work, you'll need to change `_set_stale` to just change the status and not clear the value.

Also, if the `status` is `Resolved`, it should panic, as it means the prop author made a mistake.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+            }
+            PropStatus::Unresolved => {
+                panic!("Cannot get value of prop with Unresolved status")
+            }
+        }
+    }
+
+    /// Get the cached value of the prop without any side computations.
+    fn get_cached_value(&self) -> Option<Ref<PropValue>> {
+        // This code doesn't seem to work with `self.value.borrow().map(...)`. The borrow checker
```
dqnykamp: Yep, I remember having to use `Ref::map` to implement this. I don't remember it being an issue with the borrow checker but just the fact that `.borrow()` gives the type `Ref<T>`. which doesn't implement `.map()`.

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,254 @@
...
+        PropCache {
+            store: GraphNodeLookup::new(),
+            change_tracker: GraphNodeLookup::new(),
+        }
+    }
+    
+    /// Get the status of a prop.
+    pub fn get_prop_status<A: borrow::Borrow<GraphNode>>(&self, prop_node: A) -> PropStatus {
+        let prop_node = prop_node.borrow();
+        if self.store.get_tag(prop_node).is_none() {
```
dqnykamp: `.map_or_else()`?

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// **For internal use only.**
+    fn _set_value(&self, value: PropValue) {
+        *self.value.borrow_mut() = Some(Rc::new(value));
+        self.meta.borrow_mut().status = PropStatus::Fresh;
+        let change_counter = self.meta.borrow().change_counter;
+        // A little bit of safety in case someone wiggles their mouse 4 billion times
+        self.meta.borrow_mut().change_counter = change_counter.wrapping_add(1);
+    }
+
+    /// Sets the prop as stale and clears the cached value using interior mutability.
```
dqnykamp, line 59: It no longer clears the cached value

packages/doenetml-worker-rust/lib-doenetml-core/src/new_core/props/cache.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// since the other counter was last set.
+    pub change_counter: u32,
+}
+
+/// Store a prop's and other meta information needed to operate a cache
+/// that intelligently recomputes props when their dependencies change.
+#[derive(Debug)]
+struct CachedProp {
+    /// Cached value of the prop. Will be `None` if the prop hasn't been computed or is stale.
+    value: RefCell<Option<Rc<PropValue>>>,
```
dqnykamp, line 33: I'm still working out how this would be used. But, it seems to me that it'd be more useful with having an `Rc<RefCell<>>`. That way, one can store a copy of the cache somewhere and then have a handle to get its updated value. But maybe that is not the paradigm you are going for.

Having a `Rc` of an immutable value seems useful only to avoid cloning the `PropValue` or dealing with lifetimes/open Refs, which may be the reason you put the `Rc` here.

siefkenj: Yeah. I did it for lifetime issues. Only one thing should ever be changing the cache. If someone is holding onto an old reference when the cached value is changed, that's fine. It's their duty to re-ask the cache for the value.

dqnykamp (approved):
OK, we have the base features working again. Let's merge it in.

PR #135, opened 2024-03-20
[No Merge] Prototype of nextra docs

siefkenj:
This is a result of my conversation with Jason about how to get a DoenetML component rendering in Nextra.

This PR should not be merged.


siefkenj:
Nextra docs are up and running, so this is no longer needed.

PR #140, opened 2024-04-03
Numbering in Section Component

siefkenj:
This PR adds the ability for a `<section>` to display its number.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+        /// Children of this component will be searched
+        container: PropSource,
+        /// How to filter the children. This should be a [`ContentFilter`] or a
+        /// composition of [`ContentFilter`]s.
+        ///
+        /// See [`DataQuery::ComponentRefs`] for an example.
+        filter: Rc<dyn for<'a> BindableAsGraphNodeFilter<'a>>,
+    },
+
+    AncestorRefs {
```
dqnykamp, line 139: docstring explaining what this query does. Does it return `ComonentRefs`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
-    /// Query for all children that have a prop that matches the prescribed `PropProfile`s.
-    /// Returns the matching props
-    ChildPropProfile {
-        /// The data query will match child components that have at least one of these profiles
-        match_profiles: Vec<PropProfile>,
-    },
-
-    /// Query for all child GraphNodes,
-    /// filtering element nodes for components which contain PropProfile props given specified values
+    /// Query for all child GraphNodes, filtering element nodes based on the supplied
```
dqnykamp, line 106: Does it return `GraphNodes` or `ComponentRefs`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+    ///    filter: Rc::new(Op::And(
+    ///      ContentFilter::IsType("section"),
+    ///      ContentFilter::HasPropMatchingProfile(PropProfile::Hidden),
+    ///    ))
+    /// }
+    /// # ;
+    /// ```
+    /// This query will return all children of the querying component that are `<section>` components
+    /// and have a prop matching the `PropProfile::Hidden` profile.
+    ComponentRefs {
```
dqnykamp, line 129: If this just does children and multiple queries return `ComponentRefs`, maybe call this `ChildRefs` to be parallel to `AncestorRefs`?

dqnykamp, line 129: I'm wondering if we want parallels to the queries that return props and and queries that return `ComponentRefs`. Could the filtering/specifiying be made the same, with the only difference being that in one case, we return a `ComponentRef`, and in another case we return a `prop`?

It seems we have the cases:
1. Return a reference to possibly multiple component
2. Return a reference to a single component (could be rolled into 1)
3. Return a single prop from a single component
4. Return a single prop from multiple components
5. Return multiple props from a single component
6. Return multiple props from multiple components.

If we think of cases 1 and 2 as returns zero props, then could all follow the same structure.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+        /// Children of this component will be searched
+        container: PropSource,
```
dqnykamp, line 131: It's confusing why this is called a `PropSource` as there are no references to props here.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+        /// How to filter the children. This should be a [`ContentFilter`] or a
+        /// composition of [`ContentFilter`]s.
+        ///
+        /// See [`DataQuery::ComponentRefs`] for an example.
+        filter: Rc<dyn for<'a> BindableAsGraphNodeFilter<'a>>,
```
dqnykamp, line 136: I guess the fact that this `filter` is so different from the `prop_specifier` is an argument for having a different mechanism for the cases where we aren't grabbing a prop. But, from a high level, it seems like the cases with or without selecting a prop should be similar.

I still haven't done the work to figure out how a `BindableAsGraphNodeFilter` works. Have you decided whether or not we should remove one level of dynamic trait objects?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
     },
 
     /// Query for a particular prop of a component
     Prop {
-        /// If None, prop is from the component making the query.
-        component: PropComponent,
+        /// Where to look for the desired prop.
+        source: PropSource,
+        /// What prop to look for. Node `PropSpecifier::LocalIdx` is only valid
```
dqnykamp, line 154: "Node" -> "Note that"

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+        /// when used in conjunction with `PropComponent::Me`.
+        prop_specifier: PropSpecifier,
+    },
 
-        /// The prop from component_idx or component making the query.
+    /// Query multiple components and pick a prop from each.
+    PickProp {
+        /// Where to find the component's whose props will be searched.
+        source: PickPropSource,
+        /// How to filter the props to pick. Node that `PropSpecifier::LocalIdx` is forbidden
```
dqnykamp, line 163: "Node that" -> "Note that"

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -89,43 +101,67 @@ impl From<LocalPropIdx> for PropSpecifier {
...
+        /// Where to look for the desired prop.
+        source: PropSource,
+        /// What prop to look for. Node `PropSpecifier::LocalIdx` is only valid
+        /// when used in conjunction with `PropComponent::Me`.
+        prop_specifier: PropSpecifier,
+    },
 
-        /// The prop from component_idx or component making the query.
+    /// Query multiple components and pick a prop from each.
+    PickProp {
```
dqnykamp, line 160: The fact that `Pick` implies multiple components is a bit confusing terminology. (It seems like that is the difference.) Any choice here is influenced by the larger discussion, above, on comparison to the cases without props.

Looking at this more, I guess the difference isn't just single component versus multiple components. In the case of `PickPropSource::NearestMatchingAncestor`, you still just get a single component, but the component selected is determined by the `PropSpecifier` (as opposed to the `Prop` query, where the component is first selected). The way this works seems quite similar to the filtering of components.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -170,4 +209,6 @@ pub struct ComponentTypeDataQueryFilter {
 pub enum DataQueryFilterComparison {
     Equal,
     NotEqual,
+    // TODO: this is not the right way to do this. Cleanup how to build queries.
```
dqnykamp, line 212: With  `ContentFilter::HasPropMatchingProfileAndCondition`, is `ProfilePresent` needed anymore? If we merge the two types of filtering, maybe none of the `DataQueryFilterComparison` will be needed anymore.

siefkenj: The comparions should not be used anywhere anymore. I will delete them.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
siefkenj: This replaced the previous `ComponentProps`.

Ruby on Rails uses the function `pluck` for when you take one entry from each item in an array. SQL uses `select`. Preferences? `SelectProp` does sound good to anyone with a SQL background.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
siefkenj: I think it's fine to keep. It's all working, and does make the usage nice. The only non-nice thing is that `BindableAsGrahNodeFilter` doesn't tell you that you should be using `Op::...` to build your query.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
siefkenj: I think that would be confusing. So..you want your siblings, so you make a `ChildRefs` query?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/prop_value.rs
```diff
@@ -56,15 +58,17 @@ pub mod prop_type {
...
+    use crate::state::types::{content_refs, element_refs};
 
     pub type String = Rc<std::string::String>;
     pub type Number = f64;
     pub type Integer = i64;
     pub type Boolean = bool;
     pub type Math = Rc<MathExpr>;
-    pub type ElementRefs = Rc<element_refs::ElementRefs>;
-    pub type GraphNodes = Rc<Vec<GraphNode>>;
+    pub type ComponentRef = Option<element_refs::ElementRef>;
```
dqnykamp, line 68: Having both terms `ComponentRef` and `ElementRef` is confusing. They seem to mean the same thing: a reference to a component via its `ComponentIdx`. I don't have a strong preference, but we should pick one and standardize around it. I guess with `ComponentIdx`, we seem to use `component` more than `element`.

`ComponentRef/ElementRef` is in contrast to `ContentRef` which is either a component or a string. I'm assuming that is the only difference. In that case, I'm confused why `prop_type:ComponentRef` is an option while `prop_type:ContentRef` is not. I'm going to assume it is due to a use case that we've had with `ComponentRef/ElementRef` that we haven't had with `ContentRef`.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
```diff
@@ -35,8 +32,26 @@ mod component {
         )]
         Hidden,
 
+        /// The position of the section relative to other siblings with serial numbers.
+        /// E.g. in `<section /><section />` the first would have serial number 1 and the second 2.
```
dqnykamp, line 36: This description of serial number is incorrect, as is currently 0-indexed (and the tests check that). I think, though, that rather than changing the description, we should actually make `SerialNumber` be 1-indexed to correspond to `CodeNumber`.

Should we also make `SerialNumber` and `CodeNumber` be public? I could see them as being useful for an author to reference. If we do make them public, then `SerialNumber` definitely should be 1-indexed.

dqnykamp: If we make them public, then we should add a test where they are referenced.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
```diff
@@ -35,8 +32,26 @@ mod component {
...
+        /// The code-number uniquely identifying this `<section />`. E.g. the `1.2.3`
+        /// in _Section 1.2.3_.
+        #[prop(
+               value_type = PropValueType::String,
+               profile = PropProfile::CodeNumber,
+               for_render
+           )]
+        CodeNumber,
+
+        //SelfRef,
```
dqnykamp, line 52: eliminate comment

packages/doenetml-prototype/src/renderers/components/section.tsx
```diff
@@ -3,15 +3,19 @@ import { BasicComponentWithPassthroughChildren } from "../types";
...
+    const displayName = `Section${codeNumber ? ` ${codeNumber}.` : ""}`;
 
-    const title = titleElmId != null ? <Element id={titleElmId} /> : "Section";
+    const title = titleElmId != null ? <Element id={titleElmId} /> : "";
 
     return (
         <div className="section" ref={visibilityRef}>
-            <h3>{title}</h3>
+            <h3>
+                {displayName} {title}
```
dqnykamp, line 17: Eventually, we want the ability to suppress `displayName` or just the word `Section`. Not that we need to add that yet, but just noting that will be a future requirement. (Also adjusting the length of the look-back in `codeNumber`.)

(Since DoenetML is typically used in small activities, I was asked to suppress `displayName` by default.)

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
```diff
@@ -106,62 +245,45 @@ mod custom_props {
...
+                    container: PropSource::Me,
+                    filter: Rc::new(Op::And(
+                        // This is what would be normally included in rendered children
+                        Op::Or(
+                            // Keep things without a "hidden" prop
+                            OpNot(ContentFilter::HasPropMatchingProfile(PropProfile::Hidden)),
+                            // Keep things with a "hidden != true" prop
+                            ContentFilter::HasPropMatchingProfileAndCondition(
+                                PropProfile::Hidden,
+                                Cond::NotEq(PropValue::Boolean(true)),
```
dqnykamp, line 266: We can change this to `hidden==false` now.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/component_refs.rs
```diff
@@ -0,0 +1,40 @@
...
+/// and then use the general array prop mechanism to refer to multiple components
+#[derive(Debug, Clone, Default, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ComponentRefs(pub Vec<ComponentIdx>);
+
+/// A reference to a single component
+#[derive(Debug, Clone, Default, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ComponentRef(pub ComponentIdx);
+
+impl ComponentRef {
+    pub fn as_content_ref(&self) -> ContentRef {
```
dqnykamp, line 17: This is identical to the `impl From<ComponentRef> for ContentRef` in `content_refs.rs`, presumably created as a convenience for cases such as in `SerialNumberProp`. Just curious for rationale for knowing when to create something like `as_content_ref()` rather than using `into()`.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
```diff
@@ -84,15 +88,150 @@ impl PropGetUpdater for SectionProps {
...
+            fn data_queries(&self) -> Vec<DataQuery> {
+                RequiredData::to_data_queries()
+            }
+            fn calculate(&self, data: DataQueryResults) -> PropCalcResult<Self::PropType> {
+                let required_data = RequiredData::try_from_data_query_results(data).unwrap();
+                // Create a "." separated string of our ancestors' code number followed by our serial number.
+                let code_number = required_data
+                    .nearest_ancestor_code_number
+                    .iter()
+                    .map(|serial_number| serial_number.value.to_string())
```
dqnykamp, line 154: argument of map is actually a `code_number`

packages/doenetml-worker-rust/lib-doenetml-core/src/components/types.rs
```diff
@@ -98,6 +99,51 @@ impl<T: IntoGraphNode> From<T> for ComponentIdx {
     }
 }
 
+/// The index of the component in `Core.components`
```
dqnykamp, line 102: Need to correct docstring.

It looks like `StringIdx` is used only in `ContentRef` and it can be converted into a `GraphNode`. I don't believe we need to derive `Hash` on it.

I can't put a comment on line 52 as not part of this changeset. But, we should update the docstring to refer to the index of the component in `DocumentStructure.components`.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/types.rs
```diff
@@ -120,6 +166,7 @@ pub struct PropSource {
     /// For example, given `<textInput name="i"/>`, a direct ref would be `$i.value` by itself,
     /// unlike `<text extend="$i.value"/>`.
     ///
+    /// XXX: Incorrect comment?
```
dqnykamp, line 169: Right, it's an outdated comment. It is no longer done in the DataQuery but accomplished when we build the structure graph itself.

Attempt at a revised comment:
If we are extending from a direct ref,
we need to add the referenced prop as a child in the structure graph
because an actual child component for the prop is not created.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/types.rs
siefkenj: It turns out that you use `Hash` in `get_flat_dast_updates`

packages/doenetml-prototype/src/renderers/components/section.tsx
siefkenj: Yep! This is just the workaround for the moment. Things will get more detailed when we have all the props implemented.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
siefkenj: I originally had this 1 indexed and then I changed it because: sometimes the labels are not numbers. If we are indexing into an array of `['A', 'B', 'C', ...]` then we want 0-indexed stuff.

I don't think we should expose this right now. I suspect the user will want the displayed serial_number (which could be a letter, etc.) rather than the actual one.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/types.rs
dqnykamp: Yes, we need to have `ComponentIdx` hashed for the current implementation of flat dast updates, but not `StringIdx`.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/section.rs
dqnykamp: Good point. I hadn't thought about non-numbers. Then, I'd just correct the description. Makes sense to come back to the issue of what to expose at a later point when more pieces are in place.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
```diff
@@ -76,10 +73,36 @@ impl DocumentModel {
...
+                assert_ne!(
+                    prop_node, to,
+                    "Self-loop detected; DataQuery requested a prop that is the same as the origin prop."
+                );
+                dependency_graph.add_edge(from, to);
+                linked_nodes.push(to);
+            }
+        };
+
+        // Resolve a `PropComponent` to a component index.
```
dqnykamp, line 90: Prop component is now prop source.

dqnykamp:
Not in this change set, but a clean up we could sneak into this: The docstring for `get_component_prop()`, line 119 of `graph_node.rs`, incorrectly refers to attributes, as it was apparently copied from the previous function. Same goes for the panic message and variable names later in the function.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation_helpers/data_query_prop.rs
```diff
@@ -0,0 +1,37 @@
+use crate::{
+    components::types::{ComponentIdx, PropPointer},
+    core::document_structure::DocumentStructure,
+    graph_node::GraphNode,
+    props::PropSpecifier,
+};
+
+/// Process a `DataQuery::Prop`.
+///
+/// Returns a vector of edges (of the form `(from, to)`) that should be added to the dependency graph.
```
dqnykamp, line 10: Add something to the effect that this vector will be of length 0 or 1.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation_helpers/data_query_prop.rs
```diff
@@ -0,0 +1,37 @@
...
+            .map(|prop_pointer| prop_pointer.local_prop_idx),
+    };
+
+    if let Some(local_prop_idx) = local_prop_idx {
+        let target_prop_node = PropPointer {
+            component_idx,
+            local_prop_idx,
+        }
+        .into_prop_node(document_structure);
+        ret.push((query_node, target_prop_node));
```
dqnykamp, line 33: The fact that we have length zero or one would be slightly clearer if it were
```
if () {
  ...
  vec![(query_node, target_prop_node)]
} else {
  vec![]
}

Comment (dqnykamp) [reply packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation_helpers/data_query_prop.rs:33]: Then you can get rid of a `mut`, too! :)

File: packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation_helpers/data_query_pick_prop.rs [hunk 22]
```diff
@@ -0,0 +1,47 @@
...
+            let prop = document_structure.get_prop_definition(node);
+
+            let profile = prop.meta.profile;
+            if profile.is_some() && match_profiles.contains(&profile.unwrap()) {
+                return Some(node);
+            }
+        }
+        GraphNode::State(_) | GraphNode::Virtual(_) | GraphNode::Query(_) => {
+            unreachable!(
+                "Cannot have GraphNode of type {:?} as a content child of a component's children.",
```
dqnykamp, line 40: This panic message needs to be generalized to refer to picking a prop or something like that.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query.rs
```diff
@@ -81,6 +83,19 @@ pub enum PropSpecifier {
...
 }
 
+/// The source of a  [`DataQuery::PickProp`].
+#[derive(Debug, Clone, PartialEq)]
+pub enum PickPropSource {
+    /// Search for the props on the children of the querying component.
+    Children,
+    /// Search for the prop on ancestors of the querying component.
+    /// The results will be ordered so that the closest ancestor is first in the result.
+    Ancestors,
```
dqnykamp, line 93: Given that you deleted `AncestorRefs` due to its lack of use, you might as well delete `PickPropSource::Ancestors` for the same reason. Unlike descendants, which I know we'll need, I'm not sure if we'll have the need to grab all ancestors.

packages/doenetml-prototype/src/renderers/components/boolean.tsx
```diff
@@ -1,7 +1,8 @@
 import React from "react";
 import { BasicComponent } from "../types";
+import type { BooleanProps } from "@doenet/doenetml-worker-rust";
```
dqnykamp, line 3: cool!

packages/doenetml-worker-rust/lib-doenetml-core/src/components/types.rs
dqnykamp: Still need to correct this docstring and the docstirng in line 52.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
dqnykamp: Could you change the `PropComponent` comment and the variables `prop_component` and `resolve_prop_component` to use "source"?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/filters.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Satisfied if either `Left` or `Right` are satisfied
+    Or(Left, Right),
+}
+
+/// Not operation used to build complex queries
+/// Satisfied only if `T` is not satisfied
+#[derive(Debug, Clone, PartialEq)]
+pub struct OpNot<T>(pub T);
+
+impl<T, Left, Right> ApplyTest<T, GraphNode> for Op<Left, Right>
```
dqnykamp, line 38: It's confusing how `Op` is implemented with `GraphNode` and `OpNot` is implemented with generic `D`. You run into trouble with the generic `D` only for cases where the return type of `accumulate_deps` cannot be inferred.

My suggestion is to make `Op` and `OpNot` be implemented with generic `D`, since they can always be inferred from their arguments. Then, make `bool` (as well as `Cond`) be implemented only for `GraphNode`. Those are the two troublesome implementations as they don't have dependencies. (And it's more pleasing to have `Op` and `OpNot` symmetric, especially if we eventually want to combine them.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/filters.test.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    let op = Op::And(true, true);
+    assert_eq!(op.apply_test(&true), true);
+
+    let op = Op::Or(true, false);
+    assert_eq!(op.apply_test(&true), true);
+
+    let op = Op::Or(OpNot(Op::And(true, false)), false);
+    assert_eq!(op.apply_test(&true), true);
+}
```
dqnykamp, line 20: If you make `bool` be the one that is implemented only for `GraphNode`, then you can also add this test:
```
    let op = OpNot(true);
    assert_eq!(op.apply_test(&true), false);
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// The value of the matched prop is tested against `Cond` and only matches
+    /// if `Cond` is true.
+    HasPropMatchingProfileAndCondition(PropProfile, Cond<PropValue>),
+    /// Match content is a component.
+    IsComponent,
+    /// Match content is a string.
+    IsString,
+}
+
+/// Information needed to determine of `node` matches a given filter condition.
```
dqnykamp, line 26: of -> if

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+                _ => false,
+            },
+            ContentFilter::HasPropMatchingProfile(profile) => match node {
+                GraphNode::Component(component_idx) => {
+                    let prop =
+                        document_model.get_component_prop_by_profile(component_idx, &[*profile]);
+                    prop.is_some()
+                }
+                _ => {
+                    // `PropProfile::String` is the only profile that non-component nodes can match.
```
dqnykamp, line 58: Maybe clearer if this were in a `GraphNode::String(_)` match arm?

If we have a `GraphNode::String(_)`, it should return true if we have a `String` or  `LiteralString` prop profile. I don't have a test checking this as there is no longer a component using `LiteralString`. Is this an argument for removing `LiteralString` for now?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+                    });
+
+                    prop.map(|prop| {
+                        let value = prop.value;
+                        // Test the condition
+                        cond.apply_test(&value)
+                    })
+                    .unwrap_or(false)
+                }
+                _ => false,
```
dqnykamp, line 85: We need a `GraphNode::String(_)` match arm. If `profile` is `String` or `LiteralString`, then we need to get the value of the string as a `PropValue` and then apply `cond` to that value.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn accumulate_deps(&self, data: &FilterData<'_>) -> Vec<GraphNode> {
+        // Warning: **Do not get prop values in this function**
+        // We won't initialize their deps here. Only return the required prop nodes.
+        let node = data.node;
+        let document_model = data.document_model;
+
+        match node {
+            GraphNode::Component(_) => {}
+            _ => {
+                // The only allowable deps are from props. If we are here, we're querying a non-component
```
dqnykamp, line 101: If we allow one to filter by the value of a string, as suggested above, then we'd need to add the string as a dependency, as we'd have to rerun the query if the value of the string changed.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/filters.test.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    // Match on multiple props
+    let filter = Op::Or(cond1.clone(), cond2.clone()).bind(query_node, document_model);
+    let deps = content_children
+        .iter()
+        .cloned()
+        .map(|n| filter.accumulate_deps(&n))
+        .collect::<Vec<_>>();
+
+    // The `value` props of the text components are what we're looking for.
```
dqnykamp, line 270: `value` and `hidden` props

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/prop_calculation.rs
```diff
@@ -136,85 +136,77 @@ impl DocumentModel {
...
-                                            .get_prop_unchecked(prop_node, query_node)
-                                            .value;
+        // Get the prop pointer to the prop that owns the current query.
+        let get_prop_pointer = || {
+            let prop_node = self.get_nearest_prop_ancestor_of_query(query_node);
+            let prop_node = prop_node.expect("Query node was not owned by a unique prop.");
+            self.get_prop_pointer(prop_node)
+        };
+        // Resolve a `PropSource` to a component index.
+        let resolve_prop_component = |prop_component: &PropSource| match prop_component {
```
dqnykamp, line 146: `..prop_component` -> `..prop_source`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/alias_prop.rs
```diff
@@ -32,19 +32,21 @@ where
     T: Default + Clone + TryFrom<PropValue> + std::fmt::Debug + 'static,
     PropValue: From<T>,
     <T as TryFrom<PropValue>>::Error: std::fmt::Debug,
+    <T as TryFrom<PropValue>>::Error: std::fmt::Display,
```
dqnykamp, line 35: For clarify, use the same format as the below cases (lines 49, 58, etc.):
```<T as TryFrom<PropValue>>::Error: std::fmt::Display + std::fmt::Debug,```

File: packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs [hunk 34]
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug)]
+pub struct ComponentRefProp {
+    /// The data query that indicates how the dependencies of this prop will be created.
+    data_query: DataQuery,
+
+    elements_to_select: Option<ElementsToSelect>,
+}
+
+#[derive(Debug)]
+enum ElementsToSelect {
```
dqnykamp, line 19: `ElementToSelect`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
```diff
@@ -0,0 +1,95 @@
...
+    state::types::{component_refs::ComponentRef, content_refs::ContentRef},
+};
+
+/// A prop that references a single component
+#[derive(Debug)]
+pub struct ComponentRefProp {
+    /// The data query that indicates how the dependencies of this prop will be created.
+    data_query: DataQuery,
+
+    elements_to_select: Option<ElementsToSelect>,
```
dqnykamp, line 15: `element_to_select`

Or maybe if we are using `component` rather than `element` now, then `component_to_select`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn new_self_ref() -> Self {
+        ComponentRefProp {
+            data_query: DataQuery::SelfRef,
+            elements_to_select: None,
+        }
+    }
+}
+
+impl PropUpdater for ComponentRefProp {
```
dqnykamp, line 55: It's interesting how here, and in `component_refs_prop`, we're not using the `RequiredData` structure. My inclination would be to convert these to use that structure and `TryFromDataQueryResults`, etc., macros as well, so that everything is consistent and we have a single pattern for developers to follow.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/independent_prop.rs
```diff
@@ -26,18 +26,20 @@ where
     T: Default + Clone + TryFrom<PropValue> + std::fmt::Debug + 'static,
     PropValue: From<T>,
     <T as TryFrom<PropValue>>::Error: std::fmt::Debug,
+    <T as TryFrom<PropValue>>::Error: std::fmt::Display,
```
dqnykamp, line 29: For clarify, use the same format as the below cases (lines 42, 51, etc.):
<T as TryFrom<PropValue>>::Error: std::fmt::Display + std::fmt::Debug,

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/rendered_children_passthrough_prop.rs
```diff
@@ -34,7 +36,7 @@ impl RenderedChildrenPassthroughProp {
 }
 
 impl PropUpdater for RenderedChildrenPassthroughProp {
```
dqnykamp, line 38: convert to `RequiredData` pattern with macros?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
siefkenj: If we remove it, will that mean we can no longer force certain attrs to be defined with an absence of references?

dqnykamp: Right. That's a good reason to keep `LiteralString`, otherwise we can't distinguish a reference turned text from a string.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
siefkenj: I think the original reason it was different was because it was combined with data queries that might return different length vecs. I can switch to using `RequiredData` and see what happens.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
siefkenj: Is the mechanism to track string changes set up yet?

dqnykamp: Yep, it just works given that strings are stored in a `StringCache` that is based on the `PropCache`.

dqnykamp: The `<textInput>` tests involving actions modify string props and string children.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn refs_query(arg: &ComponentRefProp) -> DataQuery {
+        arg.data_query.clone()
+    }
+}
+
+impl PropUpdater for ComponentRefProp {
+    type PropType = prop_type::ComponentRef;
+
+    fn data_queries(&self) -> Vec<DataQuery> {
+        vec![self.data_query.clone()]
```
dqnykamp, line 71: Here's a suggested patch. (Due to a known issue, I apparently catch attach a .patch file when using Linux, so I pasted here.)

```
diff --git a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
index c35ec9cf7..cafd8eeb7 100644
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
@@ -54,13 +54,13 @@ impl ComponentRefProp {
 
 /// Structure to hold data generated from the data queries
 #[derive(TryFromDataQueryResults, Debug)]
-#[data_query(query_trait = DataQueries, pass_data = &ComponentRefProp)]
+#[data_query(query_trait = DataQueries, pass_data = &DataQuery)]
 struct RequiredData {
     refs: PropView<prop_type::ContentRefs>,
 }
 impl DataQueries for RequiredData {
-    fn refs_query(arg: &ComponentRefProp) -> DataQuery {
-        arg.data_query.clone()
+    fn refs_query(query: &DataQuery) -> DataQuery {
+        query.clone()
     }
 }
 
@@ -68,7 +68,7 @@ impl PropUpdater for ComponentRefProp {
     type PropType = prop_type::ComponentRef;
 
     fn data_queries(&self) -> Vec<DataQuery> {
-        vec![self.data_query.clone()]
+        RequiredData::data_queries_vec(&self.data_query)
     }
 
     fn calculate(&self, data: DataQueryResults) -> PropCalcResult<Self::PropType> {
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
```diff
@@ -0,0 +1,105 @@
...
+                        &ContentRef::Component(idx) => {
+                             ComponentRef::from(idx)
+                        },
+                        _ => unreachable!("data queries for element refs prop should return component graph nodes, found {:?}", content_ref.clone())
+                    }
+                );
+                let component = match self.component_to_select {
+                    Some(ComponentToSelect::First) => elements.next(),
+                    Some(ComponentToSelect::Last) => elements.last(),
+                    None => unreachable!("component_to_select should be Some"),
```
dqnykamp, line 96: With refusing `None`, the `new_self_ref()` will break. I guess we don't use it anywhere right now. Do we need `new_self_ref()`? If we remove `new_self_ref()`, then do we want to remove the option from `component_to_select`?

dqnykamp: Oh, I guess you are just missing a match arm for `self.data_query` to catch `DataQuery::SelfRef`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
dqnykamp: A patch:
```
diff --git a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
index d3ae1cc6f..f08cec7e0 100644
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
@@ -55,14 +55,9 @@ impl ApplyTest<FilterData<'_>, GraphNode> for ContentFilter {
                     prop.is_some()
                 }
                 GraphNode::String(_) => {
-                    // `PropProfile::String` is the only profile that non-component nodes can match.
-                    // We special case it here.
-                    if matches!(profile, PropProfile::String)
-                        && matches!(node, GraphNode::String(_))
-                    {
-                        return true;
-                    }
-                    false
+                    // `PropProfile::String` and `PropProfile::LiteralString` are the only profiles that non-component nodes can match.
+                    // We special case them here.
+                    matches!(profile, PropProfile::String | PropProfile::LiteralString)
                 }
                 _ => unreachable!("Only components and strings can match profiles"),
             },
```

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
```diff
@@ -0,0 +1,159 @@
...
+
+                    prop.map(|prop| {
+                        let value = prop.value;
+                        // Test the condition
+                        cond.apply_test(&value)
+                    })
+                    .unwrap_or(false)
+                }
+                GraphNode::String(_) => {
+                    // `PropProfile::String` is the only profile that non-component nodes can match.
```
dqnykamp, line 87: patch:
```
diff --git a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
index d3ae1cc6f..b6041dded 100644
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/content_filter.rs
@@ -84,11 +84,9 @@ impl ApplyTest<FilterData<'_>, GraphNode> for ContentFilter {
                     .unwrap_or(false)
                 }
                 GraphNode::String(_) => {
-                    // `PropProfile::String` is the only profile that non-component nodes can match.
-                    // We special case it here.
-                    if !matches!(profile, PropProfile::String)
-                        || !matches!(profile, PropProfile::LiteralString)
-                    {
+                    // `PropProfile::String` and `PropProfile::LiteralString` are the only profiles that non-component nodes can match.
+                    // We special case them here.
+                    if !matches!(profile, PropProfile::String | PropProfile::LiteralString) {
                         return false;
                     }
                     let str_val = document_model.get_string_value(node);
```

PR #142, opened 2024-04-12
runestone splice events

jaltekruse:


jaltekruse:
Going to move the mathjax upgrade to a separate PR so we can just merge the splice changes. That part still needs some more work and testing because it caused some performance regressions.

dqnykamp (approved):

PR #144, opened 2024-04-17
Revive tests

dqnykamp:


packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/document_model.test.rs
```diff
@@ -0,0 +1,620 @@
...
+
+#[test]
+fn blank_attribute_is_empty_string() {
+    let dast_root = dast_root_no_position(r#"<text hide />"#);
+    let mut core = Core::new();
+    core.init_from_dast_root(&dast_root);
+
+    let text_node = GraphNode::Component(1);
+
+    let hide_idx = TextAttributes::Hide.idx();
```
siefkenj, line 23: I suppose this is nicer than `Text::ATTRIBUTE_NAMES.position("hide").unwrap()`

dqnykamp: Aah, I was looking for something on `TextAttributes` rather than `Text`. Since this is only used in testing, I'd rather use this than add another macro function.

PR #145, opened 2024-04-18
Add a `preserve_refs` annotation for attributes

siefkenj:
Attributes can now be annotated with `preserve_refs`. If so, any `$foo` refs that appear in the specified attribute will not be expanded to clones of their referent. Instead, they will be converted to special `_ref` components which hold the `idx` of their referent.

### DIFF
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
@@ -320,7 +320,9 @@ impl ComponentBuilder {
             .take(normalized_root.nodes.len())
             .collect();
 
-        // Creating the nodes lowest-index first should lead to less queueing than the other way around.
+        // Creating the nodes lowest-index first. This
+        //  1. is assumed by the algorithm which specializes the `extend` expansion/resolution, and
+        //  2. should lead to less queueing than the other way around.
         let mut queue: Vec<usize> = Vec::from_iter((0..components.len()).rev());
 
         while let Some(idx) = queue.pop() {
@@ -377,7 +379,7 @@ impl ComponentBuilder {
                     },
                 );
 
-                // A some of a component's attributes may specify that refs in those attributes should not be expanded.
+                // Some of a component's attributes may specify that refs in those attributes should not be expanded.
                 // We mark the corresponding components.
                 // Because the component's attributes have not been processed yet, we need to look up the corresponding
                 // attributes from `node`
@@ -417,6 +419,7 @@ impl ComponentBuilder {
                         };
                         if ref_resolution.unresolved_path.is_some() {
                             // XXX: this should result in the component being converted into an error, not a hard panic.
+                            // TODO: Not only should this not be an error, but sometimes it is valid. For example `<point name="p">(3,2)</point><updateValue target="$p.x" newValue="$p.x+1" />`
                             panic!("Cannot preserve_refs if there is a remaining path part. Remaining: {:?}", ref_resolution.unresolved_path);
                         }
                         // If we made it here, we are a ref pointing to a component and we should not actually be expanded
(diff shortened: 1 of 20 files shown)


packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
```diff
@@ -362,7 +376,60 @@ impl ComponentBuilder {
                         unrecognized_attributes: HashMap::new(),
                     },
                 );
-                if let Some(Source::Ref(ref_resolution)) = &elm.extending {
+
+                // A some of a component's attributes may specify that refs in those attributes should not be expanded.
```
dqnykamp, line 380: A some -> Some

packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
```diff
@@ -362,7 +376,60 @@ impl ComponentBuilder {
...
+                        match child {
+                            UntaggedContent::Ref(idx) => {
+                                self.component_preserve_refs[ComponentIdx::from(*idx)] = true;
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+
+                if self.component_preserve_refs[component.get_idx()] {
```
dqnykamp, line 407: The reason we know we will have set `component_preserve_refs` before we get to `component` is because we are processing components starting with the lowest index, which means we'll process a component before we process the components inside its attributes. We should include a comment here to that effect.

More importantly, though, is to add a comment before line 324 indicating the requirement that we attempt to create nodes lowest-index first. The current comment before line 324 would lead one to think it is optional (as it was when that comment was created).

I originally thought that we'd run into trouble with self references, such as `<xref name="x" ref="$x" />`, as in this case the attribute will get created before the `<xref>`. However, the critical code from just above will get run the first time through trying to make `<xref>` even though that try ends up with an `Err()`. I recommend some comment indicating that we need to make sure the above does run even if creating `<xref>` fails. More importantly, we should have a test checking that it works even with such a self-reference.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
```diff
@@ -362,7 +376,60 @@ impl ComponentBuilder {
...
+                    // are protected and should be converted into a special `_ref` component instead of being linked
+                    // to their extend referent.
+                    if let Some(ref_source) = &elm.extending {
+                        let ref_resolution = match ref_source {
+                            Source::Ref(ref_resolution) => ref_resolution,
+                            _ => unreachable!("Encountered an extend of style <foo extend='$xxx' /> inside of an attribute. It should be syntactically impossible to parse such an element in an attribute."),
+                        };
+                        if ref_resolution.unresolved_path.is_some() {
+                            // XXX: this should result in the component being converted into an error, not a hard panic.
+                            panic!("Cannot preserve_refs if there is a remaining path part. Remaining: {:?}", ref_resolution.unresolved_path);
```
dqnykamp, line 420: Actually, we will eventually need the ability to preserve refs even with an remaining part. For example,
```
<point name="p">(3,2)</point>
<updateValue target="$p.x" newValue="$p.x+1" />
```
should increment the x-component of the point on each click of the button. And for `target` we want to preserve refs so we can find the target to update it.

Not that we need to do that now, but it should be a TODO if not an XXX.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/component_ref_prop.rs
```diff
@@ -102,6 +114,11 @@ impl PropUpdater for ComponentRefProp {
                     data.vec[0].values[0].to_owned().into_prop_view();
                 PropCalcResult::Calculated(component_ref.value)
             }
+            DataQuery::Attribute { .. } => {
```
dqnykamp, line 117: The comment on line 87 is now outdated.

dqnykamp:
I like how this worked out!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
siefkenj: Not exactly sure what the last paragraph is asking me to do, but I added a test of a self-referential xref.

dqnykamp (approved):

PR #149, opened 2024-04-20
Xrefs to Sections

siefkenj:
This PR allows for `<section name="s"/><xref ref="$s"/> to expand to an appropriately named reference (link).

The only controversial part of this PR may be the addition of `PropSource::StaticComponentRef` which allows you to use a prop whose content is a single `ComponentRef` as the source of the data query. The catch is, the data query will _not_ be updated if this prop changes. Instead the DQ will always be linked to the referent from the first time it was called.

### DIFF
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
@@ -1,5 +1,5 @@
 use crate::{
-    components::prelude::DataQuery,
+    components::{prelude::DataQuery, types::PropPointer},
     props::{FilterData, PickPropSource, PropSource, PropSpecifier},
 };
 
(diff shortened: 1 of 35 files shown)


packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
```diff
@@ -88,14 +88,39 @@ impl DocumentModel {
...
+                PropSource::ByIdx(component_idx) => *component_idx,
+                PropSource::StaticComponentRef(local_prop_idx) => {
+                    // This is the prop that contains the ref.
+                    let prop_node = self.prop_pointer_to_prop_node(PropPointer {
+                        component_idx: prop_pointer.component_idx,
+                        local_prop_idx: *local_prop_idx,
+                    });
+                    // We could encounter a circularly-recursive call here if there was a programmer error.
+                    // If all goes well, `prop` will be a non-null ComponentRef.
+                    let prop = self.get_prop_untracked(prop_node, query_node);
```
dqnykamp, line 108: I'm really worried about calling `get_prop_untracked`, which will call `resolve_prop` while we are in the middle of `add_data_query` which is being called by `resolve_prop`. This makes it hard to reason what the code is doing and troubleshoot errors.

For example, what happens, if while you are looping over the data queries in line 45 of `prop_calculation.rs` and inside a call to `add_data_query`, the call to `get_prop_unchecked` ends up resolving one of the data queries further in the loop of line 45? Then, one will end up calling `add_data_query` twice on that data query, which will mess up the dependencies.

I'm not sure that my scenario is actually possible. But, this pattern opens up a lot more possibilities of code paths while we're creating the dependency graph that I feel could bite us hard as the code gets more complicated.

I'll keep thinking about this to see if there is a better pattern to get this result.

dqnykamp: What about this algorithm?

When a data query needs the value of a prop in order to create its dependency structure, then `add_data_query` checks if it is resolved. If it is resolved, then it gets the value as usual and proceeds.

However, if `add_data_query` needs the value of an unresolved `dependent_prop`, it aborts and reports that it needs  `dependent_prop` resolved. Then, `resolve_prop` adds the failed prop back to `resolve_stack` and subsequently adds `dependent_prop` to `resolve_stack` so that  `dependent_prop` will be resolved before the original prop.

The only wrinkle is that, the second time the original prop is popped off the stack, it may have already run `add_data_query` on some of its queries. So, we'd have to include bookkeeping about whether or not a data query has been added. (This bookkeeping might be necessary anyway when we have to recalculate queries in response to components to being dynamically added.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/dependency_creation.rs
```diff
@@ -137,7 +162,14 @@ impl DocumentModel {
                         }
                     }
                 }
-                let component_idx = resolve_prop_component(&source);
+                let component_idx = match resolve_prop_component(&source) {
+                    Ok(idx) => idx,
+                    Err(_) => {
+                        // If we can't resolve the component, then we can't resolve the prop.
+                        // Avoid a hard panic here.
+                        return linked_nodes;
```
dqnykamp, line 170: If we allow `add_data_query` to return a truncated set of `linked_nodes`, then when we get to `_execute_data_query_with_fresh_deps()` in `prop_calculation.rs`, then it will simply return a smaller set of results in `DataQueryResult`. This mean, in `calculate()` of a `PropUpdater`, it might get fewer results than expected. Are there scenarios where this is going to cause a problem?

If the `RequiredData` had a field of the form `PropView<type>`, then this would lead to a panic when `RequiredData::try_from_data_query_results` is unwrapped. We wouldn't want a case where we may or may not get a panic depending on what the author wrote.

I see you coded defensively in `xref.rs` by making `label` be a `Vec`. Eventually, it'd be nice if you could make it an `Option` instead. Then, if we know which types of data queries could potentially lead to failure, we could document that one should make the required data be an option.

Could we make it always panic if one codes a case where the `RequiredData` isn't an Option/Vec and the data query might return nothing?

Ideally, we want to make it hard for developers to create components that work most of the time but then fail a small percentage of time.

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/xref.rs
```diff
@@ -49,3 +45,60 @@ fn xref_referring_to_itself_wont_crash_the_system() {
...
+    let xref_idx = ComponentIdx::from(2);
+
+    let prop = core.get_prop(xref_idx, XrefProps::DisplayText.local_idx());
+    let pv: PropView<prop_type::String> = prop.into_prop_view();
+    assert_eq!(*pv.value, "Section 1".to_string());
+}
+
+#[test]
+fn xref_ref_can_make_display_text_with_custom_children() {
+    let dast_root = dast_root_no_position(r#"<section name="foo"/><xref ref="$foo">Foo</xref>"#);
```
dqnykamp, line 66: From this test, I started out confused what the `Foo` string child is supposed to do. Maybe showing that it is the rendered child will at least make it seem like it wasn't lost?

siefkenj: In the UI this should render as `Foo 1`, with a space between the `Foo` and the `1`

PR #150, opened 2024-04-24
Add support for `<ol>`

siefkenj:
This PR adds support for `<ol>`-style lists. They can also be referenced via `<xref>`


packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
```diff
@@ -0,0 +1,214 @@
...
+            ListMarker::UpperRoman => Self::Decimal { start: 1 },
+            ListMarker::Disc => Self::Circle,
+            ListMarker::Circle => Self::Square,
+            ListMarker::Square => Self::Disc,
+        }
+    }
+
+    /// Convert `index` into the string representation given by this marker.
+    /// `0` maps to the number `"0"`. Lists are assumed to start indexing at `1`,
+    /// though when using `ListMarker::Decimal`, any integer value of `index` can be used.
```
dqnykamp, line 43: Apparently, you can use any integer value for the roman markers, as well (and will get empty strings for non-positive index for the alpha markers).

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
```diff
@@ -0,0 +1,214 @@
+/// A list marker. This is used by both `<ol>` and `<ul>` elements.
+/// They are named similarly to their [HTML equivalents](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type).
+#[derive(Debug, Clone, Copy, PartialEq, serde::Serialize, serde::Deserialize)]
+#[cfg_attr(feature = "web", derive(tsify_next::Tsify))]
+pub enum ListMarker {
+    // Ordered list
+    /// Label with a number, starting at `start`. e.g. `1`, `2`, `3`, ...
+    Decimal { start: u8 },
```
dqnykamp, line 8: I am confused as I cannot find the use of the value of this `start` field anywhere. Does it do anything?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
```diff
@@ -0,0 +1,214 @@
...
+            ListMarker::LowerAlpha
+        );
+        assert_eq!(ListMarker::LowerAlpha.next_marker(), ListMarker::LowerRoman);
+        assert_eq!(ListMarker::LowerRoman.next_marker(), ListMarker::UpperAlpha);
+        assert_eq!(ListMarker::UpperAlpha.next_marker(), ListMarker::UpperRoman);
+    }
+
+    #[test]
+    fn can_convert_to_strings() {
+        assert_eq!(ListMarker::Decimal { start: 1 }.index_to_string(1), "1");
```
dqnykamp, line 197: I suspect you'll also get "1" if you use another value for `start`.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
siefkenj: It gets used as long as the `start` attribute is not used.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
siefkenj: Yes. I wasn't quite sure what to do here. Should I document this behavior? Should I allow negative letters as well?

packages/doenetml-worker-rust/lib-doenetml-core/src/components/special/_external.rs
```diff
@@ -1,15 +1,45 @@
...
+    const PROP_IS_PUBLICS: &'static [bool] = &[false];
+    const PROP_VALUE_TYPES: &'static [PropValueType] = &[PropValueType::ContentRefs];
+    const DEFAULT_PROP: Option<LocalPropIdx> = None;
+}
+
+///The props for this component are: `renderedChildren`
+#[derive(Debug, Clone, Copy)]
+pub enum _ExternalProps {
+    /**
+    - Name: `"renderedChildren"`
```
dqnykamp, line 28: This comment is to indicate what the setting would be if the props were created from a macro? Do we want to just go ahead and derive `_External` using the macro? It would seem easier to maintain. I can't remember if there is other special case treatment needed that precludes the use of the macro.

siefkenj: I generated from the macro and copy-pasted. We cannot use the macro because the name isn't static.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/core.rs
```diff
@@ -35,14 +37,15 @@ impl Core {
     ///
```
dqnykamp, line 37: Add some description of what the return value is. (It apparently is used only in tests.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
dqnykamp: I'd just document it and only bother extending to negative letters if there is some use case.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
dqnykamp: OK, it's just information about the list marker that can be used by outside consumers.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/enum_prop.rs
```diff
@@ -0,0 +1,75 @@
+use std::{marker::PhantomData, rc::Rc};
+
+use crate::{components::prelude::*, props::UpdaterObject};
+
+/// A prop that represents a value from an enum. The enum must implement `From<String>`.
```
dqnykamp, line 5: `From<String>` -> `From<&str>`

packages/doenetml-worker-rust/lib-doenetml-core/src/components/special/_external.rs
dqnykamp: Oh, right, that was the special case.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/list_marker.rs
siefkenj: The line right below uses another value. The `1` is completely ignored when you call `index_to_string`.

PR #152, opened 2024-04-25
Serialize `ContentRefs` into `FlatDastElementContent`

siefkenj:
If `ContentRefs` are serialized directly, then `ContentRef::String(usize)` cannot be used by Javascript since the string table is not set to JS. This PR changes the serialization of `ContentRefs` so that they are serialized as `FlatDastElementContent`. In particular, strings become literal strings with their value.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_renderer/to_flat_dast.rs
```diff
@@ -9,13 +10,21 @@ use crate::{
...
+        ForRenderProps,
     },
     graph::directed_graph::Taggable,
     props::{cache::PropWithMeta, PropProfile, PropValue},
+    state::types::content_refs::ContentRef,
 };
 
-use super::{super::graph_node::GraphNode, DocumentRenderer};
+/// When to include a prop in the serialized rendered props.
+pub enum SerializeCondition {
```
dqnykamp, line 22: Don't want to get burned anymore by misreading a boolean condition?

siefkenj: You know me so well!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/content_refs.rs
```diff
@@ -41,15 +41,42 @@ impl From<Vec<ContentRef>> for ContentRefs {
...
+// Note: we do not do `#[cfg_attr(feature = "web", derive(tsify_next::Tsify))]` here because `ContentRef` is never actually
+// serialized. Instead it is swapped out for `FlatDastElementContent` before serialization. In order to keep the correct types,
+// we manually override.
 pub enum ContentRef {
     Component(ComponentIdx),
     String(StringIdx),
 }
 
+#[cfg(feature = "web")]
+const _: () = {
```
dqnykamp, line 63: What is the `const _: () = {stuff}` trick for?

dqnykamp (approved):

PR #153, opened 2024-04-26
Add DoenetML preprocessor

siefkenj:
This PR adds a preprocessor that swaps some doenetml elements for others. In particular it maps
 - `<section>` -> `<division type="section">`
 - `<chapter>` -> `<division type="chapter">`
 - etc.


packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/division.rs
```diff
@@ -49,19 +50,35 @@ mod component {
...
+               value_type = PropValueType::DivisionType,
+               profile = PropProfile::DivisionType,
+               for_render
+           )]
+        DivisionType,
+
+        /// The value of the `division` attribute.
+        #[prop(
+               value_type = PropValueType::DivisionType,
+               profile = PropProfile::DivisionType,
```
dqnykamp, line 72: I suspect that you want to remove this profile. When another component is looking for `PropProfile::DivisionType`, you want them to just find the `DivisionType` prop.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/division.rs
```diff
@@ -49,19 +50,35 @@ mod component {
...
-        /// the section is referred to by `<xref>`.
+        /// The type of this `<division />`. E.g. "section", "chapter", "subsection", etc.
+        #[prop(
+               value_type = PropValueType::DivisionType,
+               profile = PropProfile::DivisionType,
+               for_render
+           )]
+        DivisionType,
+
+        /// The value of the `division` attribute.
```
dqnykamp, line 69: If we hadn't had a previous conversation about `FooAttr` props for temporary processing, I would have been confused by this. I suggest adding to the docstring of `DivisionTypeAttr` that this is an internal intermediate value used to calculate the `DivisionType` prop.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/division.rs
```diff
@@ -417,4 +455,63 @@ mod custom_props {
...
     }
+
+    pub use division_type::*;
+    mod division_type {
+        use super::*;
+
+        /// The depth of this component in the hierarchy of divisions. E.g., how many
+        /// times this `<section>` is nested in another.
+        #[derive(Debug, Default)]
+        pub struct DivisionTypeProp {}
```
dqnykamp, line 466: If I am reading this correctly, you have a separate `DivisionType` prop that isn't just the `DivisionTypeAttr` prop that is calculated with the general prop update for just this reason: to handle the case where someone omits the `type` attribute and the `<division>` infers its type. Having this explained somewhere would help make it clearer why we have these two props.

dqnykamp: There is a separate question of whether or not we make this more ergonomic for developers who want to do such tweaking to an attribute. In the JS DoenetML, the equivalent to the `attribute` data query does more processing. It knows the type of the attribute so it actually does something similar to the general prop updaters when you request the attribute, giving you something of the attribute type rather than its components.

I can think of a couple possible avenues we could pursue if we want to go this route, but I guess it's a bit beyond the scope of this PR!

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/division.rs
```diff
@@ -0,0 +1,66 @@
...
+    assert_eq!(ty, prop_type::DivisionType::Section);
+
+    let ty: prop_type::DivisionType = core.get_prop_value_typed(
+        core.get_component_index_by_name("d2"),
+        DivisionProps::DivisionType.local_idx(),
+    );
+    assert_eq!(ty, prop_type::DivisionType::Subsection);
+}
+
+#[test]
```
dqnykamp, line 48: I'd like one more test showing that you get the right division type from `<chapter>`, etc.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/division.rs
siefkenj: Yeah, we can discuss this in another PR :-)

packages/doenetml-worker-rust/lib-doenetml-core/tests/by_component/division.rs
siefkenj: What do you mean? On line 65, it tests `<chapter />`.

dqnykamp: Right, it tests the label. But it doesn't test the `DivisionType`. Not that you could get the right label without the right DivisionType.

PR #154, opened 2024-04-26
Make sure components that are referenced in props show up in the render tree

siefkenj:
Some components may be referenced in a `for_render` prop of type `ComponentRefs`/etc.. These referenced components may not appear as children of any node, but they should still be rendered, because the UI Javascript code may decide to display them.

This PR is based on #153 and marks `in_render_tree` components that are referenced in `for_render` props.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_renderer/to_flat_dast.rs
```diff
@@ -126,6 +129,63 @@ impl DocumentRenderer {
...
+            .get_for_render_prop_pointers(component_idx)
+            .flat_map(|prop_pointer| {
+                if !self.prop_may_contain_children(prop_pointer, document_model) {
+                    return vec![];
+                }
+                let prop = self.get_prop_for_render(prop_pointer, document_model);
+                match prop.value {
+                    PropValue::ComponentRefs(refs) => refs
+                        .iter()
+                        .map(|c| GraphNode::Component(c.as_usize()))
```
dqnykamp, line 148: for consistency/simplicity, use `c.as_graph_node()` here and on link 155

dqnykamp (approved):

PR #155, opened 2024-04-27
Add compile-time checks to make sure PropProfile and PropValueType agree

siefkenj:
Add compile-time checks to make sure PropProfile and PropValueType agree.

### DIFF
--- a/packages/doenetml-worker-rust/lib-rust-macros/src/component_module/items/props.rs
+++ b/packages/doenetml-worker-rust/lib-rust-macros/src/component_module/items/props.rs
@@ -338,6 +338,53 @@ impl PropsEnum {
             };
         }
     }
+
+    /// Generate compile-time checks that the prop_profile and the prop_value are compatible.
+    fn generate_profile_value_checks(&self) -> TokenStream {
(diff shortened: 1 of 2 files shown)


packages/doenetml-worker-rust/lib-rust-macros/src/component_module/items/props.rs
```diff
@@ -338,6 +338,53 @@ impl PropsEnum {
...
+                    name,
+                    profile_str,
+                    value_type_str
+                );
+                quote! {
+                    if !matches!(
+                        crate::props::prop_profile_to_type(#profile),
+                        #value_type
+                    ) {
+                        panic!(#panic_string);
```
dqnykamp, line 376: This is pretty cool. So the macro panics, which makes the compilation fail. I can't imagine a way to write an automated test for this. I had to try it to see for myself that this works.

siefkenj: Yeah. Since it is compile time, I am not sure how to test it. Maybe it could be testable in a separate crate...

dqnykamp (approved):

PR #156, opened 2024-04-27
Make sure HTML ids are unique

siefkenj:
The same component may appear multiple times as children of different elements (e.g., from extending another component). If this component is a `<division>` it is normally assigned an html id. However, html ids should be globally _unique_.

This PR passes in data about whether a rendered component is `original` or `duplicate`. This data is then used to compute whether or not a rendered component should receive an html id.

It builds off of PR #153


packages/doenetml-prototype/src/renderers/components/division.tsx
```diff
@@ -2,19 +2,25 @@ import React from "react";
 import { BasicComponentWithPassthroughChildren } from "../types";
 import { Element } from "../element";
 import type { DivisionProps } from "@doenet/doenetml-worker-rust";
+import { extendAncestorChain, generateHtmlId } from "../utils";
```
dqnykamp, line 5: extendAncestorChain isn't used here

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/content_refs.rs
```diff
@@ -52,6 +53,56 @@ impl serde::Serialize for ContentRefs {
...
+impl From<Vec<(ContentRef, ElementRefAnnotation)>> for AnnotatedContentRefs {
+    fn from(t: Vec<(ContentRef, ElementRefAnnotation)>) -> Self {
+        AnnotatedContentRefs(t)
+    }
+}
+
+impl serde::Serialize for AnnotatedContentRefs {
+    fn serialize<S: serde::Serializer>(&self, _serializer: S) -> Result<S::Ok, S::Error> {
+        // We cannot properly serialize `ContentRef::String` because we don't have access to the string table here.
+        Err(serde::ser::Error::custom("ContentRefs should not be serialized directly. They should be converted to `FlatDastElementContent` first."))
```
dqnykamp, line 102: ContentRefs -> AnnotatedContentRefs

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/content_refs.rs
```diff
@@ -72,6 +123,8 @@ const _: () = {
     // process and turn all `ContentRef` into `FlatDastElementContent` before serializing.
     #[declare]
     type ContentRef = FlatDastElementContent;
+    #[declare]
```
dqnykamp, line 126: Just curious why this has to be done for ContentRef and AnnotatedContentRefs but not ContentRefs.

siefkenj: `ContentRefs` serializes as an array of `ContentRef`, which is the correct type for Typescript :-)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/component_builder.rs
```diff
@@ -633,10 +637,15 @@ impl ComponentBuilder {
         let referent_children_virtual_node = self
             .structure_graph
             .get_component_children_virtual_node(referent_idx);
-        self.structure_graph.prepend_edge(
-            component_children_virtual_node,
-            referent_children_virtual_node,
-        );
+        // We create an intermediate virtual node so that we can mark it. This will indicate its descendants
+        // came from extending another component and aren't "original".
+        let intermediate_virtual_node = self.new_virtual_node();
```
dqnykamp, line 642: Just making sure I understand the new document structure. If a component1 is original, we still have

component1 -> component1-children-virtual-node -> component1's children

But, if we component2 is extending component1, we have
component2 -> component2-children-virtual-node -> marked-intermediate -> component-children-virtual-node
as well as
component2 -> component2-children-virtual-node -> component2's children

siefkenj: Yep. You got it!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/rendered_children_passthrough_prop.rs
```diff
@@ -37,25 +39,32 @@ impl RenderedChildrenPassthroughProp {
...
 
     fn data_queries(&self) -> Vec<DataQuery> {
-        vec![self.data_query.clone()]
+        RequiredData::data_queries_vec(&self.data_query)
     }
 
     fn calculate(&self, data: DataQueryResults) -> PropCalcResult<Self::PropType> {
         let required_data = RequiredData::try_from_data_query_results(data).unwrap();
-        PropCalcResult::Calculated(required_data.refs.value)
+        let children = required_data
```
dqnykamp, line 61: I'm trying to understand why you need all this rigamarole now and you didn't before. `RequiredData.refs.value` is an `Rc<AnnotatedContentRefs>`, which is the same as our `PropType`, i.e., `prop_type::AnnotatedContentRefs`. Why can't you just stick it in the `PropCalcResult::Calculated()`? If I revert this change, it still compiles and the test pass.

siefkenj: I think that is left over from when I was converting `AnnotatedComponentRefs` back to `ComponentRefs`, just to make sure things were working. I will clean it up!

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/prop_calculation.rs
```diff
@@ -320,43 +322,94 @@ impl DocumentModel {
...
                         }
                     }
-                }
+                    // `AnnotatedComponentRefs` are very similar to `ComponentRefs`, but they also include
+                    // information about whether the content is original or comes from extending some other element.
+                    DataQuery::AnnotatedComponentRefs { .. } => {
+                        let content_children =
+                            self.get_component_content_children_annotated(component_idx);
+
+                        let mut content_refs: Vec<(ContentRef, ElementRefAnnotation)> = Vec::new();
```
dqnykamp, line 373: I'd like another name than `content_refs` to clarify that this is more. (I was initially confused how you could do a `content_refs.into()` to get an `AnnotatedContentRefs`.) Maybe something like `content_refs_with_annotations`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/graph_node.rs
```diff
@@ -197,10 +221,72 @@ mod iterators {
...
+    ///
+    /// This also returns a boolean indicating whether the node is marked by `mark`.
+    pub struct MarkedContentChildrenIterator<'a> {
+        graph: &'a Graph,
+        /// Stack storing the nodes for iteration in _reverse_ order.
+        stack: Vec<(GraphNode, bool)>,
+        marker: &'a GraphNodeLookup<bool>,
+    }
+    impl<'a> MarkedContentChildrenIterator<'a> {
+        pub fn new(graph: &'a Graph, start: GraphNode, mark: &'a GraphNodeLookup<bool>) -> Self {
```
dqnykamp, line 237: I think this `mark` is a `marker`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/graph_node.rs
```diff
@@ -197,10 +221,72 @@ mod iterators {
             }
         }
     }
+
+    /// Iterate through the "content" children of a node. That is,
+    /// all the non-virtual children. If a virtual child is detected,
+    /// it's children are recursively iterated over.
+    ///
+    /// This also returns a boolean indicating whether the node is marked by `mark`.
```
dqnykamp, line 229: `mark` -> `marker`

It returns a boolean indicating whether the iterator encountered virtual node ancestor of the node that is marked by `marker`. (Or something indicating that the node is marked based on its virtual node ancestor.)

packages/doenetml-worker-rust/lib-doenetml-core/src/core/graph_node.rs
```diff
@@ -197,10 +221,72 @@ mod iterators {
...
+    impl<'a> MarkedContentChildrenIterator<'a> {
+        pub fn new(graph: &'a Graph, start: GraphNode, mark: &'a GraphNodeLookup<bool>) -> Self {
+            let mut stack = graph
+                .get_children(start)
+                .into_iter()
+                .map(|n| (n, false))
+                .collect::<Vec<_>>();
+            // Order matters and we will be popping values off the end of the stack, so
+            // we reverse it.
+            stack.reverse();
```
dqnykamp, line 245: Is it any more efficient to put the `.rev()` before the `collect()` rather than reversing the vector?

dqnykamp:
Do we have any tests in rust where it actually looks for a `Duplicate` annotation coming from a component that is extending? If so, I missed it.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/graph_node.rs
siefkenj: Maybe!

siefkenj:
> Do we have any tests in rust where it actually looks for a `Duplicate` annotation coming from a component that is extending? If so, I missed it.

Added a test in `to_flat_dast.test.rs` :-)

PR #159, opened 2024-04-29
Add type checking to RequriedData/DataQueries

siefkenj:
This PR adds type checking ensuring consistency between `RequiredData` objects and the data queries they request. The checks are done during testing, because I could not figure out how to do them during runtime.

They work as follows:
 1. Every prop_type derives the `TypeDiscriminant` trait, which gives access to the `PropTypeValue` enum directly from the type.
 2. A function `.guess_return_type:<T>()` was added to data queries which returns a vector of `TypeDiscriminant` of the possible return types from that data query. The `T` must be the type of the component those data queries come from. This allows `PropSource::Me` to properly resolve the types from a `LocalPropIdx`-style query.
 3. Every `RequiredData` struct should derive `TestDataQueryTypes` and specify `#[owning_component(Foo)]` where `Foo` is the component that the prop relates to. The derive macro for this trait _also_ adds a test which calls `TestDataQueryTypes::_test_data_query_types` function. Ultimately, that test will panic if the types are not consistent.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/prop_value.rs
```diff
@@ -68,28 +72,61 @@ pub mod prop_type {
...
-    pub type Number = f64;
-    #[cfg_attr(feature = "web", tsify_next::declare)]
-    pub type Integer = i64;
-    #[cfg_attr(feature = "web", tsify_next::declare)]
-    pub type Boolean = bool;
-    #[cfg_attr(feature = "web", tsify_next::declare)]
-    pub type Math = Rc<MathExpr>;
+    /// A macro that declares a type and implements `TypeDiscriminant`.
+    ///
+    /// For example, `with_discriminant!(Foo, type::Name)` that expands to
```
dqnykamp, line 77: "that expands to" -> "expands to"

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/traits/test_data_query_types.rs
```diff
@@ -0,0 +1,70 @@
...
+pub trait TestDataQueryTypes: TryFromDataQueryResults {
+    /// A list of tuples `(field_name, expected_type)` that are derived from the `RequiredData` struct's fields.
+    /// For example
+    /// ```ignore
+    /// struct RequiredData {
+    ///   a: PropView<prop_type::String>,
+    ///   b: PropView<prop_type::Number>,
+    /// }
+    /// ```
+    /// would result in the list `&[("a", PropValueType::String), ("b", PropValueType::Number)]`.
```
dqnykamp, line 18: I believe it should be `&[("a", Some(PropValueType::String)), ("b", Some(PropValueType::Number))]`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query_type_check.rs
```diff
@@ -0,0 +1,47 @@
...
+use super::{DataQuery, PropSpecifier, PropValueType};
+
+impl DataQuery {
+    /// Guess the return type of a data query. Since this return type may depend on
+    /// retrieving a prop from a component, you must pass in a reference to the component.
+    ///
+    /// Note: this function assumes the data query is well formed (E.g., it only uses `PropSpecifier::LocalIdx` in conjunction
+    /// with `PropSource::Me`). If the data query is not well formed, this function may return an incorrect type.
+    #[allow(clippy::result_unit_err)]
+    pub fn guess_return_type<T: ComponentVariantPropTypes + 'static>(
```
dqnykamp, line 15: If I understand this correctly, it used only in tests. If so, then include text to mention that, and maybe start with an `_` like the other functions.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/data_query/data_query_type_check.rs
```diff
@@ -0,0 +1,47 @@
...
+        &self,
+    ) -> Result<Vec<PropValueType>, ()> {
```
dqnykamp, line 17: For the queries that include matching `PropProfile`s, the result is a vector of props, each of which could be one of the `PropType`s from one of the `PropProfile`s. It is not a list of props which match the `PropProfile`s in sequence.

siefkenj: I don't understand this comment...`Vec<PropValueType>` is meant to store the union type of several `PropValueType` variants.

dqnykamp: Nice. That's good to know. I thought the "return type of a data query" was a vector saying the types of each component of the returned vector; I didn't realize it was a union. I haven't gotten to the implementation part of the testing feature yet, as you can tell.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/traits/test_data_query_types.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Test that the types of the data queries match the types declared in the `RequiredData` struct.
+    /// This function should only be used as part of tests.
+    ///
+    /// It must be called with a type parameter of the related component. For example, if `RequiredData`
+    /// is for a prop on the `Text` component, this function is called
+    /// ```ignore
+    /// RequiredData::_test_data_query_types::<Text>();
+    /// ```
+    fn _test_data_query_types<T: ComponentVariantPropTypes + 'static>() -> anyhow::Result<()> {
```
dqnykamp, line 31: I'm wondering if there is a way to generalize this test to general props. I looked through all the current general props, and none of them have a `Prop` data query with `PropSpecifier::LocalIdx`. which seems to be the only reason one needs `ComponentVariantPropType` (other than for the error message, which obviously couldn't use a component name to identify the source of the error if it were a general prop).

Would it be hard to make it that if a `RequiredData` isn't given the `#[owning_component()]` attribute, then the test fails if one uses a `Prop` data query with `PropSpecifier::LocalIdx`, but otherwise works as normal (other than without the component name)? It'd be nice to have these checks when making general props.

packages/doenetml-worker-rust/lib-rust-macros/src/test_data_query_types/test_data_query_types.rs
```diff
@@ -0,0 +1,106 @@
...
+                #(#type_entry),*
+            ];
+
+            const _STRUCT_NAME: &'static str = #struct_name_str;
+        }
+    };
+
+    let test = quote! {
+        #[cfg(test)]
+        mod __types_test_auto_generated {
```
dqnykamp, line 75: If you didn't put the test in this separate module, what would be different?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/traits/test_data_query_types.rs
siefkenj: One could add a non-generic `_test_data_query_types` and write the tests manually for the general props.

packages/doenetml-worker-rust/lib-rust-macros/src/test_data_query_types/test_data_query_types.rs
siefkenj: Shouldn't be a difference, but `#[cfg(test)]` will ensure the entire module is not compiled unless test=true. If we ever added another test, we can put it in the same module. Otherwise we would need to add multiple `#[cfg(test)]` lines.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/traits/test_data_query_types.rs
dqnykamp: Sure, I could do that. But I can't think of a reason why I'd want anything customized. I want the same behavior as these, other than no support for `PropSpecifier::LocalIdx`. Or am I missing something?

PR #175, opened 2024-05-27
Parser turns html entities to strings

dqnykamp:
With this PR, the doenetml parser converts html entities and the like into strings and merges consecutive strings.

With this change, cypress tests are finally passing again.

### DIFF
--- a/packages/parser/src/parser.ts
+++ b/packages/parser/src/parser.ts
@@ -209,7 +209,12 @@ export function parseAndCompile(inText: string) {
 
             // Corresponds to the entity non-terminal in the grammar
             while (cursor.nextSibling()) {
-                if (cursor.name === "Text") {
+                if (
+                    cursor.name === "Text" ||
+                    cursor.name === "Ampersand" ||
+                    cursor.name === "EntityReference" ||
+                    cursor.name === "CharacterReference"
+                ) {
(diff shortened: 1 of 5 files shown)


packages/parser/src/parser.ts
```diff
@@ -209,7 +209,12 @@ export function parseAndCompile(inText: string) {
 
             // Corresponds to the entity non-terminal in the grammar
             while (cursor.nextSibling()) {
-                if (cursor.name === "Text") {
+                if (
+                    cursor.name === "Text" ||
+                    cursor.name === "Ampersand" ||
+                    cursor.name === "EntityReference" ||
+                    cursor.name === "CharacterReference"
+                ) {
```
siefkenj, line 217: `parser.ts` is only used in legacy code, right? The new DAST code doesn't touch it?

siefkenj:
Looks good to me! It'll be good to have all the tests working :-)

packages/parser/src/parser.ts
dqnykamp: There are other exports from `parser.ts` used in the new code, but they don't use the `parseAndCompile` export, which is the one changed in this PR. (If the had used `parseAndCompile`, they would have been losing any entities, etc.

PR #176, opened 2024-05-28
constrainToInterior component

dqnykamp:


dqnykamp:
I used a winding number definition of interior of a polygon, as it accurately calculates the interior of a non-simple polygon. However, it appears that jsxgraph is using a ray-casting algorithm for filling polygons, so the filled region is not consistent with the region in which the point is constrained! (See video). I hope we can live with this inconsistency, as I'd rather not add an attribute for two different types of constraining to the interior.

https://github.com/Doenet/DoenetML/assets/1037056/6125b9d6-d1cf-4f7f-b3ce-f10b2cb144b6

PR #182, opened 2024-06-06
Add toMarkdown function to LSP

siefkenj:
This PR adds a `doenetToMarkdown` command to the language server and adds a hook to call the command from the vscode extension. It does a reasonable job of guessing how the code should be converted, but I have not gone through all doenet tags. Unknown tags are omitted and only their text content is rendered.


packages/lsp-tools/test/doenet-to-markdown.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        );
+        source = `<ul><li>Item 1</li><li>Item 2</li></ul>`;
+        expect(doenetToMarkdown(source)).toEqual(
+            `* Item 1
+
+* Item 2
+`,
+        );
+    });
+    it("Can function macros", () => {
```
dqnykamp, line 87: can function macros -> can format function macros

packages/lsp-tools/test/doenet-to-markdown.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        let source: string;
+
+        source = `$$f(x)`;
+        expect(doenetToMarkdown(source)).toEqual(
+            `\`$$f(x)\`
+`,
+        );
+    });
```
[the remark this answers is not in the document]
dqnykamp: If the macro is within a paragraph, like `<p>We have $x in here</p>`, then it properly is all in one line. I recommend a test for both cases of the macro in a paragraph/component and outside.

packages/lsp-tools/src/doenet-to-markdown/index.ts
```diff
@@ -0,0 +1,253 @@
...
+            return [
+                {
+                    type: "inlineCode",
+                    value: toXml(node),
+                },
+            ];
+        case "root":
+            return node.children.flatMap((n) => doenetToMdast(n, sourceObj));
+        case "element":
+            // Many elements we have specific handling for
```
dqnykamp, line 118: incomplete comment

packages/lsp-tools/test/doenet-to-markdown.test.ts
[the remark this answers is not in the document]
siefkenj: This sounds like too much work for right now. This is the same problem as the placement of paragraphs for invalid pretext. Is `<p>Hello<p>$x` actually supposed to mean `<p>Hello</p><p>$x</p>`? How much does that next paragraph absorb. I sense that any algorithm will require a lot of tweaking to get right...

packages/lsp-tools/src/doenet-to-markdown/index.ts
siefkenj: Wasn't incomplete, but I clarified.

packages/lsp-tools/test/doenet-to-markdown.test.ts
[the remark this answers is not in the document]
dqnykamp: Yeah, I was just thinking in terms of how Doenet handles `<p>a</p>b`, where b is on the next line. Doenet just sends `<p>a</p>b` to the browser, and the browsers I've tried make it look like similar to new paragraph (though not with the styling of the `<p>`).

packages/lsp-tools/src/doenet-to-markdown/index.ts
```diff
@@ -0,0 +1,254 @@
...
+                    children: [{ type: "text", value: "\n\n" }],
+                },
+            ];
+        }
+        return [node];
+    });
+
+    let text = toMarkdown({ type: "root", children: mdast });
+    // Occurrences of `&#x20;` at the start of a line are probably a mistake, so trim them.
+    text = text.replace(/^&#x20;/gm, "");
```
dqnykamp, line 61: Where are the `&#x20;` coming from? Do we care that 
```
<section>
    <title>title 1</title>
    hello 1
    <section>
        <title>title 2</title>
        hello 2
    </section>
</section>
```
leads to stray values of `&#x20;` lying around. They go away if you add `<p>`s around the text.

packages/lsp-tools/src/doenet-to-markdown/index.ts
```diff
@@ -0,0 +1,254 @@
...
+                            depth: clampToHeadingNumber(depth),
+                            children: mapChildren(
+                                node.children,
+                            ) as MdastPhrasingContent[],
+                        },
+                    ];
+                }
+                case "section":
+                    // Directly pass through. Everything will be handled by the `<title>` element
+                    return node.children.flatMap((n) =>
```
dqnykamp, line 221: Or just `return mapChildren(node.children);` to emphasize it is the same as others.

packages/vscode-extension/extension/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## v0.7.7
+
+-  Update DoenetML to v0.7alpha
```
dqnykamp, line 3: 0.7.0-alpha7, presumably

packages/lsp-tools/src/doenet-to-markdown/index.ts
siefkenj: I vote for ignoring them for now. They are there because they are the text encoding of newlines. Since we don't know where the `<p>`s start and stop, I decided not to trim whitespace from the start/end of nodes, lest everything gets shoved to one line. A consequence is sometimes a `\n` shows up in a `type: "text"` node which gets turned to `&#x20;`

dqnykamp (approved):

PR #193, opened 2024-06-11
Make a DoenetMLIframe component

siefkenj:
This PR creates a `DoenetMLIframe` component. When using this component, you specify a copy of `standalone` to use (e.g., from a CDN URL). Its contents are completely isolated inside of an iframe. This means, using `DoementMLIframe`, you can run multiple versions of DoenetML at the same time on the same page.

One drawback of this approach is that you cannot directly pass the React component the usual callbacks. Instead, you have to use the provided `onMessage(...)` prop to receive messages coming from the `<DoenetML ... />` component that is rendered in the iframe.

### DIFF
--- a/packages/doenetml-iframe/src/index.tsx
+++ b/packages/doenetml-iframe/src/index.tsx
@@ -1,0 +1,137 @@
+/*
+ * This file is for running a dev test of the codemirror component.
+ * It does not show up in the bundled package.
+ */
+
+import React from "react";
+import type { DoenetML } from "@doenet/doenetml";
+// @ts-ignore
+import iframeJsSource from "../dist/iframe/iframe-index.iife.js?raw";
+import { watchForResize } from "./resize-watcher";
+
+type DoenetMLProps = React.ComponentProps<typeof DoenetML>;
+
+/**
+ * A message that is sent from an iframe to the parent window.
+ */
+type IframeMessage = {
+    origin: string;
+    data: Record<string, unknown>;
+};
+
+export type DoenetMLIframeProps = {
+    doenetML: string;
+    doenetMLProps: DoenetMLProps;
+    /**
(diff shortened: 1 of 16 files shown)


packages/doenetml-iframe/src/index.tsx
```diff
@@ -0,0 +1,137 @@
+/*
+ * This file is for running a dev test of the codemirror component.
```
dqnykamp, line 2: ??

packages/doenetml-iframe/vite.config.ts
```diff
@@ -0,0 +1,33 @@
...
+
+// These are the dependencies that will not be bundled into the library.
+const EXTERNAL_DEPS = ["react", "react-dom"];
+
+// https://vitejs.dev/config/
+export default defineConfig({
+    base: "./",
+    plugins: [dts({ rollupTypes: true })],
+    build: {
+        minify: false,
```
dqnykamp, line 12: Is there a reason we aren't minifying these (not they are at all big)?

siefkenj: Minifying should only be done at the final stage for end-user code that is deployed. Since this is to be included as a library in other code, the calling code is responsible for minifying. There's no point in minifying twice (and I think it's kinda mean, in case someone has to debug something).

All that is to say, `standalone` (and maybe the worker) is the only package that currently comes to mind that I think should be minified.

packages/doenetml/vite.config.ts
```diff
@@ -31,7 +32,9 @@ export default defineConfig({
                 {
                     src: "package.json",
                     dest: "./",
-                    transform: transformPackageJson,
```
dqnykamp, line 34: Delete the tranformPackageJson code, below.

PR #194, opened 2024-06-11
Tests in WASM

siefkenj:
This PR adds tests that execute in a headless browser where they can spawn a real instance of the wasm core. This (with some advanced configuring) should allow testing of rust functions that call out to JS.


dqnykamp (approved):
That's great that we can now test math!

PR #198, opened 2024-06-13
Rework Virtual Keyboard

siefkenj:
This PR introduces a new virtual keyboard. It will be page-unique and create JSON messages when a key is pressed. This will allow a "virtual virtual keyboard" to forward the produced messages to recoil, and also for non-recoil methods to work.

This should allow a virtual keyboard to work outside of an iframe and send messages to a copy of doenet running inside an iframe.


packages/doenetml-worker/package.json
```diff
@@ -51,7 +51,7 @@
         "styled-components": "^5.3.11"
     },
     "dependencies": {
-        "compromise": "^13.11.4",
+        "compromise": "^14.13.0",
```
dqnykamp, line 54: This change should be reverted.

dqnykamp (approved):

PR #204, opened 2024-06-30
Move virtual keyboard outside iframe

dqnykamp:
This PR moves the virtual keyboard outside the iframe if DoenetViewer or DoenetEditor are access via `doenetml-iframe`. This allows the keyboard to placed at the bottom of the browser window regardless of the size of the iframe.

### DIFF
--- a/packages/doenetml-iframe/src/iframe-editor-index.ts
+++ b/packages/doenetml-iframe/src/iframe-editor-index.ts
@@ -21,7 +21,7 @@ document.addEventListener("DOMContentLoaded", () => {
         undefined,
         {
             ...doenetEditorProps,
-            keyboardIsOutsideIframe: true,
+            externalVirtualKeyboardProvided: true,
             // Callbacks have to be explicitly overridden here so that they
             // can message the parent React component (outside the iframe).
             doenetmlChangeCallback: (args: unknown) => {
--- a/packages/doenetml-iframe/src/iframe-editor-index.ts
+++ b/packages/doenetml-iframe/src/iframe-editor-index.ts
@@ -21,6 +21,7 @@ document.addEventListener("DOMContentLoaded", () => {
         undefined,
         {
             ...doenetEditorProps,
+            keyboardIsOutsideIframe: true,
             // Callbacks have to be explicitly overridden here so that they
             // can message the parent React component (outside the iframe).
             doenetmlChangeCallback: (args: unknown) => {
--- a/packages/doenetml-iframe/src/index.tsx
+++ b/packages/doenetml-iframe/src/index.tsx
@@ -55,11 +55,6 @@ export type DoenetViewerIframeProps = DoenetViewerProps & {
      * overwriting any doenetmlVersion or urls provided
      */
     autodetectVersion?: boolean;
-    /**
-     * Added to remove the scrollableContainer attribute from DoenetViewerProps
-     * that will be stringified (as it has circular structure)
-     */
-    scrollableContainer?: any;
 };
 
 export type DoenetEditorIframeProps = DoenetEditorProps & {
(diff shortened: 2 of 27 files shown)


dqnykamp:
Got the keyboard to work for the case when we have the iframe. The keyboard no longer works if don't have an iframe. Presumably, we need to have `doenetml-iframe` send a parameter to `doenetml` so that we can change behavior based on the context.

siefkenj:
I think something in the global doenetConfig should be set if we're in an iframe. The `RecoilVirtualKeyboard` could listen for that and behave normally if it's not in an iframe.

dqnykamp:
We don't need to know if we are in any iframe; we need to know specifically if we are in an iframe created by the `doenetml-iframe` package (or, I suppose, any other iframe that adds an `<OutsideIframeKeyboard />` outside the iframe). I added a `keyboardIsOutsideIframe` prop to `DoenetViewer` and `DoenetEditor` that is set by `doenetml-iframe` to make the keyboard look for the outside one.

siefkenj:
> We don't need to know if we are in any iframe; we need to know specifically if we are in an iframe created by the `doenetml-iframe` package (or, I suppose, any other iframe that adds an `<OutsideIframeKeyboard />` outside the iframe). I added a `keyboardIsOutsideIframe` prop to `DoenetViewer` and `DoenetEditor` that is set by `doenetml-iframe` to make the keyboard look for the outside one.

The idea would be that the global config would be set by `doenetml-iframe`, so a new prop wouldn't be needed.

dqnykamp:
I'm not sure how to make `doenetml-iframe` change a global config. Feel free to change the mechanism if that is preferred.

packages/doenetml-iframe/src/iframe-editor-index.ts
```diff
@@ -21,6 +21,7 @@ document.addEventListener("DOMContentLoaded", () => {
         undefined,
         {
             ...doenetEditorProps,
+            keyboardIsOutsideIframe: true,
```
siefkenj, line 24: This isn't really about whether the keyboard is outside the iframe, as whether the built-in keyboard should be suppressed. Maybe `externalVirtualKeyboardProvided`?

packages/doenetml-iframe/src/index.tsx
```diff
@@ -244,24 +241,27 @@ export function DoenetViewer({
...
-                width: "100%",
-                boxSizing: "border-box",
-                overflow: "hidden",
-                border: "none",
-                minHeight: 200,
-            }}
-            height={height}
-        />
+        <React.Fragment>
+            <OutsideIframeKeyboard />
```
siefkenj, line 245: Maybe rename this `ExternalVirtualKeyboard`?

packages/virtual-keyboard/src/virtual-keyboard/unique-keyboard-tray.tsx
```diff
@@ -3,13 +3,24 @@ import { createRoot, Root } from "react-dom/client";
...
-    count: 0,
-    keyboardDomNode: null as HTMLElement | null,
-    keyboardReactRoot: null as Root | null,
-    callbacks: [] as OnClick[],
+let virtualKeyboardState: {
+    count: number;
+    keyboardDomNode: HTMLElement | null;
+    keyboardReactRoot: Root | null;
+    callbacks: OnClick[];
 };
```
siefkenj, line 11: This would look better defined as a type. They you can do


`const virtualKeyboardState: VirtualKeyboardState = (window as any).virtualKeyboardState || { ... }`

packages/doenetml-iframe/src/iframe-editor-index.ts
dqnykamp: True, though we do then assume that messages come in from `window.parent`. Not sure if that exists other than with an iframe. I'm happy to rename the prop, though.

packages/doenetml-iframe/src/index.tsx
dqnykamp: done

packages/virtual-keyboard/src/virtual-keyboard/unique-keyboard-tray.tsx
dqnykamp: done, though now I set window.virtualKeyboardState every time since I don't have an if statement

PR #210, opened 2024-07-09
create a point component

dqnykamp:
This PR creates a basic `<point>` component where x and y coordinates must be specified via attributes.  A rudimentary renderer and a `move` action, allow the renderer point location and the core props to maintain correspondence.


packages/doenetml-prototype/src/renderers/components/point.tsx
```diff
@@ -3,11 +3,26 @@ import { BasicComponent } from "../types";
...
+import { PointProps } from "@doenet/doenetml-worker-rust";
 
-export const PointInGraph: BasicComponent = ({ node }) => {
+type PointData = { props: PointProps };
+
+export const PointInGraph: BasicComponent<PointData> = ({ node }) => {
     const board = React.useContext(GraphContext);
     const pointRef = React.useRef<JSG.Point | null>(null);
 
+    let x = JSON.parse(
```
siefkenj, line 15: Why not `const`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/prop_profile.rs
```diff
@@ -89,6 +92,7 @@ pub const fn prop_profile_to_type(profile: PropProfile) -> PropValueType {
         PropProfile::ListCodeNumber => PropValueType::String,
         PropProfile::Renderable => PropValueType::ComponentRef,
         PropProfile::RenderedChildren => PropValueType::AnnotatedContentRefs,
+        PropProfile::SplitSymbols => super::PropValueDiscriminants::Boolean,
```
siefkenj, line 95: Use `PropValueType` instead

packages/doenetml-worker-rust/lib-doenetml-core/src/core/state/types/math_expr.rs
```diff
@@ -68,7 +69,9 @@ impl serde::Serialize for MathExpr {
     where
         S: serde::Serializer,
     {
-        serializer.serialize_str(&self.math_object.0)
+        let mut m = serializer.serialize_struct("MathExpr", 1)?;
```
siefkenj, line 72: What are these magic numbers? e.g. `1`?

Comments, please.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/props/prop_profile.rs
dqnykamp: Oops, didn't notice the autocomplete!

packages/doenetml-prototype/src/renderers/components/jsxgraph/listeners.ts
```diff
@@ -1,56 +1,143 @@
 import * as JSG from "jsxgraph";
 
-type InteractionState = {
+export type InteractionState = {
     dragActive: boolean;
     pointerAtDown: [number, number];
     pointerIsDown: boolean;
 };
 
+// functions giving that actions to be performed within the listeners
```
siefkenj, line 9: Use `/**  */` style comments for JS docstrings

packages/doenetml-prototype/src/renderers/components/jsxgraph/listeners.ts
```diff
@@ -1,56 +1,143 @@
...
+
+    if (listenerActions.up) {
+        const upListener = function (_e: Event) {
+            const e = _e as MouseEvent;
             interactionState.dragActive = false;
-        }
-        console.log("keydown", obj.elType, { interactionState, obj });
-    });
+            interactionState.pointerIsDown = false;
+            listenerActions.up?.(e, interactionState);
```
siefkenj, line 80: Maybe we want to save `const up = listenerActions.up;` and then use `up` directly here. It would never be null and wouldn't change if `listenerActions` changed.

dqnykamp: Ah, that's better. I was annoyed that I had to use the `?`.

PR #214, opened 2024-07-12
number prop

dqnykamp:
This PR creates a new number prop that can be calculaated from math, number or string children or attirubte components.


packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/number_prop.rs
```diff
@@ -0,0 +1,262 @@
...
+                if numbers_maths_and_strings
+                    .iter()
+                    .any(|prop| matches!(&prop.value, PropValue::Math(_) | PropValue::Number(_)))
+                {
+                    // invalid combination. Haven't implemented number dependency with others
+                    PropCalcResult::Calculated(prop_type::Number::NAN)
+                } else {
+                    // Have multiple string variables. Concatenate the string values into a single string
+
+                    if numbers_maths_and_strings.iter().any(|prop| prop.changed) {
```
siefkenj, line 192: I prefer to have `if !changed { ... }` with early return to avoid some nesting.

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/number_prop.rs
```diff
@@ -0,0 +1,262 @@
...
+                        let mut value = String::new();
+
+                        for prop in numbers_maths_and_strings {
+                            match &prop.value {
+                                PropValue::String(string_value) => value += string_value,
+                                _ => {
+                                    panic!(
+                                        "Should get boolean or string for boolean, found {:?}",
```
siefkenj, line 200: `boolean`? You just just have strings here, right? And this should be `unreachable!`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/test_utils.rs
```diff
@@ -0,0 +1,181 @@
+use crate::props::{cache::PropWithMeta, prop_type, DataQueryResult, PropCalcResult, PropValue};
```
siefkenj, line 1: Can we name this file with a `.test.rs`?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/number_prop.rs
dqnykamp: It *should* be unreachable, but since the required data is `Vec<PropView<PropValue>>`, there's no place before this to catch that someone connected an invalid data query that returned another type of prop.

siefkenj: I suggest adding a blanket test at line #142

dqnykamp: done

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/test_utils.rs
dqnykamp: done

PR #217, opened 2024-07-13
Add a <number> component

dqnykamp:
This PR adds a `<number>` component, updates the number and math props, and adds additional number, math, and point tests.


packages/doenetml-prototype/src/renderers/doenet/point.tsx
```diff
@@ -12,6 +12,8 @@ import { serializedComponentsReviver } from "@doenet/utils";
 import { Action, PointProps } from "@doenet/doenetml-worker-rust";
 import { useAppDispatch } from "../../state/hooks";
 import { coreActions } from "../../state/redux-slices/core";
+//@ts-ignore
```
siefkenj, line 15: This is a bit frustrating. Maybe we could re-export some typed versions of the functions that we're using?

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/number.rs
```diff
@@ -0,0 +1,66 @@
+use crate::components::prelude::*;
+use crate::general_prop::NumberProp;
+use crate::general_prop::NumberToStringProp;
+use crate::props::UpdaterObject;
+
+/// The `<number>` component calculates a numerical value from its contents and display the result as text
```
siefkenj, line 6: Does it display fractions as well as decimals?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/general_prop/math_prop.rs
```diff
@@ -544,6 +471,115 @@ impl PropUpdater for MathProp {
...
 // }
 
+/// Calculate a math expression from multiple components,
+/// where components are math, number, or string prop values.
+///
+/// Note: we assume that `math_number_strings` has already been
+/// verified to have only math, number, and string values.
+///
+///
+/// /// Parameters:
```
siefkenj, line 481: `/// Parameters` -> `# Parameters`

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/number.rs
dqnykamp: No, it is an `f64` representation, so only decimals. This is approximately equivalent to the legacy `<number>` component, which uses a Javascript `Number`. At this point, if you want fractions, you have to stick with `<math>`.

packages/doenetml-prototype/src/renderers/doenet/point.tsx
dqnykamp: Sounds like a worthwhile goal. Seems overwhelming to me, so I could use some help getting started!

siefkenj: I wasn't suggesting doing the whole thing. Just create and re-export the function you used in this file.

dqnykamp: Do you mean within DoenetML rather than in `math-expressions`? I could try that.

dqnykamp: I can only do that for a function that returns something simple like a number. The math expression object itself has a zillion attributes, so just describing that type would be non-trivial.

dqnykamp: done for this example

PR #218, opened 2024-07-16
graph axis limit attributes

dqnykamp:
This PR adds attributes and props on `<graph>` to control and read out the limits of the axes.

### DIFF
--- a/packages/doenetml-prototype/src/renderers/doenet/graph.tsx
+++ b/packages/doenetml-prototype/src/renderers/doenet/graph.tsx
@@ -18,6 +18,7 @@ import { Element } from "../element";
 import { Action, GraphProps } from "@doenet/doenetml-worker-rust";
 import { useAppDispatch } from "../../state/hooks";
 import { coreActions } from "../../state/redux-slices/core";
+import { arrayEq } from "../../utils/array";
 
 (window as any).JSG = JSG;
 
@@ -60,7 +61,7 @@ export const Graph: BasicComponent<GraphData> = ({ node }) => {
             return;
         }
 
-        let boundingBox: BoundingBox = [
+        const boundingBox: BoundingBox = [
             node.data.props.xMin,
             node.data.props.yMax,
             node.data.props.xMax,
@@ -121,35 +122,30 @@ export const Graph: BasicComponent<GraphData> = ({ node }) => {
 
         board.on("boundingbox", () => {
             if (!settingBoundingBox.current) {
-                let newBoundingBox = board.getBoundingBox();
-                let [xMin, yMax, xMax, yMin] = newBoundingBox;
+                const newBoundingBox = board.getBoundingBox();
+                const [xMin, yMax, xMax, yMin] = newBoundingBox;
 
                 // look for a change in bounding box that isn't due to round-off error
-                let xScale = Math.abs(xMax - xMin);
-                let yScale = Math.abs(yMax - yMin);
-                let diffs = newBoundingBox.map((v, i) =>
+                const xScale = Math.abs(xMax - xMin);
+                const yScale = Math.abs(yMax - yMin);
+                const diffs = newBoundingBox.map((v, i) =>
                     Math.abs(v - (previousBoundingBox.current?.[i] ?? 0)),
                 );
+                const eps = 1e-12;
+
                 if (
-                    Math.max(
-                        diffs[0] / xScale,
-                        diffs[1] / yScale,
-                        diffs[2] / xScale,
-                        diffs[3] / yScale,
-                    ) > 1e-12
+                    diffs[0] / xScale > eps ||
+                    diffs[1] / yScale > eps ||
+                    diffs[2] / xScale > eps ||
+                    diffs[3] / yScale > eps
                 ) {
                     previousBoundingBox.current = newBoundingBox;
 
-                    let action: Action = {
+                    const action: Action = {
                         component: "graph",
                         actionName: "changeAxisLimits",
                         componentIdx: node.data.id,
-                        args: {
-                            x_min: xMin,
-                            x_max: xMax,
-                            y_min: yMin,
-                            y_max: yMax,
-                        },
+                        args: { xMin, xMax, yMin, yMax },
                     };
                     dispatch(coreActions.dispatchAction(action));
                 }
@@ -163,7 +159,7 @@ export const Graph: BasicComponent<GraphData> = ({ node }) => {
         };
     }, [boardRef]);
 
-    let boundingBox: BoundingBox = [
+    const boundingBox: BoundingBox = [
         node.data.props.xMin,
         node.data.props.yMax,
         node.data.props.xMax,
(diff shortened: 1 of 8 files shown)


packages/doenetml-prototype/src/renderers/doenet/graph.tsx
```diff
@@ -98,9 +119,70 @@ export const Graph: BasicComponent = ({ node }) => {
...
+                    ) > 1e-12
+                ) {
+                    previousBoundingBox.current = newBoundingBox;
+
+                    let action: Action = {
+                        component: "graph",
+                        actionName: "changeAxisLimits",
+                        componentIdx: node.data.id,
+                        args: {
+                            x_min: xMin,
```
dqnykamp, line 148: this mix of camel case and snake case indicates to me I didn't do something right!

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/graph.rs
```diff
@@ -5,20 +5,51 @@ use crate::props::UpdaterObject;
...
+    #[serde(expecting = "`x_min`, `x_max`, `y_min`, and `y_max` must be numbers")]
+    pub struct GraphChangeAxisLimitsActionArgs {
+        pub x_min: prop_type::Number,
+        pub x_max: prop_type::Number,
+        pub y_min: prop_type::Number,
+        pub y_max: prop_type::Number,
+    }
+
+    enum Actions {
+        ChangeAxisLimits(ActionBody<GraphChangeAxisLimitsActionArgs>),
```
siefkenj, line 52: Do you consistently call these `AxisLimits`? To me they are more `ViewportBounds`

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/graph.rs
```diff
@@ -5,20 +5,51 @@ use crate::props::UpdaterObject;
...
+        #[attribute(prop = NumberProp, default = -10.0)]
+        YMin,
+        #[attribute(prop = NumberProp, default = 10.0)]
+        YMax,
+    }
+
+    #[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
+    #[cfg_attr(feature = "web", derive(tsify_next::Tsify))]
+    #[cfg_attr(feature = "web", tsify(from_wasm_abi))]
+    #[serde(expecting = "`x_min`, `x_max`, `y_min`, and `y_max` must be numbers")]
```
siefkenj, line 43: You also want to add a serde rename (and tsify rename?) to make these camelCase

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
```diff
@@ -98,9 +119,70 @@ export const Graph: BasicComponent = ({ node }) => {
...
+                let newBoundingBox = board.getBoundingBox();
+                let [xMin, yMax, xMax, yMin] = newBoundingBox;
+
+                // look for a change in bounding box that isn't due to round-off error
+                let xScale = Math.abs(xMax - xMin);
+                let yScale = Math.abs(yMax - yMin);
+                let diffs = newBoundingBox.map((v, i) =>
+                    Math.abs(v - (previousBoundingBox.current?.[i] ?? 0)),
+                );
+                if (
```
siefkenj, line 133: Explain what this test is for.

I also prefer early return rather than nesting logic in the if statement.

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
siefkenj: See my comment on the Rust code.

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
```diff
@@ -98,9 +119,70 @@ export const Graph: BasicComponent = ({ node }) => {
...
+        });
+
         setBoard(board);
+
+        return () => {
+            board.off("boundingbox");
+        };
     }, [boardRef]);
 
+    let boundingBox: BoundingBox = [
```
siefkenj, line 166: `const`

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
```diff
@@ -98,9 +119,70 @@ export const Graph: BasicComponent = ({ node }) => {
...
+        node.data.props.yMax,
+        node.data.props.xMax,
+        node.data.props.yMin,
+    ];
+
+    if (
+        board &&
+        boundingBox.some((v, i) => v !== previousBoundingBox.current?.[i])
+    ) {
+        settingBoundingBox.current = true;
```
siefkenj, line 177: Why do you need this ref? This code is running synchronously.

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
```diff
@@ -98,9 +119,70 @@ export const Graph: BasicComponent = ({ node }) => {
...
+    let boundingBox: BoundingBox = [
+        node.data.props.xMin,
+        node.data.props.yMax,
+        node.data.props.xMax,
+        node.data.props.yMin,
+    ];
+
+    if (
+        board &&
+        boundingBox.some((v, i) => v !== previousBoundingBox.current?.[i])
```
siefkenj, line 175: I'd just make a util `function arrayEq` and use that so the purpose is clearer. (I'm sure one of the libraries we have already includes such a thing, but it's such a small function it couldn't hurt)

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/graph.rs
dqnykamp: Or, it could be `bounding box` in jsxgraph speak. I'm just thinking of it more in terms of the mathematical object than the widget. But, given that this is not visible by authors, I don't feel strongly about the name.

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
dqnykamp: The explanation is above on line 127. Is more needed?

dqnykamp: I'm not entirely sure what you mean by early return. But, I changed the logic to be short-circuiting or's.

packages/doenetml-prototype/src/renderers/doenet/graph.tsx
dqnykamp: This code has been through quite a bit of hardening, but unfortunately with no automated testing (as I couldn't figure out how to get cypress to interact with jsxgraph). I'm loath to change it because I remember having some case that was fixed with this check. Is it possible that jsxgraph does something asynchronously? I remember that making the check that `settingBoundingBox.current` is `false` in `board.on("boundingbox")` stopped some kind of bad behavior.

dqnykamp: OK. Well, I removed it, and I can't find any adverse effects. I'll just imagine that I changed multiple variables at once when adding in this check however many years ago, and any correlation with behavior was coincidental.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/graph.rs
dqnykamp: OK

siefkenj:
Looks good!

PR #220, opened 2024-07-18
ability to have separate rendered props in a graph

dqnykamp:
This PR extends the `for_render` attribute on a prop so one can separately specify if the prop is `for_render` when the component is in a graph versus in text.

The PR also adds a separate point renderer and rendered props when the point is not in a graph.


dqnykamp:
The PR builds on #218, so the first three commits are not part of the commit.

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/_fragment.rs
```diff
@@ -6,6 +6,7 @@ use crate::props::UpdaterObject;
 /// For example, this component is used to render `$s.title` in `<section name="s" />$s.title`.
 #[component(name = _Fragment)]
 mod component {
+
```
siefkenj, line 9: Random blank line?

packages/doenetml-worker-rust/lib-doenetml-core/src/components/doenet/point.rs
```diff
@@ -94,3 +109,91 @@ impl ComponentOnAction for Point {
...
+
+    pub use coords::*;
+
+    mod coords {
+
+        use crate::state::types::math_expr::MathExpr;
+
+        use super::*;
+
+        /// Information about how to reference this component from an `xref`
```
siefkenj, line 124: Incorrect comment?

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/document_model.rs
```diff
@@ -174,9 +175,26 @@ impl DocumentModel {
     ) -> impl Iterator<Item = PropPointer> {
         let local_prop_indices = {
             let document_structure = self.document_structure.borrow();
+
+            // Determine whether or not component is in a graph (i.e., has a graph ancestor).
+            // If so, we'll look for props that are rendered `in_graph` rather than `in_text`.
+            let in_graph = document_structure
+                .get_true_component_ancestors(component_idx)
```
siefkenj, line 182: I don't recall what this function exactly does, but does the `_true` mean it ignores parents coming from `extend`? Please verify that this is actually the call we want.

dqnykamp: That's right. It gets the parent from the DAST and skips extra parents such as the component extended.

packages/doenetml-worker-rust/lib-rust-macros/src/component_module/items/props.rs
```diff
@@ -169,12 +199,31 @@ impl PropsEnum {
...
             }
         }
-        match variant.for_render {
-            true => descriptions.push(
-                "- ForRender: this prop is always rendered and available to the UI.".to_string(),
-            ),
-            false => descriptions.push("- NotForRender: this prop is not rendered.".to_string()),
+
+        #[allow(clippy::collapsible_else_if)]
+        if variant.for_render().in_graph {
```
siefkenj, line 204: Since this is exhaustive, it could be done with a `match (variant.for_render().in_graph, variant.for_render().in_text) { ... }`

packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_model/document_model.rs
siefkenj: But is that what we want? If we have a point in a graph and we extend that point outside of a graph, we want the one outside the graph to use the text renderer.

dqnykamp: That's right. That's what this should do, given that the "true" ancestors of the extended point will not include a graph. (In the dast, the extended point doesn't descend from a graph, though it does descend from a graph via its extended link from its source.)

I didn't rely on my understanding; I tested examples. I now made an automated test demonstrating this: `rendered_props_with_references()` inside the point tests. Let me know if there is another case I should consider.

siefkenj: I understand now. We do want the true parent and not the inherited one! Tests give me confidence :-)

siefkenj:
Looks good to go :-)

PR #228, opened 2024-08-01
Nextra codeblocks for displaying doenet code and rendered results

dqnykamp:
This PR adds the codeblocks for `doenet-viewer`, `doenet-editor`, and `doenet-example` to the Doenet documentation.

The codeblocks are short hand for calling `<DoenetViewer>`, `<DoenetEditor>`, and `<DoenetExample>`.


packages/docs-nextra/components/doenet.tsx
```diff
@@ -1,6 +1,27 @@
...
+}
+
+export function DoenetEditor({
+    source,
+    showFormatter = false,
+}: React.PropsWithChildren<{ source: string; showFormatter?: boolean }>) {
+    console.log("source for editor", JSON.stringify(source));
+    return (
+        <div className="doenet-editor-container">
+            <DoenetEditorOrig doenetML={source} showFormatter={showFormatter} />
```
siefkenj, line 24: I think the default view should be editor on top and rendered below, rather than side-by-side

packages/docs-nextra/pages/tutorials/quickStart/quickStart-1.mdx
```diff
@@ -15,8 +15,7 @@ This tutorial covers a sample project in 5 parts; each part is about 5-10 minute
 
 # Part 1: Creating a Doenet Account & Starting an Activity
 
-<DoenetML>
-{`
+<DoenetViewer source={`
```
siefkenj, line 18: Is there a reason this isn't a triple blacktic?

packages/docs-nextra/scripts/wrap-doenet-editor.ts
```diff
@@ -0,0 +1,47 @@
...
+                    return;
+                }
+                if (node.lang !== "doenet-editor") {
+                    return;
+                }
+
+                console.log(
+                    "value for wrapped editor",
+                    JSON.stringify(node.value),
+                );
```
siefkenj, line 28: This is going to make the logs look really ugly as hundreds of examples get integrated into the docs.

packages/docs-nextra/scripts/wrap-doenet-example.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Wraps codeblocks of type `doenet-example` with a JSX element that renders the code and the output.
+ */
+export const wrapDoenetExample: Plugin<void[], MdastRoot, MdastRoot> =
+    function () {
+        return (tree) => {
+            visit(tree, (node) => {
+                type MdxJsxFlowElement = typeof node & {
+                    type: "mdxJsxFlowElement";
+                };
```
siefkenj, line 17: This type is used in several files. I think we can put it in a `types.ts` and import it (that file can also import `"mdast-util-mdx-jsx"`

packages/doenetml-iframe/src/index.tsx
```diff
@@ -259,7 +259,7 @@ export function DoenetViewer({
                     boxSizing: "border-box",
                     overflow: "hidden",
                     border: "none",
-                    minHeight: 200,
+                    minHeight: 50,
```
siefkenj, line 262: Ah, I think the min-height was to accommodate the virtual keyboard at some point.

packages/docs-nextra/scripts/wrap-doenet-editor.ts
dqnykamp: Yes, this is only there until I figure out why the newlines are disappearing. They are still here with this `console.log`, but then they are gone in the `console.log` on line 21 of `packages/docs-nextra/components/doenet.tsx`. It's mysterious and basically makes triple backtick doenet-editor non-functional. I'm not sure where to look for this removal of whitespace and am hoping you have ideas!

packages/docs-nextra/components/doenet.tsx
dqnykamp: I'll check with Virginia, as I thought she wanted side-by-side. At least side-by-side does switch to top-to-bottom on mobile devices/narrow screens. We can make different versions or parametrized versions to allow the author to specify.

packages/docs-nextra/pages/tutorials/quickStart/quickStart-1.mdx
dqnykamp: No, just being lazy as I think Virginia is going to delete this.

packages/docs-nextra/scripts/wrap-doenet-example.ts
dqnykamp: I reused the original type name `MdxJsxFlowElement` like the previous code did. Not sure if that is a good idea.

packages/docs-nextra/components/doenet.tsx
dqnykamp: I ended up making a `doenet-editor` that is vertical and a `doenet-editor-horiz` that is side-by-side.

PR #232, opened 2024-08-02
Make open keyboard button more visible

dqnykamp:
This PR makes the open keyboard button be white with a black border to make it more visible, especially when it is in front of a dark background.

### DIFF
--- a/packages/virtual-keyboard/src/virtual-keyboard/keyboard-tray.css
+++ b/packages/virtual-keyboard/src/virtual-keyboard/keyboard-tray.css
@@ -23,7 +23,6 @@ Tailwind base is purposely not included because we want all styles to be prefixe
      * Some styles from tailwind preflight need to be included here
      */
     box-sizing: border-box;
-    border-stylke: solid;
     --tw-border-spacing-x: 0;
     --tw-border-spacing-y: 0;
     --tw-translate-x: 0;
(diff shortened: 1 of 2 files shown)


packages/virtual-keyboard/src/virtual-keyboard/keyboard-tray.css
```diff
@@ -64,6 +63,8 @@ Tailwind base is purposely not included because we want all styles to be prefixe
         @apply border-b-black border-b;
         @apply hover:bg-slate-300 active:bg-slate-400;
         @apply w-12 h-[1.5rem] flex items-center justify-center;
+        background-color: white;
```
siefkenj, line 66: This is fine. You can also do `bg-white` inside of an `@apply` block to do it tailwind style.

dqnykamp: I was wondering about the `@apply`, but figured you'd let me know!

dqnykamp: Is there a `border-width` `@apply` version I should use as well?

siefkenj: https://tailwindcss.com/docs/border-style

siefkenj:
Looks good.

PR #307, opened 2025-04-05
rewrite subset-of-reals in TypeScript

dqnykamp:
This PR converts the code for subset-of-reals, reorganizing to follow conventions that TypeScript recognizes.


siefkenj:
This looks like it will work. You might be able to clean it up a little bit with namespaces. It looks like sets are never constructed directly now, so you could have

```
const Constructors = {
   class RealLine implements ... { ... }
}
```
and then call `new Constructors.RealLine` instead of suffixing everything with `Class`.

It's also a bit weird to have types and functions with the same name defined but with contradicting information. You can namespace all the types with `namespace Interfaces { type ...; }` and then reference them as `Interface.RealLine`, etc.

The last thing is that it is traditional to use typescript's `interface` instead of `type` when defining the type of a class, though this is probably not a big deal.

dqnykamp:
> The last thing is that it is traditional to use typescript's `interface` instead of `type` when defining the type of a class, though this is probably not a big deal.

Is there a way with interfaces to share the common `SubsetMethods` without repeatedly writing the function signatures? When creating this approach, I was reading about the difference between interfaces and types and learned that interfaces don't support unions.

siefkenj:
You can do `extends` and list multiple interfaces if you want. I sometimes do `extends Exclude<"methodToExclude", OrigInterface>` and then override `methodToExclude` or leave it off.

packages/utils/src/math/subset-of-reals.ts
```diff
@@ -0,0 +1,895 @@
...
+            };
+        }
+
+        copy() {
+            return subsetReviver(null, this.toJSON());
+        }
+    },
+};
+
+export const Constructors = {
```
siefkenj, line 628: Not quite what I meant. I meant for what you currently call `Classes` to be called `Constructors` and the functions here to be exported directly.

dqnykamp: OK. I'll go with that. Makes the exports easier to use.

siefkenj:
Looks great!

PR #308, opened 2025-04-07
codemirror styling

dqnykamp:
This PR updates the codemirror styling to make sure that selections are visible over the active line and read-only styles.

It also upgrades a number of dependencies.


packages/test-viewer/src/test/reverse.doenet
```diff
@@ -0,0 +1,1128 @@
+<text name="n000" >0</text>
```
siefkenj, line 1: Take away

packages/test-viewer/vite.config.ts
```diff
@@ -21,17 +21,14 @@ export default defineConfig({
...
-                        "../fonts/*",
-                    ),
-                    dest: "fonts/",
-                },
-                {
-                    src: "./media/*",
-                    dest: "media/",
-                },
+                // Note: we need to copy fonts only with @doenetml, not @doenetml-prototype
+                // {
```
siefkenj, line 25: Delete instead of comment?

PR #310, opened 2025-04-08
autocomplete fixes

dqnykamp:
This PR fixes the autocompletion of tags when the cursor is at the end of the document.

It employs a temporary workaround were an extra space is appended to the source before being fed to the autocompleter.


packages/lsp-tools/src/auto-completer/index.ts
```diff
@@ -40,7 +40,7 @@ export class AutoCompleter {
         schema: ElementSchema[] = doenetSchema.elements,
     ) {
         if (source != null) {
-            this.sourceObj.setSource(source);
+            this.sourceObj.setSource(source + " ");
```
siefkenj, line 43: Add a comment about why we are doing this.

PR #311, opened 2025-04-08
Run javascript core through the rust core worker

dqnykamp:
This PR refactors `doenetml-worker-javascript` so that it is loaded through `CoreWorker.ts` of `doenetml-worker`. The postMessage interface with the javascript core was replaced with Comlink proxies to match the interface for the rust worker.

This PR is a step toward integrating rust functions with the Javascript core.


packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
+import Core from "./Core";
+import { removeFunctionsMathExpressionClass } from "./utils/math";
+import { createComponentInfoObjects } from "./utils/componentInfoObjects";
+import {
+    addDocumentIfItsMissing,
+    expandDoenetMLsToFullSerializedComponents,
+} from "./utils/expandDoenetML";
+import { returnAllPossibleVariants } from "./utils/returnAllPossibleVariants";
+// import { dastToSerialized } from "./utils/dastToSerializedComponents";
```
siefkenj, line 9: delete comment

packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
...
+
+export class PublicDoenetMLCore {
```
siefkenj, line 11: Docstring here as well as for each item talking about what they are for. The typescript types look unhelpful. Maybe a TODO indicating that they should be made correct? Or is this all going to be deleted soon?

packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
...
+                expandResult.fullSerializedComponents[0],
+            )[0],
+            allDoenetMLs: expandResult.allDoenetMLs,
+            preliminaryErrors: expandResult.errors,
+            preliminaryWarnings: expandResult.warnings,
+            componentInfoObjects,
+        };
+
+        // const serializedDocument2 = dastToSerialized(normalizedDast);
+        // console.log({ serializedDocument2 });
```
siefkenj, line 87: Delete

packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
...
+            );
+
+        return {
+            success: true as const,
+            allPossibleVariants,
+            baseComponentCounts,
+        };
+    }
+
+    async createCoreGenerateDast(
```
siefkenj, line 115: Docstring

packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
...
+        reportScoreAndStateCallback: (data: unknown) => void,
+        requestAnimationFrame: (args: {
+            action: { actionName: string; componentName?: string };
+            actionArgs: Record<string, any>;
+            delay?: number;
+            animationId: string;
+        }) => void,
+        cancelAnimationFrame: (animationId: string) => void,
+        copyToClipboard: (args: { text: string; actionId?: string }) => void,
+    ) {
```
siefkenj, line 140: Is the return type properly inferred? It could be nice to explicitly specify it.

packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    async terminate() {
+        if (this.core) {
+            await this.core.terminate();
+            this.core = null;
+        }
+    }
+}
+
+// globalThis.onmessage = function (e) {
```
siefkenj, line 306: Delete

packages/doenetml-worker-javascript/src/components/abstract/BaseComponent.js
```diff
@@ -1684,11 +1684,12 @@ export default class BaseComponent {
         return { success: true, desiredVariant };
     }
 
+    // TODO: not functioning properly, but also not called anywhere
     async copyDoenetMLToClipboard({ actionId }) {
```
siefkenj, line 1688: Why are we keeping the function if its broken and not used?

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -129,6 +135,26 @@ const pluginExpandAliasedElements: Plugin<[], DastRoot, DastRoot> = () => {
     };
 };
 
+const pluginAddCompatibilityNames: Plugin<[], DastRoot, DastRoot> = () => {
```
siefkenj, line 138: Docstring

packages/doenetml/src/Viewer/DocViewer.tsx
```diff
@@ -309,171 +314,38 @@ export function DocViewer({
...
+        //         });
+        //     } else if (e.data.messageType === "sendEvent") {
+        //         window.postMessage({
+        //             ...e.data,
+        //             subject: "SPLICE.sendEvent",
+        //             activityId,
+        //             docId,
+        //         });
+        //     }
+        // };
```
siefkenj, line 343: Why is this still here?

packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -93,13 +108,133 @@ export class CoreWorker {
...
+            this.javascriptCore.setFlags(args.flags);
+        }
 
         this.doenetCore.set_flags(JSON.stringify(args.flags));
         this.flags_set = true;
 
         resolve();
     }
 
+    async returnNormalizedRoot(): Promise<NormalizedRoot> {
```
siefkenj, line 124: docstring

packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -93,13 +108,133 @@ export class CoreWorker {
...
+
+    async generateJavascriptDast(
+        args: {
+            coreId: string;
+            userId?: string;
+            cid: string | null;
+            theme?: "dark" | "light";
+            requestedVariant?: Record<string, any>;
+            stateVariableChanges?: string;
+            initializeCounters: Record<string, number>;
```
siefkenj, line 191: These types are used multiple places? Should they be consolidated?

packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -122,6 +257,33 @@ export class CoreWorker {
         }
     }
 
+    async dispatchActionJavascript(actionArgs: {
```
siefkenj, line 260: docstring

packages/doenetml-worker-javascript/vite.config.ts
```diff
@@ -1,32 +1,20 @@
...
-        rollupOptions: {
-            // @ts-ignore
-            plugins: [nodePolyfills()],
-            external: ["react", "react-dom", "styled-components"],
-            output: {
-                globals: {
-                    react: "react",
-                    "react-dom": "react-dom",
-                    "styled-components": "styled-components",
-                },
```
siefkenj, line 28: It's great we were able to get rid of these! Is this because Core and the UICore are separate now?

siefkenj: Changing `minify` to `false` is causing a pretty big increase in build size, no?

packages/doenetml-worker-javascript/src/utils/dastToSerializedComponents.ts
```diff
@@ -0,0 +1,54 @@
...
+                if (typeof child === "string") {
+                    return child;
+                } else {
+                    return dastToSerialized(normalizedDast, child);
+                }
+            }),
+            extending: dastNode.extending,
+        };
+        return component;
+    } else {
```
siefkenj, line 42: You already have an early return, so no need for `else`

packages/doenetml-worker-javascript/package.json
```diff
@@ -12,12 +12,11 @@
...
-            "import": "./dist/CoreWorker.js",
-            "require": "./dist/CoreWorker.js"
+            "import": "./dist/index.js"
         },
-        "./*": {
-            "import": "./dist/*",
-            "require": "./dist/*"
+        "./index.js": {
+            "import": "./dist/index.js",
+            "require": "./dist/index.js"
```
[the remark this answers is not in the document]
dqnykamp: I changed this when I was pulling out my hair getting the last PR to work (which turned out to due to that gitignored package.json)! I just straight out copied a working package (utils) trying to determine why it still didn't work when I had everything matching.

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -12995,11 +12902,11 @@ export default class Core {
             this.saveChangesToDatabase();
         }, 60000);
 
-        postMessage({
-            messageType: "reportScoreAndState",
+        this.reportScoreAndStateCallback({
             state: { ...this.docStateToBeSavedToDatabase },
             score: await this.document.stateValues.creditAchieved,
         });
+
```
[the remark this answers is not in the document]
dqnykamp: Ya think? My hope is the whole file become obsolete.

packages/doenetml-worker-javascript/src/components/abstract/BaseComponent.js
dqnykamp: Oops. I fixed it and forgot to remove the comment.

I also forgot that it was an "action" so can be called from DoenetML despite the fact that it isn't called anywhere in the codebase itself. Will need to document this.

packages/doenetml-worker-javascript/vite.config.ts
dqnykamp: Core should never have had those dependencies. It doesn't do anything with react or the DOM. I think we just missed that when breaking up the packages before.

I switched minify to `false` as the build artifact of this package is no longer used directly. Instead, we now only use the build of `doenetml-worker` which includes `doenetml-worker-javascript`. If I remember right, we don't minify for packages that are intended to be imported by others. I assume it doesn't affect final build size as it will get minified by `doenetml-worker`. (Oops, it looks like `doenetml-worker` didn't have minify to `true`. Turning minify on `doenetml-worker` shrunk `index.js` from 16MB down to 12 MB.)

packages/doenetml/src/Viewer/DocViewer.tsx
dqnykamp: I still need to add back this functionality (or for the "navigate" functions, decide if that is still part of the feature set).

packages/doenetml-worker-javascript/src/CoreWorker.ts
dqnykamp: No, this function was not originally designed to return anything, so it isn't clean. I have now made it return enough to preserve the current functionality with the Comlink framework. But, it is one of the first pieces we want to replace as we connect it to the new renderers.

I'll at least add documentation to that effect.

packages/doenetml-worker-javascript/src/CoreWorker.ts
dqnykamp: Made some docstrings as well as a TODO. This part is very much in flux.

PR #315, opened 2025-04-10
Convert `point.jsx` and doenetml elements A-B to typescript

siefkenj:
I am breaking this into pieces to make it easier.

This is the start of moving the renders to Typescript so we can start improving them.

The fancy logic is in the dynamic imports of `DocViewer.tsx` where I now whitelist already converted files to add a `.tsx` extension to them.


packages/doenetml/src/Viewer/renderers/angle.tsx
```diff
@@ -205,23 +207,16 @@ export default React.memo(function Angle(props) {
             board.updateRenderer();
         }
 
-        return (
-            <>
-                <a name={id} />
-            </>
-        );
+        return null;
```
dqnykamp, line 210: My thought was that, for objects that are in graphs, one might still want to `<ref>` them (or maybe `<xref>` then if we convert to PreTeXt's tag). In that case, we'd still want something in the DOM at that location so the browser can scroll to where the graph is. If we are removing the `<a>` anchors, should we put in a `<span id={id} />` so the browser can locate it? Same for point/button/booleanInput/math and other things that go in graph. (I don't feel too strongly about it.)

packages/doenetml/src/Viewer/renderers/booleanInput.tsx
```diff
@@ -58,20 +63,20 @@ export default React.memo(function BooleanInput(props) {
...
+    let inputJXG = useRef<JXGObject | null>(null);
+    let anchorPointJXG = useRef<JXGObject | null>(null);
+    let anchorRel = useRef<[string, string] | null>(null);
 
     const board = useContext(BoardContext);
 
-    let pointerAtDown = useRef(false);
-    let pointAtDown = useRef(false);
+    let pointerAtDown = useRef<[number, number] | null>(null);
+    let pointAtDown = useRef<[number, number] | null>(null);
```
dqnykamp, line 73: `scrCoords` (and hence pointAtDown) is actually a `[number, number, number]`

packages/doenetml/src/Viewer/renderers/button.tsx
```diff
@@ -112,17 +115,19 @@ export default React.memo(function ButtonComponent(props) {
 
         newButtonJXG.isDraggable = !fixLocation.current;
 
-        newButtonJXG.on("down", function (e) {
+        newButtonJXG.on("down", function (e: JXGEvent) {
             pointerAtDown.current = [e.x, e.y];
-            pointAtDown.current = [...newAnchorPointJXG.coords.scrCoords];
+            pointAtDown.current = [
+                ...(newAnchorPointJXG.coords.scrCoords as [number, number]),
```
dqnykamp, line 121: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/math.tsx
```diff
@@ -2,34 +2,37 @@ import React, { useContext, useEffect, useRef } from "react";
...
+    let mathJXG = useRef<JXGObject | null>(null);
+    let anchorPointJXG = useRef<JXGObject | null>(null);
+    let anchorRel = useRef<[string, string] | null>(null);
 
     const board = useContext(BoardContext);
 
-    let pointerAtDown = useRef(null);
-    let pointAtDown = useRef(null);
+    let pointerAtDown = useRef<[number, number] | null>(null);
+    let pointAtDown = useRef<[number, number] | null>(null);
```
dqnykamp, line 27: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/math.tsx
```diff
@@ -155,7 +158,9 @@ export default React.memo(function MathComponent(props) {
 
         newMathJXG.on("down", function (e) {
             pointerAtDown.current = [e.x, e.y];
-            pointAtDown.current = [...newAnchorPointJXG.coords.scrCoords];
+            pointAtDown.current = [
+                ...(newAnchorPointJXG.coords.scrCoords as [number, number]),
```
dqnykamp, line 162: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/math.tsx
```diff
@@ -168,7 +173,9 @@ export default React.memo(function MathComponent(props) {
         });
 
         newMathJXG.on("hit", function (e) {
-            pointAtDown.current = [...newAnchorPointJXG.coords.scrCoords];
+            pointAtDown.current = [
+                ...(newAnchorPointJXG.coords.scrCoords as [number, number]),
```
dqnykamp, line 177: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/point.tsx
```diff
@@ -11,31 +11,33 @@ import {
...
 
-    let pointJXG = useRef(null);
-    let shadowPointJXG = useRef(null);
+    let pointJXG = useRef<JXGObject | null>(null);
+    let shadowPointJXG = useRef<JXGObject | null>(null);
 
-    let pointerAtDown = useRef(null);
-    let pointAtDown = useRef(null);
+    let pointerAtDown = useRef<[number, number] | null>(null);
+    let pointAtDown = useRef<[number, number] | null>(null);
```
dqnykamp, line 31: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/point.tsx
```diff
@@ -190,15 +192,18 @@ export default React.memo(function Point(props) {
 
         let newPointJXG = board.create("point", coords, jsxPointAttributes);
 
-        newShadowPointJXG.on("down", function (e) {
+        newShadowPointJXG.on("down", function (e: { x: number; y: number }) {
             pointerAtDown.current = [e.x, e.y];
-            pointAtDown.current = [...newShadowPointJXG.coords.scrCoords];
+            pointAtDown.current = [
+                ...(newShadowPointJXG.coords.scrCoords as [number, number]),
```
dqnykamp, line 198: `[number, number, number]`

packages/doenetml/src/Viewer/renderers/booleanInput.tsx
dqnykamp: JSXGraph uses [3D homogeneous coordinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates) to represent a point in the plane.

packages/doenetml/src/Viewer/renderers/angle.tsx
siefkenj: Yes, a `span` with an id should be fine.

packages/doenetml/src/Viewer/renderers/booleanInput.tsx
```diff
@@ -275,6 +282,7 @@ export default React.memo(function BooleanInput(props) {
 
                 calculatedY.current =
                     (o[2] -
+                        // @ts-ignore
```
dqnykamp, line 285: This won't be needed once fix the types

packages/doenetml/src/Viewer/renderers/point.tsx
```diff
@@ -267,106 +274,116 @@ export default React.memo(function Point(props) {
...
+                    var o = board.origin.scrCoords;
+                    calculatedX.current =
+                        (pointAtDown.current[1] +
+                            e.x -
+                            pointerAtDown.current[0] -
+                            o[1]) /
+                        board.unitX;
+                    calculatedY.current =
+                        (o[2] -
+                            // TODO: Is this correct? What is the `[2]` doing?
```
dqnykamp, line 337: Yes, it's correct, as mentioned in previous comments about homogeneous coordinates.

siefkenj: There seems to be a lot of duplicated JSXGraph code. Would it be easy to pull out or is it all customized per component in subtle ways?

dqnykamp: For components represented by a single location, I expect it is all copy and paste. There are presumably customizations for objects, lines and more, based on multiple points.

You will find a ton of copy and paste, though, in this renderer code.

packages/doenetml/src/Viewer/renderers/booleanInput.tsx
siefkenj: Is this also true of `pointerAtDown`?

dqnykamp: No, that doesn't come from JSXGraph.

dqnykamp (approved):
It's great to get some of these renderers into TypeScript!

PR #316, opened 2025-04-11
Convert C tags to typescript

siefkenj:
Convert tags starting with `C` to Typescript. Some of the larger files were skipped with `// @ts-nocheck`. They can become proper typescript when they are rewritten.


packages/doenetml/src/Viewer/DocViewer.tsx
```diff
@@ -964,6 +964,18 @@ export function DocViewer({
                 "button",
                 "point",
```
dqnykamp, line 965: "alert" is missing from this list.

dqnykamp (approved):

PR #317, opened 2025-04-15
request solution view callback when solutionDisplayMode is "buttonRequirePermission"

dqnykamp:
This PR introduces a new `solutionDisplayMode`: "buttonRequirePermission". In this mode, a solution starts off collapsed, identical to "button" mode. However, when a user attempts to open a solution, a "SPLICE.requestSolutionView" message is sent. In order for the solution to open, a matching "SPLICE.requestSolutionView.response" must be received with `{allowView: true}`. Otherwise, the solution remains closed, and the user cannot view the solution.

This mode allows applications to verify and record the fact that a student viewed a solution before the solution will be revealed. A loss of a connection with the server will disallow solution views rather than allowing the solution to be viewed without detecting the fact.

### DIFF
--- a/packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
+++ b/packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
@@ -355,6 +355,8 @@ export class CoreWorker {
         } else {
             resolve();
         }
+
+        close(); // Terminate the worker itself
     }
 
     async returnAllStateVariables(consoleLogComponents = false) {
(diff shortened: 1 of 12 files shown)


packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -352,6 +355,8 @@ export class CoreWorker {
         } else {
             resolve();
         }
+
+        close(); // Terminate the worker itself
```
siefkenj, line 359: Better to have this comment before the `close()` rather than trailing it.

PR #318, opened 2025-04-16
remove namespaces and assignNames, resolve components from rust

dqnykamp:
This PR reformulates how naming and references of components works in DoenetML, removing namespaces and the `assignName` attribute and resolving references to components using the rust resolver. It also adds, removes, and modifies some components.

### Naming and referencing changes
Component names no longer need to be unique. (Internally, components are now indexed by an integer `componentIdx`.) With the removal of namespaces, references to `$a` now use the algorithm created in the rust core to resolve the name. Namely:
- The parent of the references is checked if it has a unique descendant of name `a`. If so, that element is the referent.
- Otherwise, recurse to the parents of the parent, resolving if a unique descendant of name `a` is found.
- If at any point, more than one descendant matches, the reference fails to find a match.
- If no descendant is found by the time we reach the document root, the reference fails to find a match.

For referencing, the attributes `copySource`, `copyProp`, `source`,`prop`, and `link` have been removed. Instead, one must use the `extend` or `copy` attributes. The `extend` attribute is for references that link to their source and the `copy` attribute is for unlinked copied (the former `link="false"` condition).

The use of attributes with references, such as `$a{b="c"}` has been removed. Instead, one must use the `extend` or `copy` attributes to add attributes.

The `<copy>` and `<extract>` tags have been removed. Instead, one must specify the resulting component type to add attributes to a reference. To allow one to add attributes to references of list, the list components such as `<pointList>` have been turned into actual components. For example, one can use syntax such as `<pointList extend="$line.points" name="point">` to add attributes to a reference point list.

All references now uniformly require a dollar sign, including in every attribute.

With the remove of namespaces, a slash is no longer valid syntax in references, including relative paths, such as `$(../a)`. Instead, the dot syntax is used to specify both descendant names and attributes. If `$a` is resolved to a component, then `$a.b` will first look for a unique descendant of `a` with the name of `b`. If found, `b` is the referent. If multiple descendants of `a` with the name of `b` exist, then the reference fails. Otherwise, a public state variable (prop) of the component `a` is looked for and becomes the referent if found.

With the removal of `assignNames`, one must use index notation or descendant names (with a `.`) to reference subparts of a component.

#### Removed components
- `<copy>`
- `<extract>`
- `<map>` (renamed to `<repeat>` and reformulated)
- `<template>`
- `<sources>`
- `<customAttribute>` (due to `<module>` reformulation)

#### New components
- `<repeat>` (a reformulation `<map>`)
- `<repeatFromSequence>`
- `<moduleAttributes>` (due to `<modulate>` reformulation`)

#### Modified components
- `<modulate>`
- `<pointList>`, `<vectorList>`, `<intervalList>`, `<lineList>` are now actual components (rather than abstract components starting with a `_`)
- `<collect>`

### DIFF
--- a/packages/doenetml-worker-javascript/src/components/PointList.js
+++ b/packages/doenetml-worker-javascript/src/components/PointList.js
@@ -478,6 +478,10 @@ export default class PointList extends CompositeComponent {
                     dependencyType: "stateVariable",
                     variableName: "childIndicesByPoint",
                 },
+                numDimensions: {
+                    dependencyType: "stateVariable",
+                    variableName: "numDimensions",
+                },
             }),
             // When this state variable is marked stale
             // it indicates we should update replacements.
--- a/packages/doenetml-worker-javascript/src/test/tagSpecific/line.test.ts
+++ b/packages/doenetml-worker-javascript/src/test/tagSpecific/line.test.ts
@@ -6108,4 +6108,29 @@ describe("Line tag tests", async () => {
         await test_items("light");
         await test_items("dark");
     });
+
+    it("line through a point that depends on a function reference", async () => {
+        // Checks to make sure that a function reference such as `$$f(x)`
+        // is considered as a references in the sugar for the point list
+        let { core, resolveComponentName } = await createTestCore({
+            doenetML: `
+
+  <function name="f">sin(x)</function>
+  <graph>
+    <line through="(0, $$f(0)) (pi/2, $$f(pi/2))" name="l" />
+  </graph>
+  <p name="p1">y-intercept: $l.yIntercept</p>
+  <p name="p2">slope: $l.slope</p>
+
+    `,
+        });
+
+        let stateVariables = await core.returnAllStateVariables(false, true);
+        expect(stateVariables[resolveComponentName("p1")].stateValues.text).eq(
+            "y-intercept: 0",
+        );
+        expect(stateVariables[resolveComponentName("p2")].stateValues.text).eq(
+            "slope: 2/π",
+        );
+    });
 });
--- a/packages/doenetml-worker-javascript/src/utils/dast/convertToCopy.ts
+++ b/packages/doenetml-worker-javascript/src/utils/dast/convertToCopy.ts
@@ -288,7 +288,9 @@ function convertEvaluate({
         ],
     };
 
-    delete evaluateComponent.extending;
+    // Note: we don't delete the `extending` attribute of the `<evaluate>` even though it isn't directly used,
+    // as it is used in sugar to determine whether or not the component is a reference
+    // delete evaluateComponent.extending;
 
     // The child of the evaluate is a list of the form `<ol><li></li><li></li></ol>""
     // The grandchildren become children of the input attribute,


packages/doenetml-worker-javascript/src/utils/dast/types.ts
```diff
@@ -0,0 +1,243 @@
...
+};
+
+export type WarningRecord = ErrorRecord & { level: number };
+
+/**
+ * Information about the definition of an attribute for a component type.
+ *
+ * Typically most fields are undefined, as many cover specialized cases.
+ */
+export type AttributeDefinition<T> = {
```
siefkenj, line 163: So nice to have these listed!

packages/doenetml-worker-javascript/src/utils/dast/types.ts
```diff
@@ -0,0 +1,243 @@
+import { Position, RefResolution, Source } from "@doenet/doenetml-worker";
+
```
siefkenj, line 2: This file is part of the bridge between DAST and the old-style serialized state?

packages/doenetml-worker-javascript/src/utils/dast/sugar.ts
```diff
@@ -0,0 +1,334 @@
...
+    const newComponents: (string | SerializedComponent)[] = [];
+
+    for (let component of serializedComponents) {
+        if (typeof component === "string") {
+            newComponents.push(component);
+            continue;
+        }
+
+        try {
+            let componentType = component.componentType;
```
siefkenj, line 59: I assume you've been in Rust mode, and most of these are meant to be `const`?

packages/doenetml-worker-javascript/src/utils/dast/sugar.ts
```diff
@@ -0,0 +1,334 @@
...
+                if (attribute.type === "primitive") {
+                    componentAttributes[attrName] = attribute.primitive;
+                }
+            }
+
+            let newParentParametersFromSugar = {};
+
+            let newComponent = component;
+
+            if (!component.skipSugar) {
```
siefkenj, line 80: It'd be nicer to short circuit if `skipSugar` and avoid the extra indentation.

packages/doenetml-worker-javascript/src/utils/dast/sugar.ts
dqnykamp: This code was copied and pasted, and lightly edited from code I had written before I knew you were supposed to use `const`.

packages/doenetml-worker-javascript/src/utils/dast/types.ts
dqnykamp: Yes! Since the code I extracted these from has not been converted to TypeScript, I'm not confident I got every field. But most of them.

packages/doenetml-worker-javascript/src/utils/dast/types.ts
dqnykamp: Yes, this file describes the serialized state, which is pretty close to the old-style serialized state, that I'm attempting to use as the target for dast than I usable in the old code without too many modifications.

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -1369,28 +1313,24 @@ export default class Core {
...
-                componentIdx = createUniqueName(
-                    serializedComponent.componentType.toLowerCase(),
-                    longNameId,
+                // throw Error(
+                //     "Found a serialized component without a componentIdx",
+                //     serializedComponent,
+                // );
+
+                // XXX: is this the right approach for adding new components?
+                // One flaw: if we add components via sugar and we need references between then,
```
siefkenj, line 1324: Grammar

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -2952,15 +2806,13 @@ export default class Core {
                 );
             }
         } else {
-            // since original names came from the targetComponent
-            // we can use them only if we created a new namespace
-            let originalNamesAreConsistent = newNamespace;
+            // XXX: can we always keep original names in the new regime?
+            let originalNamesAreConsistent = true;
```
siefkenj, line 2810: What does this flag indicate? Is there ever a time when it's false?

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -3498,9 +3257,9 @@ export default class Core {
                 component,
                 varName: "isInactiveCompositeReplacement",
             });
-            for (let childIdx in component.allChildren) {
+            for (const childIdxStr in component.allChildren) {
                 await this.changeInactiveComponentAndDescendants(
-                    this._components[childIdx],
+                    this._components[childIdxStr],
```
siefkenj, line 3262: Does this have to be done sequentially? You could `.map` to a bunch of promises and do `await Promise.all`

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -9289,19 +9037,19 @@ export default class Core {
...
                 );
             }
         }
 
         // delete component from parent's defining children
         // and record parents
         let allParents = [];
-        for (let parentIdx in parentsOfPotentiallyDeleted) {
-            let parentObj = parentsOfPotentiallyDeleted[parentIdx];
+        for (let parentIdxStr in parentsOfPotentiallyDeleted) {
```
siefkenj, line 9051: `const`

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -9289,19 +9037,19 @@ export default class Core {
             }
         }
 
-        for (let compositeIdx of replacementsDeletedFromComposites) {
-            if (!(compositeIdx in componentsToDelete)) {
+        for (let compositeIdxStr of replacementsDeletedFromComposites) {
```
siefkenj, line 9040: `const`

packages/doenetml-worker-javascript/src/utils/dast/sugar.ts
dqnykamp: except that we still recurse to children and attributes, below, within the loop, even if `skipSugar` is `true`

packages/doenetml-worker-javascript/src/Core.js
dqnykamp: It played an important role in the namespace regime. I'm pretty sure that it will go away in the new regime; hence the comment. I'll be more confident after I get through the more complicated composites like `<map>`.

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -160,6 +161,120 @@ const pluginAddCompatibilityNames: Plugin<[], DastRoot, DastRoot> = () => {
...
+/**
+ * Perform substitutions for syntactic sugar based on the component type
+ */
+const pluginComponentSugar: Plugin<[], DastRoot, DastRoot> = () => {
+    return (tree) => {
+        visit(tree, (node) => {
+            if (!isDastElement(node)) {
+                return;
+            }
+            if (node.name === "repeat") {
```
siefkenj, line 176: There will be a few more of these, right? Maybe a `switch` is more appropriate?

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -160,6 +161,120 @@ const pluginAddCompatibilityNames: Plugin<[], DastRoot, DastRoot> = () => {
...
+                        // change `else` to a `case`
+                        child.name = "case";
+                        nCaseChildren++;
+                    }
+                }
+
+                if (nCaseChildren === 0) {
+                    // If there are no `<case>` children (and no `<else>` children that were converted)
+                    // then wrap all children in a `<case>` child and move the `condition` attribute
+                    // from the `<conditionalContent>` to the `<case>`.
```
siefkenj: Good plan :-)

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -160,6 +161,120 @@ const pluginAddCompatibilityNames: Plugin<[], DastRoot, DastRoot> = () => {
...
+                    const attrChildren = node.attributes.valueName.children;
+                    if (
+                        attrChildren.length === 1 &&
+                        attrChildren[0].type === "text"
+                    ) {
+                        const valueName = attrChildren[0].value;
+
+                        setupChildren.push({
+                            type: "element",
+                            name: "_placeholder",
```
siefkenj: If we're doing stuff like this a lot, it would make sense to have constructor functions, but for now I think its fine to specify the object explicitly.

packages/doenetml-worker-javascript/src/Core.js
dqnykamp: I ended up removing the code associated with this flag, as expected.

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -160,6 +161,120 @@ const pluginAddCompatibilityNames: Plugin<[], DastRoot, DastRoot> = () => {
...
+                // If a `<repeat>` element has a `"valueName"` and/or `"indexName" attribute with a single text child,
+                // then add a `<_repeatSetup>` child to the `<repeat>` that contains children
+                // named by the values of those attributes.
+                // These children will not be rendered, but they create targets for references to `valueName` and `indexName`.
+                // Mapping those references to the correct target will be addressed when the `<repeat>` is expanded.
+                let setupChildren: DastElementContent[] = [];
+                if (node.attributes.valueName) {
+                    const attrChildren = node.attributes.valueName.children;
+                    if (
+                        attrChildren.length === 1 &&
```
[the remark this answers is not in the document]
dqnykamp: We don't ignore `valueName="i+1"` but we do ignore `valueName="$i"`. The former case, it isn't ignored here, but presumably the resulting `name="i+1"` is ignored/dealt with in the usual code dealing with names.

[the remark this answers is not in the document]
siefkenj: But doesn't `attrChildren.length===1` mean that the attribute has only a very simple structure?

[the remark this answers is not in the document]
dqnykamp: Right, either a single text node like "i+1" or a single reference like "$i". Text nodes aren't further processed unless there is an HTML entity or something similar, right? If they are, then multiple-piece results will be ignored, which is fine (maybe with eventually giving a warning).

packages/parser/src/dast-normalize/component-sugar/utils/lists.ts
```diff
@@ -0,0 +1,122 @@
+import { DastElementContent } from "../../../types";
```
dqnykamp, line 1: @siefkenj Could you review this?

packages/parser/src/dast-normalize/component-sugar/utils/lists.ts
```diff
@@ -0,0 +1,122 @@
...
+    for (let child of children) {
+        if (child.type !== "text") {
+            // If child is a reference, add it to the current group.
+            currentGroup.push(child);
+        } else {
+            let s = child.value;
+
+            let beginInd = 0;
+
+            for (let ind = 0; ind < s.length; ind++) {
```
siefkenj, line 50: I think this could be made cleaner if we split all the strings by `/\(\) /` first (we can compute the positions during the split), and then run the parens algorithm.

Something like

```
/**
 * Split a text node based on `(`, `)`, and ` `. The
 * parens and spaces are kept in the output.
 */
function splitAtParensOrSpace(node: DastElementContent & {type: "text"}) {
   // Split up the text node and make sure the positions of the returned elements are correct.
   // This code does almost all of the work:
   // Array.from("aa b (c d) e".matchAll(/([() ]|[^() ]+)/g), (m) => ({value: m[0], position: {start: m.index, end: m.index + m[0].length}}) )
}
```

Then you can do something like

```
const splitNodes = nodes.flatMap(n => (switch (node.type) { case "text": return splitAtPerensOrSpace(node); default: return node; } );
```
And then you can run the splitting algorithm. If you wanted, you our even do something like,

```
if (Math.sum(...splitNodes.map(n => n.type == "text" && n.value == "(" ? 1 : 0))
 !== Math.sum(...splitNodes.map(n => n.type == "text" && n.value == ")" ? 1 : 0))) {
   // return failure
 }
```

right up front.

dqnykamp: Is the updated algorithm better?

(I couldn't easily use the shortcut to determine failure, as it is OK if a given text node has unbalanced parentheses.)

siefkenj: I think splitting first is better. Whether you count balanced parens first or you track until the end, I have no opinion on.

packages/doenetml-worker/lib-js-wasm-binding/src/lib.rs
```diff
@@ -122,6 +128,13 @@ impl PublicDoenetMLCore {
         resolver2
     }
 
+    pub fn delete_nodes_from_resolver(resolver: Resolver, node_list: NodeList) -> Resolver {
+        let mut resolver2 = resolver;
```
siefkenj, line 132: Why is this needed? Can you not do `mut Resolver` in the function signature?

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -461,6 +498,29 @@ impl Resolver {
     }
 }
 
+/// Get the name of `element` from its `name` attribute,
+/// where the `name` attribute must have exactly one text child to be considered valid.
+fn get_element_name(element: &FlatElement) -> Option<String> {
+    // XXX: we need to enforce restrictions on valid names. Is this the right place?
```
siefkenj, line 504: I think we should allow all names when we get to the rust code. Any validation of names should happen in the `normalize_dast` phase

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -461,6 +498,29 @@ impl Resolver {
...
+        .iter()
+        .find(|attr| {
+            attr.name == "name"
+                && attr.children.len() == 1
+                && matches!(attr.children[0], UntaggedContent::Text(_))
+        })
+        .and_then(|attr| {
+            match (attr.children.len(), attr.children.first()) {
+                // A name attribute should have exactly one text child. Otherwise it is considered invalid.
+                (1, Some(UntaggedContent::Text(name))) => Some(name.clone()),
```
siefkenj, line 517: Why are you doing these tests twice? It seems that a consequence is that if multiple `name=".."` attributes are specified, and one is valid, that one will be used. But if we assume there is only one `name` attribute, you can get rid of the `&&` clauses in the preceding `.find`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -169,6 +171,56 @@ impl Resolver {
         }
     }
 
+    /// Delete `nodes` from the resolver
+    pub fn delete_nodes(&mut self, nodes: &[FlatNode]) {
+        for node in nodes.iter() {
+            if let FlatNode::Element(element) = node {
+                if let Some(name) = get_element_name(element) {
```
siefkenj, line 178: We should be able to use a filter/map directly on the `iter` to remove two layers of if statements :-)

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -169,6 +171,56 @@ impl Resolver {
...
+                    let mut parent_option = self.node_parent[node.idx()];
```
siefkenj, line 179: `parent_option` -> `parent`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -169,6 +171,56 @@ impl Resolver {
...
+
+                    while let Some(parent_idx) = parent_option {
```
siefkenj, line 181: I think this could be written better by having `parent_idx` be the mutable thing and then directly doing a while loop over `self.node_parent[parent_idx]`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -169,6 +171,56 @@ impl Resolver {
...
+                        let name_map = &mut self.name_map[parent_idx];
+
+                        let references = name_map.remove(&name);
+
+                        match references {
+                            Some(Ref::Unique(idx)) => {
+                                // if the index doesn't match the deleted node, add it back to the name map
```
siefkenj, line 188: When would we expect this to happen?

packages/doenetml-worker/lib-js-wasm-binding/src/lib.rs
dqnykamp: Because I couldn't figure out how to do that. I had tried `resolver: mut Resolver`, but it turns out it is `mut resolver: Resolver`.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
dqnykamp: OK

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
dqnykamp: Yes, it looks like we were trying to address the case we have multiple name attributes and we want to grab the valid one. But, I'm fine with just grabbing the first name attribute and using it only if it is valid.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
dqnykamp: I expanded the comment to address this question.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -8,50 +8,223 @@ use std::{collections::HashMap, iter, mem};
...
+                ResolutionAlgorithm::SearchChildren,
+                new_len - num_prev_nodes,
+            ));
+            self.name_map
+                .extend(iter::repeat_with(HashMap::new).take(new_len - num_prev_nodes));
+        }
+
+        // Add parents and stop_propagation for new nodes
+        for node in flat_fragment.nodes.iter() {
+            self.node_parent[node.idx()] = node.parent();
```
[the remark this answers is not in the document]
dqnykamp: Not for a `flat_fragment`. The indices could be in an arbitrary order and skip around.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_expand.rs
```diff
@@ -29,7 +29,7 @@ impl Expander {
     pub fn expand(flat_root: &mut FlatRoot) -> Resolver {
         let resolver = Resolver::from_flat_root(flat_root);
         Expander::expand_refs(flat_root, &resolver);
-        Expander::consume_extend_attributes(flat_root);
+        Expander::consume_extend_copy_attributes(flat_root);
```
siefkenj, line 32: I we're going with a long name, then `consume_extend_and_copy_attributes`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_expand.rs
```diff
@@ -237,26 +265,31 @@ impl Expander {
...
                 continue;
             }
 
-            // Since this reference was inside an `extend` attribute,
+            // Since this reference was inside an `extend` or `copy` attribute,
             // we must mark it as such
             // (which indicates that we shouldn't change the component type
             // from the one specified in the element)
-            element.extending = extend_referent.map(|source| source.as_attribute());
+            if extend_or_copy.name == "copy" {
```
siefkenj, line 282: Earlier in the code we were doing a case-insensitive match. I forget of the JS normalization lowercases all the attributes...but we should stay consistent I think.

dqnykamp: Apparently the case is not normalized, as alternative case attributes break if I remove the case-insensitive comparison.

packages/doenetml-worker/lib-doenetml-core/src/dast/extended_vec.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    pub fn insert(&mut self, index: isize, element: T) {
+        if index + self.index_offset < 0 {
+            panic!("Extended vector index cannot be less than index offset");
+        }
+        self.inner
+            .insert((index + self.index_offset).try_into().unwrap(), element);
```
siefkenj, line 42: I guess there is no `into()`? You can also do `as usize`. I am not sure which is better.

packages/doenetml-worker/lib-doenetml-core/src/dast/extended_vec.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    pub fn from_vector(values: Vec<T>, index_offset: isize) -> ExtendedVector<T> {
+        ExtendedVector {
+            inner: values,
+            index_offset,
+        }
+    }
+
+    pub fn len(&self) -> usize {
```
siefkenj, line 29: We need a warning docstring at least. `v[v.len() - 1]` may panic, which is not the usual behaviour of a vector.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -327,7 +328,9 @@ impl Resolver {
             if self.resolution_algorithm[current_idx] == ResolutionAlgorithm::SearchChildren {
                 // current_idx specifies the "root" of the search. We try to resolve
                 // children based on the path part, returning an error if there is an ambiguity.
-                if let Some(referent) = self.name_map[current_idx].get(&part.name) {
+                if let Some(referent) =
+                    self.name_map[current_idx.try_into().unwrap()].get(&part.name)
```
siefkenj, line 332: I think we can just do `as isize` in these places.

packages/doenetml-worker-javascript/src/Core.js
```diff
@@ -2194,6 +2194,12 @@ export default class Core {
             let childType =
                 typeof child !== "object" ? typeof child : child.componentType;
 
+            if (childType == undefined) {
```
siefkenj, line 2197: Is this what you mean? Probably `== null` or `=== undefined`

packages/doenetml-worker/lib-doenetml-core/src/dast/extended_vec.rs
```diff
@@ -0,0 +1,92 @@
+use std::{
+    ops::{Index, IndexMut},
+    slice::Iter,
+};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ExtendedVector<T> {
```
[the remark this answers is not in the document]
dqnykamp: code removed

packages/doenetml-worker/lib-doenetml-core/src/dast/extended_vec.rs
dqnykamp: code removed

PR #321, opened 2025-04-18
Resolve references in leftover path parts

dqnykamp:
This PR extends the rust resolver to resolve references in indices of path paths. For example, in a reference `$x.y[$n]`, the `$n` from the index is resolved to its referent.

### DIFF
--- a/packages/doenetml-worker/lib-doenetml-core/src/dast/flat_dast/untagged_flat_dast.test.rs
+++ b/packages/doenetml-worker/lib-doenetml-core/src/dast/flat_dast/untagged_flat_dast.test.rs
@@ -6,7 +6,7 @@ use crate::test_utils::*;
 #[test]
 fn can_flatten_dast_root() {
     let dast_root = dast_root_no_position(
-        r#"<document>hi<foo my_attr="777 $x.y.z"/>and$$f(<bar>xxx</bar>)</document>"#,
+        r#"<document>hi<foo my_attr="777 $x.y.z[$a]"/>and$$f(<bar>xxx</bar>)</document>"#,
     );
     let flat_root = FlatRoot::from_dast(&dast_root);
     // It is easier to compare JSON, so we serialize and deserialize for the comparison.
@@ -20,7 +20,7 @@ fn can_flatten_dast_root() {
                   {
                     "type": "Element",
                     "name": "document",
-                    "children": ["hi", 1, "and", 3],
+                    "children": ["hi", 1, "and", 4],
                     "attributes": [],
                     "idx": 0
                   },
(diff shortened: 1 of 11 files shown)


packages/doenetml-worker/lib-doenetml-core/src/dast/flat_dast/untagged_flat_dast_merge.rs
```diff
@@ -197,10 +197,54 @@ impl FlatRoot {
         idx
     }
 
+    /// Convert PathParts from dast in FlatPathPars, adding nodes to the flat root
```
siefkenj, line 200: `Pars` -> `Parts`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_expand.rs
```diff
@@ -146,7 +146,8 @@ impl Expander {
         }
     }
 
-    /// Remove all `extend` attributes from the `FlatRoot` and replace them with the index of their referent.
+    /// Remove all `extend` attributes from the `FlatRoot` and replace them
+    /// an extending field that is a ref_resolution with the index of their referent.
```
siefkenj, line 150: This comment doesn't make sense to me

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.test.rs
```diff
@@ -73,7 +70,7 @@ fn can_resolve_names() {
         referent,
         Ok(RefResolution {
             node_idx: c_idx,
-            unresolved_path: Some(vec![PathPart {
+            unresolved_path: Some(vec![FlatPathPart {
```
siefkenj, line 73: Do we still use `PathPart` at all? Is that how the data is serialized in? Or can we get rid of `PathPart`?

packages/doenetml-worker/lib-doenetml-core/src/dast/flat_dast/untagged_flat_dast_compactify.rs
```diff
@@ -16,6 +16,22 @@ impl FlatNode {
...
+                    }))
+                    .chain(elm.extending.iter().flat_map(|extending| {
+                        extending
+                            .get_resolution()
+                            .unresolved_path
+                            .iter()
+                            .flat_map(|path| {
+                                path.iter().flat_map(|path_part| {
+                                    path_part.index.iter().flat_map(|index| {
+                                        index.value.iter().flat_map(|node| match node {
```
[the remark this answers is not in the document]
dqnykamp: I know, isn't it great! Not sure another way, though. The path parts are highly nested.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.test.rs
dqnykamp: `PathPart` is used with the `Dast`, so yes, the DoenetML is serialized in that form. When we convert to the `FlatDast`, it gets transformed into the `FlatPathPart`.

packages/doenetml-worker/lib-doenetml-core/src/dast/flat_dast/untagged_flat_dast.test.rs
```diff
@@ -38,26 +38,34 @@ fn can_flatten_dast_root() {
                     "type": "Ref",
                     "parent": 1,
                     "path": [
-                      { "type": "pathPart", "name": "x", "index": [] },
-                      { "type": "pathPart", "name": "y", "index": [] },
-                      { "type": "pathPart", "name": "z", "index": [] }
+                      { "name": "x", "index": [] },
```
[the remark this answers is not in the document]
dqnykamp: I added a
```
#[serde(tag = "type")]
#[serde(rename = "flatPathPart")]
```
to `FlatPathPart` to get a type to appear.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_expand.rs
dqnykamp: Hah, that's progress, as I didn't understand the original comment! I missed the preposition "with" in the comment; not sure if that caused confusion.

siefkenj: With the word `with` it makes more sense :-). I think it would be clearer to say
``replace them with `extending: RefResolution` where the `RefResolution` contains the index of the referent`` (or something like that with the correct variable names...)

PR #325, opened 2025-05-01
JXG Fixes

### LINKED ISSUE #323: Make the navigation buttons opaque to clicks
Clicking on the navigation buttons also affects the graph underneath. If there is an object underneath that changes color when clicked, clicking on the navigation button makes the object change color. This gives the impression that something is wrong with the graph.

siefkenj:
This PR fixes #323 

This PR:
 1. Fixes `packages/test-viewer`
 2. Allows you to pan a graph by clicking and dragging
 3. Prevents events from being triggered on both a nav button and a graph element when they overlap on a JXG graph.

### DIFF
--- a/packages/doenetml/src/Viewer/renderers/graph.jsx
+++ b/packages/doenetml/src/Viewer/renderers/graph.jsx
@@ -77,7 +77,7 @@ export default React.memo(function Graph(props) {
             showNavigation: false, // will add navigation buttons later so can style them
             // keepAspectRatio: SVs.identicalAxisScales,
             zoom: { wheel: !SVs.fixAxes },
-            pan: { enabled: !SVs.fixAxes },
+            pan: { enabled: !SVs.fixAxes, needShift: false },
             grid: haveFixedGrid,
         });
 
@@ -116,6 +116,24 @@ export default React.memo(function Graph(props) {
                 }
             }
         });
+
+        // If an object is below the navigation bar and it is clicked, both
+        // the navigation button and the object drag/action are triggered.
+        // We hack our way around this behaviour by preventDefault-ing on all the button events
+        // after the nav bar is created.
+        const preventDefault = (e) => {
+            e.preventDefault();
+        };
+        newBoard.containerObj
+            .querySelectorAll(".JXG_navigation_button")
+            .forEach((button) => {
+                button.addEventListener("click", preventDefault);
+                button.addEventListener("touchstart", preventDefault);
+                button.addEventListener("mousedown", preventDefault);
+                button.addEventListener("mouseup", preventDefault);
+            });
+
+        window.BOARD = newBoard;
         setBoard(newBoard);
 
         previousDimensions.current = {
@@ -171,6 +189,21 @@ export default React.memo(function Graph(props) {
 
         // on unmount
         return () => {
+            // Remove any event listers we added during teardown
+            if (newBoard.containerObj) {
+                newBoard.containerObj
+                    .querySelectorAll(".JXG_navigation_button")
+                    .forEach((button) => {
+                        button.removeEventListener("click", preventDefault);
+                        button.removeEventListener(
+                            "touchstart",
+                            preventDefault,
+                        );
+                        button.removeEventListener("mousedown", preventDefault);
+                        button.removeEventListener("mouseup", preventDefault);
+                    });
+            }
+
             newBoard.off("boundingbox");
         };
     }, []);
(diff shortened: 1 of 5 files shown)


packages/doenetml/src/Viewer/renderers/graph.jsx
```diff
@@ -77,7 +77,7 @@ export default React.memo(function Graph(props) {
             showNavigation: false, // will add navigation buttons later so can style them
             // keepAspectRatio: SVs.identicalAxisScales,
             zoom: { wheel: !SVs.fixAxes },
```
dqnykamp, line 79: Let's add `needShift: false` for zoom, too, as that's how it works for both Desmos and Geogebra.

packages/doenetml/src/Viewer/renderers/graph.jsx
```diff
@@ -116,6 +116,24 @@ export default React.memo(function Graph(props) {
...
         });
+
+        // If an object is below the navigation bar and it is clicked, both
+        // the navigation button and the object drag/action are triggered.
+        // We hack our way around this behaviour by preventDefault-ing on all the button events
+        // after the nav bar is created.
+        const preventDefault = (e) => {
+            e.preventDefault();
+        };
+        newBoard.containerObj
```
dqnykamp, line 127: At least on Ubuntu Chrome, these event listeners are not being added. A console.log of `newBoard.containerObj.querySelectorAll(".JXG_navigation_button")` is an empty list. Even a console.log of `document.querySelectorAll(".JXG_navigation_button")` is giving an empty list at the time this `useEffect()` is called,

dqnykamp: It should work if you move it after the call to `addNavigationButtons` in the `[board]` `useEffect` around line 221.

packages/doenetml/src/Viewer/renderers/graph.jsx
```diff
@@ -116,6 +116,24 @@ export default React.memo(function Graph(props) {
...
+            .querySelectorAll(".JXG_navigation_button")
+            .forEach((button) => {
+                button.addEventListener("click", preventDefault);
+                button.addEventListener("touchstart", preventDefault);
+                button.addEventListener("mousedown", preventDefault);
+                button.addEventListener("mouseup", preventDefault);
+            });
+
+        window.BOARD = newBoard;
```
dqnykamp, line 136: It doesn't appear that setting `window.BOARD` is doing anything. Plus, it would get overwritten if there were more than one graph in a document.

packages/doenetml/src/Viewer/renderers/graph.jsx
dqnykamp: Putting the callbacks in the `[board]` `useEffect` does make the callbacks get called when clicking the navigation buttons. But, it doesn't prevent the clicks from propagating to the graph.

Here's the DoenetML I'm using to test this:
```
<graph>
    <point name="P" x="3" y="1" />
    <rectangle width="40" height="40" center="(0,0)" filled name="R" />
</graph>
<number name="n">1</number>
<updateValue triggerWhenObjectsClicked="P R" newValue="$n+1" target="n" />
```
The idea is that the counter `n` should not increment when a navigation button is clicked even if the rectangle is under the navigation buttons.

dqnykamp: The reason this isn't working is that the `mouseup` event on the button is not being triggered, and `triggerWhenObjectsClicked` is called with the up handler on the rectangle.

siefkenj: Do you have a proposed fix?

packages/doenetml/src/Viewer/renderers/graph.jsx
siefkenj: Woops, this is not supposed to be there.

packages/doenetml/src/Viewer/renderers/graph.jsx
siefkenj: I could complete replace the nav with one that we control so that all events can be properly captured. Would it need anything other than a home and +/- buttons?

dqnykamp: We could follow the lead of jsxgraph and geogebra and just have the home and +/- buttons. In the example I gave, where a rectangle covered the whole graph, one cannot pan without the error buttons, as attempting to pan just moves the rectangle. But, I'm not sure how often that would happen.

dqnykamp (approved):

PR #328, opened 2025-05-02
Correctly adjust indices when compactifying the flat dast

dqnykamp:
The PR fixes a bug in flat dast compactification, where the indices of unresolved paths were not shifted.


siefkenj:
Looks good. It'll be nicer when we restrict the path a bit more so we have to do fewer nested loops.

PR #329, opened 2025-05-05
Add: initial prototype of Markdown in Doenet

bjones1:
Several tests fail; a quick inspection shows failures related to treating text as Markdown, so I'm ignoring these failures until we decide to move in this direction.

The test I added in `dast-basic.test.ts` shows that text is correctly parsed as Markdown by `normalizeDocumentDast`. However, the test-viewer application doesn't render the resulting DAST, even though the resulting DAST tree I print from `normalizeDocumentDast` seems correct when comparing `<p><em>testing</em></p>` and its Markdown equivalent, `*testing*`. Any ideas on why this DAST doesn't render?

The resulting DAST:
image

This renders as:
image

### DIFF
--- a/packages/parser/src/dast-normalize/normalize-dast.ts
+++ b/packages/parser/src/dast-normalize/normalize-dast.ts
@@ -16,6 +16,11 @@ import { selectSugar } from "./component-sugar/select";
 import { postponeRenderSugar } from "./component-sugar/postponeRender";
 import { pluginEnforceValidNames } from "./enforce-valid-names";
 import { pretzelSugar } from "./component-sugar/pretzel";
+import { lezerToDast } from "../lezer-to-dast";
+import { Parser, HtmlRenderer } from "commonmark";
+
+const commonMarkParser = new Parser();
+const commonMarkRenderer = new HtmlRenderer();
 
 /**
  * Normalize the DAST tree so that it is contained in a single `<document>` element.
@@ -25,13 +30,15 @@ export function normalizeDocumentDast(
     dast: DastRoot,
     addCompatibilityNames = false,
 ) {
+    console.log(dast);
     let processor = unified()
         .use(pluginRemoveCommentsInstructionsAndDocStrings)
         .use(pluginChangeCdataToText)
         .use(pluginEnsureDocumentElement)
         .use(pluginConvertPretextAttributes)
         .use(pluginEnforceValidNames)
-        .use(pluginExpandAliasedElements);
+        .use(pluginExpandAliasedElements)
+        .use(pluginTextToMarkdown);
     if (addCompatibilityNames) {
         processor = processor.use(pluginAddCompatibilityNames);
     }
@@ -49,6 +56,29 @@ const pluginChangeCdataToText: Plugin<[], DastRoot, DastRoot> = () => {
             if (node.type === "cdata") {
                 // @ts-ignore
                 node.type = "text";
+            }
+        });
+    };
+};
+
+/**
+ * Interpret all text nodes that are children of the root node as Markdown, transforming them to HTML.
+ */
+const pluginTextToMarkdown: Plugin<[], DastRoot, DastRoot> = () => {
+    return (tree) => {
+        visit(tree, (node, visitInfo) => {
+            // All text nodes that are children of the root node are interpreted as Markdown.
+            if (node.type === "text" && visitInfo.parents.length === 1) {
+                const html = commonMarkRenderer.render(
+                    commonMarkParser.parse(node.value),
+                );
+                // Parse this HTML into a DAST tree, then replace the text node with the new DAST tree.
+                const dast = lezerToDast(html);
+                visitInfo.parents[0].children.splice(
+                    visitInfo.index!,
+                    1,
+                    ...(dast.children as DastElementContent[]),
+                );
             }
         });
     };
--- a/packages/parser/test/dast-basic.test.ts
+++ b/packages/parser/test/dast-basic.test.ts
@@ -13,6 +13,7 @@ import {
     splitTextAtSpecialChars,
     splitTextNodeAt,
 } from "../src/lezer-to-dast/gobble-function-arguments";
+import { normalizeDocumentDast } from "../src/dast-normalize/normalize-dast";
 
 const origLog = console.log;
 console.log = (...args) => {
@@ -469,5 +470,15 @@ describe("DAST", async () => {
             "type": "root",
           }
         `);
+    });
+    it("converts Markdown text to HTML", () => {
+        let source: string;
+        let dast: ReturnType<typeof lezerToDast>;
+
+        source = `*hi*\n\n# there!`;
+        dast = lezerToDast(source);
+        expect(toXml(normalizeDocumentDast(dast))).toEqual(
+            "<document><p><em>hi</em></p><h1>there!</h1></document>",
+        );
     });
 });
--- a/packages/test-viewer/src/test/testCode.doenet
+++ b/packages/test-viewer/src/test/testCode.doenet
@@ -1,3 +1,5 @@
+**test**
+
 <graph>
     <point name="P" x="3" y="1" />
     <rectangle width="40" height="40" center="(0,0)" filled name="R" />
@@ -17,8 +19,8 @@
   <p>A lovely section.</p>
   <problem>
     <p>
-      This problem is nested inside a section. The default number does not
-      include the parent number.
+    This problem is nested inside a section. The default number does not
+    include the parent number.
     </p>
   </problem>
 </section>
@@ -27,8 +29,8 @@
   <p>Another section!</p>
   <problem includeParentNumber="true">
     <p>
-      This exercise is nested inside a section. Rendering of the parent section
-      number has been included.
+    This exercise is nested inside a section. Rendering of the parent section
+    number has been included.
     </p>
   </problem>
 </section>


dqnykamp:
For the test-viewer, first you want to change the default for `usePrototype` on line 32 of `testViewer.tsx` to `true` so you are always using the new DoenetML core.

Then, the problem is that the prototype doesn't yet have many components implemented. It doesn't look like there is an `<em>` component yet. And, unfortunately, Doenet is using the strange `<alert`>` tag from PreTeXt rather than `<strong>`. (I don't think there is a `<strong>` in PreTeXt, but we could always deviate.)

Obvious, we don't want this behavior, but you could turn your markdown into one of the existing components to test it for now, such `<m>`  or `<p>`.  You can see the small list of components in the new DoenetML in `packages/doenetml-worker/src/components/doenet` (for ones that are processed by core) or `packages/doenetml-prototype/src/renderers/doenet` (the React renderers that also catch component types ignored by core). 

Or, I just made a PR to your branch adding an `<em>` renderer using the simple strategy where Doenet core just ignores it for now. With that, "*test*" now renders as italics.

dqnykamp:
Also, for easier testing, you could use the test viewer inside the `doenetml-prototype` package. (That way you don't have to rebuild `doenetml-prototype`, only its dependent packages, to see changes.) Just use "npm run dev" in `packages/doenetml-prototype` and edit the `testCode.doenet` in `doenetml-prototype/dev`.

The current `doenetml-prototype/dev/testCode.doenet` seems to have been saved with DoenetML that currently doesn't work in prototype, so you'll have to remove what is in there to get rid of that error.

bjones1:
@dqnykamp, fantastic, thanks! That's **very** helpful -- know where to look for implemented tags, and how to run the viewer makes all the difference. With that, it looks like this is up and running. I noticed that `ol` and `li` tags, so this works nicely:

image

What's the next step? That is, given we now have the ability to render text as Markdown in the prototype Doenet viewer, does this make sense as a future Doenet feature? @siefkenj, any thoughts on this?

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -46,6 +53,29 @@ const pluginChangeCdataToText: Plugin<[], DastRoot, DastRoot> = () => {
...
+ * Interpret all text nodes that are children of the root node as Markdown, transforming them to HTML.
+ */
+const pluginTextToMarkdown: Plugin<[], DastRoot, DastRoot> = () => {
+    return (tree) => {
+        visit(tree, (node, visitInfo) => {
+            // All text nodes that are children of the root node are interpreted as Markdown.
+            if (node.type === "text" && visitInfo.parents.length === 1) {
+                const html = commonMarkRenderer.render(
+                    commonMarkParser.parse(node.value),
+                );
```
siefkenj, line 66: Very interesting strategy. It's certainly quick for now, but I think it would be better to convert the AST directly so we have full control over the nodes created (e.g, in code blocks, etc.)

siefkenj:
I think the next step is to set up some tests so we can all agree on what should be converted to what :-)

packages/parser/src/dast-normalize/normalize-dast.ts
bjones1: Do you mean that it's possible we'll be using Markdown not just with text nodes, but in other places (inside code blocks, for example)? If so, then your comment below makes a lot of sense -- define where Markdown belongs and where it doesn't.

siefkenj:
Didn't we discuss having markdown only work in the top-level components to start?

bjones1:
> Didn't we discuss having markdown only work in the top-level components to start?

That was the initial discussion and certainly the simplest approach, and (AFAIK) what this PR implements. However, I understood the following to imply this PR should cover more cases (see also [my comment](https://github.com/Doenet/DoenetML/pull/329#discussion_r2074065334)):

> Very interesting strategy. It's certainly quick for now, but I think it would be better to convert the AST directly so we have full control over the nodes created (e.g, in code blocks, etc.)

So, two questions:
1. What's next? A series of test cases for text only in the root node? Or something else?
2. How should we handle Doenet code like this?
    ```
    <section>
    Our goal this semester is to learn about encrpytion and decryption...
      
    Less common terms for these are encipherment and decipherment....
    
    An algorithm is a process or set of rules... 
    </section>
    ```
    
    To me, this text inside the `section` component should render as Markdwon.

siefkenj:
1. Yes, I think that's what's next. Once we have it working for root notes, extending the rules to work for other nodes should be very easy.

2. I'll let @dqnykamp comment on this. It's important to know what rules apply when, but it seems starting in any division would be a good place to enable the syntax.

siefkenj:
Every markdown feature should be tested, ideally in separate tests, and a few that blend some features together.

So: lists, bold, em, links, images, paragraphs (other markdown features I am forgetting?)

I would imagine tests like, `"markdown: paragraphs get converted to multiple <p> tags"`, `"markdown: nested bold and emph tags get converted to nested tags"`, etc.

Then there should be some tests about interleaving formats. So, if a user does `Pay _attention to <math>5</math>_ in the following expression`, does it emph or does it not? The behaviour should be codified in a test, even if the test is marked as `.skip` initially because it is not implemented.

One key thing for the tests to nail down is that: markdown cannot produce invalid DoenetML. If there is a feature of markdown that doesn't exist in doenet, it should be converted to its nearest equivalent.

dqnykamp:
I think it wouldn't be hard to come up with a small list of component types inside which we'd still apply markdown, such as `<section>` (and similar), `<cell>`, `<feedback>`, `<footnote>`, `<hint>`. Do we have different rules for inline tags like `<p>` and `<text>`, which shouldn't have paragraphs inside them?

dqnykamp:
I've frequently been asked for an `<hr/>`. I didn't add it because it wasn't in PreTeXt, and I didn't think of it as being semantic. I figured we'd eventually style sections to give someone the option of having a horizontal rule. But, maybe, I should just break down and add it. It's better than folks kludging it with a fake table and a single border!

>     * Line breaks (`<br/>`) -- I don't see any equivalent. Refuse to translate? Add to Doenet?

I'm not certain about line breaks. What's the use case where paragraphs wouldn't work?

bjones1:
> I've frequently been asked for an `<hr/>`. I didn't add it because it wasn't in PreTeXt, and I didn't think of it as being semantic. I figured we'd eventually style sections to give someone the option of having a horizontal rule. But, maybe, I should just break down and add it. It's better than folks kludging it with a fake table and a single border!

Sounds good. For now, I'll just pass the `<hr/>` through.

> > ```
> > * Line breaks (`<br/>`) -- I don't see any equivalent. Refuse to translate? Add to Doenet?
> > ```
> 
> I'm not certain about line breaks. What's the use case where paragraphs wouldn't work?

Here's one place I think it's helpful. Paragraphs in the default style produce two newlines, whereas at line break produces just one.

$\int_{a}^{b} x^2$ is a definite integral where\
$a$ is the lower limit,\
$b$ is the upper limit, and\
$x^2$ is the integrand.

bjones1:
I'm making good progress; I've just pushed updated code with a number of tests.

I ran into one minor problem and one major problem:

1. Major: the DoenetML parser turns text like `Given <m>x</m> is...` into two text nodes with a math node in the middle. This is bad: Markdown parses the text nodes without knowing there's a math node in the middle of the sentence that the text should be wrapped around. I think our mistake is to use the Doenet parser first; instead, we need to run the Markdown parser, then run that result through the Doenet parser.
2. Minor problem: Markdown allows autolinks: `<http://foo>` becomes <http://foo>. DoenetML interprets this as invalid, so Markdown never gets to look at it.

Thoughts?

bjones1:
More thoughts on this: parsing with Markdown first is a difficult and a bad idea. However, the current parsing idea need to change. A sketch:

1. While walking the parse tree, if a text node's parent is a Markdown-compatible component, then:
2. Identify the nearest previous sibling which is a block (not an inline) component, and the nearest next sibling which is a block component.
3. Convert nodes between (but not including) these siblings into XML: if a node is not Markdown-compatible, ensure it's separated by newlines and contains no newlines, so that Markdown will leave it untouched.
4. Parse with Markdown, convert to DoenetML, then re-parse as Doenet ML.
5. Replace these nodes with the resulting parse output.

For example, consider the following DoenetML:

```XML
<p>This is a block component. This component does not contain Markdown.</p>

<m>x^2</m> is some math in a paragraph that **should** be interpreted as Markdown. 

<choiceInput> (a block component, doesn't contain Markdown)
  <choice>**Agree**</choice>  (These are can contain Markdown)
  <choice>Disagree</choice>
</choiceInput>
```

We have something like this for the parse tree:

Doenet parser drawio

In the diagram, the dotted boxes indicate areas containing Markdown text to be processed as described above.

siefkenj:
>     2. Minor problem: Markdown allows autolinks: `<http://foo>` becomes http://foo. DoenetML interprets this as invalid, so Markdown never gets to look at it.
I think we just count that syntax as invalid. Markdown also interprets bare links as links, so there is no need for the `<...>` syntax.

packages/parser/src/dast-normalize/normalize-dast.ts
```diff
@@ -46,6 +53,72 @@ const pluginChangeCdataToText: Plugin<[], DastRoot, DastRoot> = () => {
...
+/**
+ * Interpret all text nodes that are children of the root node as Markdown, transforming them to HTML.
+ */
+const pluginTextToMarkdown: Plugin<[], DastRoot, DastRoot> = () => {
+    let processor = unified().use(pluginHtmlToDoenetml);
+    return (tree) => {
+        visit(tree, (node, visitInfo) => {
+            // All text nodes that are children of the root node are interpreted as Markdown.
+            if (node.type === "text" && visitInfo.parents.length === 1) {
+                const parse_tree = commonMarkParser.parse(node.value);
```
siefkenj, line 65: I'd really like to work directly with `parse_tree` here instead of turning it into a string and reparsing it.

siefkenj:
Let's focus on the simple situation for now where there is no mixing of XML and Markdown. Handling situations like `This _important <m>x</m>_` can wait until we figure out if this is a feature people really need.

packages/parser/src/dast-normalize/normalize-dast.ts
bjones1: Thanks for taking the time to look through this code. I appreciate the feedback!

I agree that transforming directly from the CommonMark parse tree to the DAST parse tree would be more elegant. However, I suggest postponing optimizations (such as this) for later, when we're confident this will be adopted. My thought process:

Current approach:
- Pro: currently works.
- Con: inefficient -- the parse tree is converted to text, then parsed as DAST, then  CommonMark nodes are transformed to DAST.

This suggested approach:
- Pro: more efficient, since it skips parsing HTML to DAST.
- Con: requires more code, since items that don't currently need to be translated between HTML and DAST now need code added. Also need code to convert between the CommonMark parse tree and the DAST format. 

Notes on this approach: the [usage](https://github.com/commonmark/commonmark.js?tab=readme-ov-file#usage) section of the CommonMark docs shows a handy `walker` function to walk the CommonMark parse tree, and the docs seems pretty good. Is there similar docs on the DAST structure, which would make this easier?

My conclusion: 
In this end, this is an optimization, not a change in functionality. Therefore, I'd suggest this as work for later, when we're more confident that this is the right approach.

bjones1:
> Let's focus on the simple situation for now where there is no mixing of XML and Markdown. Handling situations like `This _important <m>x</m>_` can wait until we figure out if this is a feature people really need.

Sounds good. To the best of my knowledge, the current test suit evaluates all Markdown features and either translates them to DoenetML or produces known errors (Markdown autolinks aren't supported; mixed XML and Markdown isn't supported).

The next step AFAIK is to get feedback: is this a feature people need? Therefore, what's the best approach to gather some feedback?

bjones1:
Thoughts on the approach to solving this:

1. Walk the DAST tree. Classify each node as either not markdown, inline markdown, or block markdown. Combine all inline/block  text into a single, large string with separators. Note that the classification of each node should be saved when processing children then restored afterward.
2. Convert these strings to HTML using Commonmark. Split the string into fragments to apply to each inline/block markdown node. For inline markdown nodes, remove the leading `<p>` / trailing `</p>` tags.
3. Convert each fragment to DAST. Replace the text node with the resulting DAST.

Notes:
- The separators used should ensure that each inline/block string starts clean (not inside another block, or a fenced code block, etc.)
- Need a way to match an fragment index (produced by splitting the HTML post Markdown conversion) to a node of the DAST.

siefkenj:
There is a `replaceNode` function that should allow you to replace any node with a single node or a sequence of nodes. So if you have

```
{type: "text", value: `Two Markdown

Paragraphs`}
```
You should be able to do a `replaceNode` and return an array of the two paragraphs.

If you focus just on blocks of text, does the approach work? (e.g., excluding your example that has `<m>x^2</m>` embedded)

bjones1:
Pure blocks of text without XML work.

Getting Markdown to work inside XML block tags (like `<p>**Strong** <m>x^2</m></p>`) should be possible. It will take a bit more refinement.

Getting block Markdown (blockquote, paragraph, code blocks) which contains XML tags is hard, and probably needs the process I sketched out earlier.

PR #338, opened 2025-05-15
Fix the vscode extension

siefkenj:
This PR fixes data-uri nonesense when loading the `.wasm` file for the Rust core so that the vscode extension works.

### DIFF
--- a/packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
+++ b/packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
@@ -47,7 +47,6 @@ try {
         wasmBlobUrl = URL.createObjectURL(
             new Blob([byteArray], { type: "application/wasm" }),
         );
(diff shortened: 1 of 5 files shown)


packages/doenetml-worker/lib-js-wasm-binding/src/CoreWorker.ts
```diff
@@ -25,6 +25,34 @@ import {
...
+        const byteCharacters = atob(base64);
+        const byteNumbers = new Uint8Array(byteCharacters.length);
+        for (let i = 0; i < byteCharacters.length; i++) {
+            byteNumbers[i] = byteCharacters.charCodeAt(i);
+        }
+        const byteArray = new Uint8Array(byteNumbers);
+        wasmBlobUrl = URL.createObjectURL(
+            new Blob([byteArray], { type: "application/wasm" }),
+        );
+        console.log("BLOB URL", wasmBlobUrl);
```
dqnykamp, line 50: Remove console.log

PR #455, opened 2025-05-23
Clean up `ignoreUpdates` logic

siefkenj:
This PR:
 * ads a Redux store and Recoil compatibility layer that will (soon) be used to replace Recoil (but it currently isn't used)
 * Cleans up the `upatesToIgnore` logic and removes its dependence on Recoil.


packages/doenetml-worker-javascript/src/CoreWorker.ts
```diff
@@ -234,6 +234,8 @@ export class PublicDoenetMLCore {
         }
 
         try {
+            // XXX: temporary to slow things down for testing
```
dqnykamp, line 237: Remove this before we merge the PR

packages/doenetml/src/Viewer/DocViewer.tsx
```diff
@@ -698,13 +589,19 @@ export function DocViewer({
...
-                    coreId: coreId.current,
-                    componentIdx: Number(componentIdx),
-                    stateValues: args.rendererState[componentIdx].stateValues,
-                    childrenInstructions:
-                        args.rendererState[componentIdx].childrenInstructions,
-                });
+                dispatch(
+                    mainThunks.updateRendererSVs({
+                        coreId: coreId.current,
+                        componentIdx: Number(componentIdx),
```
dqnykamp, line 595: This will currently break given that `componentIdx` won't be turned into a number until the code that removes namespaces is merged! At this point, it is still a string so one should get `NaN` here. (This code is only called when loading in state from the database, so it's probably not being triggered when you test it.)

dqnykamp:
That's great that recoil is finally eliminated!

`recoil` is still referenced in `ui-components/vite.config.ts` and `utils/vite.config.ts`. And there is a `recoil-virtual-keyboard` that should presumably be renamed.

packages/doenetml/src/Viewer/DocViewer.tsx
siefkenj: I'll remove this and we'll just deal with the TS error for now.

dqnykamp:
The cypress tests are finding bugs. Here's one: pressing Enter on a mathInput should trigger the `value` state variable to update. Right now, it is not being updated on Enter, only on blur. 

Example DoenetML:
```
<mathInput name="mi" />
<p>Value: $mi</p>
<p>Immediate value: $mi.immediateValue</p>
```

dqnykamp:
There were 61 cypress test failures, but all the ones I've spot-checked so far were from that above bug in mathInput. I'll rerun them after that bug is fixed.

dqnykamp:
Another bug: when using the virtual keyboard, the blur event happens before every keystroke, causing the `value` state variable to be updated to be the content just before the last keystroke.

siefkenj:
> Another bug: when using the virtual keyboard, the blur event happens before every keystroke, causing the `value` state variable to be updated to be the content just before the last keystroke.

Fixed!

dqnykamp (approved):
Great to have recoil gone!

PR #460, opened 2025-05-27
Redo ui-components

siefkenj:
This PR refactors the `ui-components` with modern CSS handling most of the functionality.

The `ui-components` are not used very many places, so it should be easy to test them everywhere they are used. They also seem to include many attributes that are not used anywhere (maybe they used to be used?)


packages/ui-components/src/components/ToggleButton.tsx
```diff
@@ -1,198 +1,88 @@
...
-                aria-disabled={props.disabled ? true : false}
+        <label
+            className={classNames(
+                "doenet-toggle-button-label",
+                "doenet-button",
+                "action-button",
+            )}
+        >
+            <Input
+                value=""
```
dqnykamp, line 62: The cypress test "boolean input as toggle button" from `booleaninput.cy.js` is now failing as it cannot click the toggle button (given by its id) because the button is not visible. 

The message is:

> <input aria-checked="true" id="\/bi_input" class="doenet-toggle-button" type="checkbox" value="" checked="">
> 
> This element <input#\/bi_input.doenet-toggle-button> is not visible because it has an effective width and height of: 0 x 0 pixels.
> 
> Fix this problem, or use {force: true} to disable error checking.

Changing lines 88 and 107 in that file to `click({ force: true })` as suggested does make the test pass again. (We often have to use that `force: true` trick to make cypress test pass, so this workaround isn't unusual.)

siefkenj: I changed togglebuttons to inputs instead of buttons. Maybe that is the reason?

siefkenj: Ah. I just didn't read the error. Yeah, I hid the actual inputs and they are activated by clicking on the label instead.

dqnykamp: Another strategy to get the cypress test to pass, then, could be to move `props.id` to be the id of the label.

PR #462, opened 2025-05-28
Replace resizable panels with non-chakra components

siefkenj:


packages/ui-components/src/test-main.tsx
```diff
@@ -159,6 +171,38 @@ function App() {
                 <ToggleButton value="Toggle points and intervals"></ToggleButton>
                 <ToggleButton value="Move Points"></ToggleButton>
             </ToggleButtonGroup>
+
+            <h3>Resizable Pair</h3>
+
+            <div style={{ height: "100px", width: "100%" }}>
```
dqnykamp, line 177: I'm getting extra vertical and horizontal scrollbars despite this 100% width and height as there appears to room to scroll a tiny bit in both directions. It's more prominent in Chrome, as seen in this screenshot.
 
image

PR #466, opened 2025-05-30
Remove Chakra

siefkenj:
This PR removes Chakra and replaces it with reakit :-)


siefkenj:
> s white gap below the errors/warning an intentional design change or just an inadvertent styling issue?

It wasn't intentional, but I didn't think it looks so bad. I can remove it if you like.

dqnykamp:
> > s white gap below the errors/warning an intentional design change or just an inadvertent styling issue?
> 
> It wasn't intentional, but I didn't think it looks so bad. I can remove it if you like.

Actually, there is a white gap in the previous version. It just wasn't as prominent. If I'm allowed to nitpick, then I'd suggest making the gap smaller. I think it good to have some visual divider.

image

siefkenj:
> > > s white gap below the errors/warning an intentional design change or just an inadvertent styling issue?
> > 
> > 
> > It wasn't intentional, but I didn't think it looks so bad. I can remove it if you like.
> 
> Actually, there is a white gap in the previous version. It just wasn't as prominent. If I'm allowed to nitpick, then I'd suggest making the gap smaller. I think it good to have some visual divider.
> 
> image

Turns out this changes depending on the window height, so I will fix it.

PR #472, opened 2025-06-05
Add component index to ignore-update key

dqnykamp:
This PR fixes a bug where updates to renderers could be incorrectly ignored due to the key of `updatesToIgnore` not including `componentIdx`.


packages/doenetml/src/state/slices/main.ts
```diff
@@ -92,7 +92,9 @@ export const mainThunks = {
                 const updatesToIgnore = updatesToIgnoreRef.current;
 
                 if (updatesToIgnore.size > 0) {
-                    let valueFromRenderer = updatesToIgnore.get(actionId || "");
+                    let valueFromRenderer = updatesToIgnore.get(
+                        `${actionId}|${componentIdx}`,
```
siefkenj, line 96: I thought this change appeared in 3 other places?

dqnykamp: Yeah, my bad. I checked in the wrong version where I was testing that the test broke when I reverted! It is fixed now.

PR #490, opened 2025-06-12
Resolve path immediately utility function in core

dqnykamp:


packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
```diff
@@ -0,0 +1,278 @@
...
+ * starting the search algorithm at node `origin`.
+ *
+ * Throws an error if the name is not resolved.
+ */
+export async function resolvePathImmediatelyToNodeIdx(
+    name: string,
+    coreWorker: PublicDoenetMLCore,
+    origin = 0,
+) {
+    console.log("resolve path immediately", { name, origin });
```
siefkenj, line 20: This is here because it is a debug-only function? Still, if it is used in the tests it will print a lot.

packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
```diff
@@ -0,0 +1,278 @@
...
+            resolution,
+            resolver,
+            coreWorker,
+        );
+
+        if (newResolution.unresolvedPath == null) {
+            return newResolution.nodeIdx;
+        } else {
+            // If we have any unresolved path left, then the entire `path` did not match a node.
+            return -1;
```
siefkenj, line 79: Sentinel values are not my favorite for return types.

packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
```diff
@@ -0,0 +1,278 @@
+import {
```
siefkenj, line 1: This is only used for testing, right? Do we have a place to put testing-only functions? If not, the file/functions should be clearly labeled, but my guess is that you can move this directly to the `test/` folder. If it does not end in `.test.ts` it will be ignored by vitest.

packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
dqnykamp: This was left here inadvertently.

packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
dqnykamp: I don't have a better idea. This function is already used 40K times in tests as an index to an array!

packages/doenetml-worker-javascript/src/utils/externalPathResolution.ts
dqnykamp: This function is used not just for testing but also for debug functions from the production server. Eventually, this should be moved into `debug-hooks` workspace or something similar. But, that ended up being a bigger task than seems reasonable to do at the moment, given that we have to deal with a web worker in some contexts and not in others. For now, I'm leaving it here but with a TODO to eventually move it.

PR #491, opened 2025-06-13
a utility function for determining the component type of a prop

dqnykamp:
This function to determine the component type of a prop should be helpful for making a conversion script.


packages/doenetml-worker-javascript/src/utils/debug/propType.ts
```diff
@@ -0,0 +1,151 @@
...
+ * If successful, returns the component type. Throws an error if it fails.
+ *
+ * Function will fail if `propName` is not a public state variable of `componentType`.
+ *
+ * Function will also fail for the edge case where `propName` has variable component type
+ */
+export function determinePropType(
+    componentType: string,
+    propName: string,
+    nIndices = 0,
```
siefkenj, line 17: This is the number of times it is indexed? I.e., for `$point.x[1][1]`, `nIndices === 2`?

dqnykamp: That's right.

PR #493, opened 2025-06-16
add the ability for the resolver to resolve constant indices

dqnykamp:


packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -128,13 +148,20 @@ pub struct Resolver {
     /// `name_map[0]` is a map of all names accessible from the `flat_root`
     /// and `name_map[i+1]` is a map of all names accessible from node `i`.
     name_map: Vec<HashMap<String, Ref>>,
+    /// Map of resolutions of indices that follow a match to a node, where only indices of `index_resolutions` itself are shifted by `1`.
+    /// Given a path `initial_part[i+1]` where `initial_part` has matched node `j`,
+    /// we look up `index_resolutions[j+1][i]` to determine if the (already shifted) index `i+1` matches a node.
+    /// If `index_resolutions[j+1][i] = Some(k)`, then the path `initial_part[i+1]` resolves to node `k`.
+    /// Note: `index_resolutions[j+1][i] = None` indicates that `initial_part[i+1]` would resolve to a text node,
+    /// which is not a valid ref resolution.
+    index_resolutions: Vec<Vec<Option<Index>>>,
```
siefkenj, line 157: We now have a bunch of vectors of the same length hanging around. I think we got here by adding them one by one, but I'm not sure if its serving us well anymore.

So, every node as a `parent`, `resolution_algorithm`, `name_map`, and now `index_resolutions`. I think we can make a datastructure with all of those and have the Resolver just work on a `Vec` of that structure.

There should also be less talking about `+1` on indices, I think...

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -171,7 +205,11 @@ impl Resolver {
     ///
     /// Arguments:
     /// - `flat_fragment`: a `FlatFragment` containing the new descendants added to `flat_fragment.parent_idx`.
-    pub fn add_nodes(&mut self, flat_fragment: &FlatFragment) {
+    pub fn add_nodes(
+        &mut self,
+        flat_fragment: &FlatFragment,
+        as_index_resolutions: Option<AsIndexResolutions>,
```
siefkenj, line 211: Not a fan of the `Option` here. Maybe add a `ResolutionType::Node(...)` and `ResolutionType::Index(...)` enum to wrap `flat_fragment`.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve.rs
```diff
@@ -219,6 +257,72 @@ impl Resolver {
...
+                as_index_resolutions.parent_idx,
+                &flat_fragment.children,
+                as_index_resolutions.replace_mode,
+            );
+        }
+    }
+
+    /// Find any elements in `flat_root_or_fragment` that are marked `CHILDREN_ARE_IMPLICIT_INDEX_RESOLUTIONS`
+    /// and add index resolutions so that their children will resolve as their indices
+    fn add_implicit_index_resolutions(&mut self, flat_root_or_fragment: &FlatRootOrFragment) {
```
siefkenj, line 276: I think this would be better as a more general function where you pass in `CHILDREN_ARE_IMPLICIT_INDEX_RESOLUTIONS` instead of reference a global directly.

siefkenj: Keep the global declaration, just pass it into the function so the function is self-contained (and more testable)

PR #495, opened 2025-06-17
remove the `sourceAttributesToIgnore` attribute and the special case for the `isResponse` attribute

dqnykamp:
This PR removes the `sourceAttributesToIgnore` on the abstract `<_copy>` tag (which is what the `extend` attribute produces) and the `<collect>` tag. The attribute was not widely publicized, and its main purpose was to prevent the `isResponse` attribute from being copied when referencing a component. However, we decided it was an unnecessary complexity and it was better to have consistent rules for all attributes. Now, `isResponse` is copied on references like all attributes (other than the one remaining exception of `labelIsName`).


siefkenj:
Love to see some cleanup!

PR #503, opened 2025-06-25
Calculate root names of components as shorted accessible name

dqnykamp:
This PR attempts to calculate "root names" of each component by inverting the resolver and finding the simplest name that resolves to the component when the origin is the root of the document. If a root name is found, that name is used for the id by the renderer.

Since the root names are designed to be HTML ids, the notation `[1]` for indices is replaced by `:1`. For example, if a component's shortest accessible reference from the root is `$a.b[3].c`, the root name will be `a.b:3.c`.


packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+///   "Simplest" is defined as the fewest number of pieces (indices or sub-names) used to construct the name
+///   (with ties broken by the fewest number of indices or the shortest overall name length).
+///
+/// For example, consider the DoenetML.
+/// ```xml
+/// <a name="y"><group name="g"><b /><c name="x" /></group></a>
+/// <d name="x" />
+/// ```
+/// The results for different nodes would be
+/// - `<a>`: `Some("x")`, as `<a>` is directly resolvable from the root as `$x`
```
siefkenj, line 98: `x` -> `y`

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+    // Note: resolver's data has one more entry than number of nodes as it also contains the root
+    let n_nodes = resolver.node_resolver_data.len() - 1;
+
+    // First, calculate for each node, origins from which it can be resolve a a single names or index
+    let mut resolutions_to_node: Vec<ResolutionsToNode> =
+        iter::repeat_with(ResolutionsToNode::default)
+            .take(n_nodes)
+            .collect::<Vec<_>>();
+
+    for (node_idx_plus_1, node_data) in resolver.node_resolver_data.iter().enumerate() {
```
siefkenj, line 112: It may not be worth the effort, but I could imagine an opaque type for `node_idx_plus_1` where `NodeOrRoot::from_index(...)` would automatically subtract 1 or return a Root as appropriate.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+                                panic!("Inconsistent resolver data as node can be reached by two different names");
+                            }
+
+                            name_resolution.origins.push(origin);
+                        }
+                        Ref::Ambiguous(_) => {}
+                    }
+                }
+            }
+            ResolutionAlgorithm::DontSearchChildren | ResolutionAlgorithm::Unsearchable => {}
```
siefkenj, line 141: Put these at the start of the match

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
siefkenj: I think this could be done with a `vec.extend(...)` on a filtered iterator, rather than having the `name_resolution.origins.push` buried in the code.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+    reachable_names
+        .into_iter()
+        .map(|possible_names| {
+            possible_names.unwrap().into_iter().reduce(|acc, name| {
+                let index_count_name = name.chars().filter(|c| *c == ':').count();
+                let index_count_acc = acc.chars().filter(|c| *c == ':').count();
+                let separator_count_name =
+                    index_count_name + name.chars().filter(|c| *c == '.').count();
+                let separator_count_acc =
+                    index_count_acc + acc.chars().filter(|c| *c == '.').count();
```
siefkenj, line 179: Can you do this computation before you do `concatenate_reachable_names`? It seems a little strange to turn things to a string, erasing the data types, then re-guessing the data types based on the string.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+    if let Some(ResolutionViaName { name, origins }) = &resolutions.by_name {
+        // Since `node_idx` can be reached via `name` from `origins`,
+        // determine how one can reach each origin from root and concatenate `name` with the result
+        for origin in origins.iter() {
+            match origin {
+                // node can be directly reached from root via `name`
+                NodeOrRoot::Root => names.push(name.to_string()),
+                // since `origin` is not the root, we need to recurse
+                NodeOrRoot::Node(dest_idx) => {
+                    reachable_names = concatenate_reachable_names(
```
siefkenj, line 227: This recursion happening from multiple places makes the code hard to follow. Can this process be done with a simple iterator instead?

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+/// See [`ResolutionsToNode`] for additional details and an example.
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "web", derive(Tsify))]
+#[cfg_attr(feature = "web", tsify(into_wasm_abi, from_wasm_abi))]
+pub struct ResolutionViaName {
+    name: String,
+    origins: Vec<NodeOrRoot>,
+}
+
+/// The `origin` for which a reference to the index `index` resolves to a given node.
```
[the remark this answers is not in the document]
dqnykamp: The resolver has the concept of `indexResolution`, which is using the terminology of the index of the reference `$a[2]`, not the component index. I've been so focused on the terminology from `RefResolution`, etc., that I didn't even notice the ambiguity! I hesitate to use `child`, as we decided against making `$a[1]` reference the 1st child of `a`. It only works in special circumstances where components where added explicitly as indices of their parent (or for `group`, which is in `CHILDREN_ARE_IMPLICIT_INDEX_RESOLUTIONS`).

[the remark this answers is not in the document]
dqnykamp: And, actually, at least with some of the convolutions introduced in the JS Core (with some intermediate `<copy>s` used as intermediaries), an index resolution sometimes isn't a child. I hope we can get rid of the wrinkle later, though.

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+///   where `origin` is the index of the node `<group>`.
+///   This value arises from the fact that  `<group>` takes index resolutions and `<c>` is the referent of the second index.
+///
+/// For node `<a>`, only the field `by_name` exists, with `name` being `"y"` and `origins` containing the root (`NodeOrRoot::Root`).
+///
+/// For node `<d>`, `ResolutionsToNode` would contain no entries, as `<d>` cannot be uniquely resolved from anywhere.
+#[derive(Default, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "web", derive(Tsify))]
+#[cfg_attr(feature = "web", tsify(into_wasm_abi, from_wasm_abi))]
+pub struct ResolutionsToNode {
```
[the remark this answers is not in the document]
dqnykamp: This isn't an address of a node. In my first pass, I called it `SingleHopResolutions` or something like that, as it is the result of the resolution of a single name or a single index, i.e., a single hop in the resolution path. I struggled with coming up with reasonable terms...

packages/doenetml-worker/lib-doenetml-core/src/dast/ref_resolve/root_names.rs
```diff
@@ -0,0 +1,269 @@
...
+/// - `<d>`: `None`, as `<d>` does not have a unique reference from the root
+pub fn calculate_root_names(resolver: Resolver) -> Vec<Option<String>> {
+    // Note: resolver's data has one more entry than number of nodes as it also contains the root
+    let n_nodes = resolver.node_resolver_data.len() - 1;
+
+    // First, calculate for each node, origins from which it can be resolve a a single names or index
+    let mut resolutions_to_node: Vec<ResolutionsToNode> =
+        iter::repeat_with(ResolutionsToNode::default)
+            .take(n_nodes)
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
dqnykamp: Yes, if we derive `Clone`

PR #517, opened 2025-06-30
replace `<ref>` with new `<link>` component

dqnykamp:
This PR replaces the `<ref>` component with a new `<link>` component that has a significantly different syntax and approach.

The link component has a single attribute `to` for specifying  its target. It can be used for
- a link to a component on the page via a reference: `<link to="$section1">Go to section 1</link>`
- a link to another Doenet document: `<link to="doenet:aContentIdToDoc">another doc</link>`
- a link to an external URL: `<link to="https://doenet.org">Doenet</link>`

For a link to another Doenet document, the `<DoenetViewer>` and `<DoenetEditor>` now take a single single attribute `doenetViewerUrl` and the resulting URL for `doenet:aContentIdToDoc` becomes "`doenetViewerUrl`/aContentIdToDoc". Any additional parameters to the `doenet:` reference are no longer interpreted but are just appended to the URL for the endpoint to process.

### DIFF
--- a/packages/test-cypress/cypress/e2e/tagSpecific/link.cy.js
+++ b/packages/test-cypress/cypress/e2e/tagSpecific/link.cy.js
@@ -22,7 +22,7 @@ describe("link Tag Tests", function () {
       <p>Paragraph seven</p>
       <p>Paragraph eight</p>
       <p>Goto:
-      <link name="toTwo" to="$section2">Section 2</link>,
+      <link name="toTwo" to="$section2" />,
       <link name="toThree" to="$section3">Section 3</link>
       <link name="toFour" to="$section4">Section 4</link>
       <link name="toThreeii" to="$section3.p2">Second paragraph of Section 3</link>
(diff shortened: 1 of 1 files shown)


siefkenj:
> > Was `link` an existing component or is it being added?
> 
> It is the `ref` component that we renamed to `link`.

Oh, what was the decision process there? This deviates from PreTeXt convention.

dqnykamp:
> > > Was `link` an existing component or is it being added?
> > 
> > 
> > It is the `ref` component that we renamed to `link`.
> 
> Oh, what was the decision process there? This deviates from PreTeXt convention.

We devoted quite a bit of time during an Authoring meeting to decide what to do. My initial proposal was to rename `<ref>` to be `<xref>` to make it more like PreTeXt. I wasn't overly happy with that, as I don't like the name `<xref>` and find combining that with a `ref` attribute a bit odd, but I thought it would be simpler to maintain component name compatibility where possible. The consensus from that meeting was that one needs a conversion process anyway to go from Doenet to PreTeXt, so it wasn't work making Doenet authors use a confusingly named tag. We also discussed whether or not to use the same tag name for an external link versus an internal reference (e.g., have `<link>` and `<ref>`) or stick with the Doenet convention of using the same for both. The consensus was what is in the PR, where we have a single `<link>` component and everything uses the `to` attribute.

You're always welcome to attend Authoring meetings if you want to be part of the discussions!

PR #521, opened 2025-06-30
Add `extend` and `copy` attributes to schema for all components

dqnykamp:
The PR adds an `extend` and a `copy` attribute to the base component so they show up in the schema. It also does a little clean up on static assets.

### DIFF
--- a/packages/static-assets/scripts/get-schema.ts
+++ b/packages/static-assets/scripts/get-schema.ts
@@ -161,9 +161,7 @@ export function getSchema() {
         let children: string[] = [];
         let acceptsStringChildren = false;
 
-        // All components have the name and copySource attributes,
-        // even though they aren't in the attributes object
-        let attributes = [{ name: "name" }, { name: "copySource" }];
+        let attributes = [];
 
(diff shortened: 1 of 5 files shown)


packages/static-assets/scripts/get-schema.ts
```diff
@@ -161,9 +161,7 @@ export function getSchema() {
         let children: string[] = [];
         let acceptsStringChildren = false;
 
-        // All components have the name and copySource attributes,
-        // even though they aren't in the attributes object
-        let attributes = [{ name: "name" }, { name: "copySource" }];
+        let attributes = [];
```
siefkenj, line 164: Is `name` included in the schema now?

dqnykamp: Yeah, it was in there twice before. I hadn't realized that. But, I had added a `name` attribute to the base component in #318, which was causing `name` to go in twice!

siefkenj: Okay. The rest looks good to me :-)

PR #522, opened 2025-06-30
Convert `$(../x)` macro syntax with a warning

siefkenj:


dqnykamp:
I assume that eventually we can display the warnings in editor.

PR #551, opened 2025-07-19
parent name supersedes descendants if ref resolutions

dqnykamp:
This PR changes how ref resolutions work when a parent has the same name as one or more descendants. With this PR, the parent name supersedes the descendants. From outside the parent, references to the shared name are no longer ambiguous but resolve to the parent.

For example, given this DoenetML
```
<a name="x"><b name="x" /></a>
$x
$x.x
```
the reference `$x` now uniquely resolves to `<a>` rather than being ambiguous. By extension, the reference `$x.x` now uniquely resolves to `<b>`.

This change makes the ref resolution behave in a more natural way, as `<a>` is "closer" to the references.

### DIFF
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
@@ -298,25 +298,23 @@ impl Resolver {
                                         // then the parent supersedes the name of the descendant.
                                         // We don't change the name map, and we stop looping to additional parents
                                         break;
-                                    } else {
-                                        if let Some(Ref::Unique(descendent_idx)) =
-                                            descendant_ref_of_own_name
-                                        {
-                                            if *idx == descendent_idx {
-                                                // If the `name_with_source` resolves to the descendant that `element` is superseding,
-                                                // then replace it with `element`
-                                                *x = Ref::Unique(element.idx);
-                                            } else {
-                                                // Since is already something sharing the `name_with_source` with the current element,
-                                                // the reference becomes ambiguous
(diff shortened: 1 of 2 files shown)


packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
```diff
@@ -276,24 +288,79 @@ impl Resolver {
...
-                                // so we mark it as ambiguous
-                                Ref::Unique(idx) => Ref::Ambiguous(vec![*idx, element.idx]),
+                                Ref::Unique(idx) => {
+                                    if *idx == previous_parent {
+                                        // if `name_with_source` resolves to the `previous_parent`,
+                                        // then the parent supersedes the name of the descendant.
+                                        // We don't change the name map, and we stop looping to additional parents
+                                        break;
+                                    } else if let Some(Ref::Unique(descendent_idx)) =
+                                        descendant_ref_of_own_name
```
siefkenj, line 302: I think we can `.get` `descendant_ref_of_own_name` right here and avoid looking stuff up in some cases. Maybe we can also avoid the `cloned()`

dqnykamp: We can't because we have a open mutable reference to `descendant_names[parent_idx_plus_1]` starting at line `291`. That's why I had to move it outside and add a `cloned()`. Almost all the time `descendant_ref_of_own_name` is going to be `None`, so I hope it doesn't slow things down too much. And, if it isn't `None`, then the algorithm will reach one of the two places where it is used. I can't think of a better way to do it given rust borrowing rules.

PR #552, opened 2025-07-20
improve resolver speed

dqnykamp:
This PR improves the speed of adding and deleting elements to/from the resolver. It removes the use of a regular expression to parse attribute names and removes unnecessary collections of iterators into vectors.

### DIFF
--- a/packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
+++ b/packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
@@ -1,4 +1,3 @@
-use regex::Regex;
 use rustc_hash::FxHashMap;
 use std::{iter, mem, ops::Range};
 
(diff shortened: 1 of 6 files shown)


packages/doenetml-worker-rust/lib-doenetml-core/src/dast/ref_resolve/build_resolver.rs
```diff
@@ -333,32 +335,42 @@ impl Resolver {
...
+    element: &FlatElement,
+) -> impl Iterator<Item = NameWithSource> + use<'_> {
+    element.attributes.iter().filter_map(|attr| {
+        // If the attribute name is `name` or of the form `source-n:name` where `n` is an integer,
+        // then determine both the name (which is the attribute value)
+        // and the source doc
+        let source_doc: SourceDoc = if attr.name == "name" {
+            // for the attribute 'name`, the source doc is the element's source doc`
+            element.source_doc.into()
+        } else if attr.name.starts_with("source-") && attr.name.ends_with(":name") {
```
siefkenj, line 348: Add a comment with a date that says we previously tried a regex here but it produced a major slowdown on [specify test]

dqnykamp:
> Looks good. Not sure why the PreTeXt test is timing out...

It seems that test frequently randomly fails. Should we move it to its own job so we can just rerun that one when it fails?

PR #556, opened 2025-07-21
Split v06-to-v07 into its own package

siefkenj:


packages/parser/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "@doenet/parser",
     "type": "module",
     "description": "Parse DoenetML",
-    "version": "*",
+    "version": "0.7.0",
```
dqnykamp, line 5: Could you make this `0.7.0-alpha50` so that it matches the three other npm packages `@doenet/doenetml`, `@doenet/standalone`, and `@doenet/doentml-iframe`. My plan is to always publish them together with the same version, even if one of the packages doesn't actually change.

dqnykamp: Actually, this is in the wrong place. Should be for the `v06-to-v07` package that we'll publish in npm.

dqnykamp (approved):

PR #585, opened 2025-08-06
v06-to-v07 exports toXml

dqnykamp:
This PR adds `toXml` to the exports of `v06-to-v07`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20345,7 +20345,7 @@
         },
         "packages/doenetml": {
             "name": "@doenet/doenetml",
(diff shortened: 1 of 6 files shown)


siefkenj (approved):

siefkenj:
I suppose you want to use this independently with just the `v06-to-v07` package? It is already exported from `parser`.

dqnykamp:
Right, but we don't publish parser to npm currently, so I couldn't figure
out another way to get it from DoenetTools.

PR #610, opened 2025-09-27
add aria labels to check work buttons

### LINKED ISSUE #597: Make check work buttons accessible
The shortcut check work buttons created by, for example, `<answer>2</answer>` do not give any information to screen readers that they are as submit button (and similarly for the incorrect/correct "buttons"). They are simply a button with an image inside.

dqnykamp:
This PR makes check work buttons accessible by adding aria-labels that explain their function. It also changes the answer messages to be spans rather than buttons, as they don't have an action.

Fixes #597

### DIFF
--- a/packages/test-cypress/cypress/e2e/tagSpecific/answer.cy.js
+++ b/packages/test-cypress/cypress/e2e/tagSpecific/answer.cy.js
@@ -38,7 +38,11 @@ describe("Answer Tag Tests", function () {
             let mathInputPartialAnchor = cesc("#" + mathInputName + "_partial");
 
             // cy.get(mathInputAnchor).should('have.value', '');
-            cy.get(mathInputSubmitAnchor).should("be.visible");
+            cy.get(mathInputSubmitAnchor).should(
+                "have.attr",
+                "aria-label",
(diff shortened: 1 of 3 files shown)


siefkenj (commented):
Most of this looks good. I know there are linting rules that can be installed to warn for interactive elements without accessibility features.

packages/test-cypress/cypress/e2e/tagSpecific/answer.cy.js
```diff
@@ -38,15 +38,23 @@ describe("Answer Tag Tests", function () {
             let mathInputPartialAnchor = cesc("#" + mathInputName + "_partial");
 
             // cy.get(mathInputAnchor).should('have.value', '');
-            cy.get(mathInputSubmitAnchor).should("be.visible");
+            cy.get(mathInputSubmitAnchor).should(
+                "have.attr",
+                "aria-label",
+                "Check Work",
+            );
```
siefkenj, line 45: You should be able to do this in a non-cypress test, but maybe it is easiest to put it here?

siefkenj (approved):

PR #623, opened 2025-10-08
add devcontainer support

StevenClontz:
This PR enables turnkey provisioning of an online browser-based development environment using [Codespaces](https://g4m.code4math.org/ch-coding.html#def-codespaces).


dqnykamp (changes requested):
Let's take the `test-viewer` out of the default setup procedure and instructions.

StevenClontz:
I think this is ready. Provisioning a fresh codespace to confirm.

PR #639, opened 2025-10-15
manage <aside> with keyboard controls

### LINKED ISSUE #616: Accessibility: `<aside>` cannot be opened with keyboard controls and screen reader
When navigating using keyboard controls and a screen reader, a `<hint>` can be entered and opened by clicking "Return". However, this does not work with an `<aside>`, the cursor jumps back out to the main article upon hitting "Return". 
See https://beta.doenet.org/activityViewer/vnYfvxJqrmtPHuNQmTFZ2B for an example

StevenClontz:
Closes #616

### DIFF
--- a/packages/doenetml/src/Viewer/renderers/section.tsx
+++ b/packages/doenetml/src/Viewer/renderers/section.tsx
@@ -380,6 +380,16 @@ export default React.memo(function Section(props) {
                         borderTopLeftRadius: "var(--mainBorderRadius)",
                         borderTopRightRadius: "var(--mainBorderRadius)",
                     }}
+                    tabIndex="0"
+                    onKeyPress={(e) => {
+                        if (e.key === "Enter") {
+                            callAction({
+                                action: SVs.open
+                                    ? actions.closeSection
+                                    : actions.revealSection,
+                            });
+                        }
+                    }}
                     onClick={() =>
                         callAction({
                             action: SVs.open


packages/doenetml/src/Viewer/renderers/section.tsx
```diff
@@ -380,6 +380,16 @@ export default React.memo(function Section(props) {
...
+                    tabIndex="0"
+                    onKeyPress={(e) => {
+                        if (e.key === "Enter") {
+                            callAction({
+                                action: SVs.open
+                                    ? actions.closeSection
+                                    : actions.revealSection,
+                            });
+                        }
+                    }}
```
StevenClontz: I think a better fix would be to implement this (and hints) as a button. Accessibility isn't exactly my expertise but I think if it's supposed to be interacted with, the `<button/>` element is the more sematic choice rather than adding functionality into a `div` or `aside` element.

StevenClontz: Looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details this is exactly what we should be doing. The `<summary>` is what we're calling the title, and the rest is the content that appears when we expand. But I would still prefer to make that a separate PR; we should probably refactor all the expandy things to use `<details>` and that would also require looking at CSS.

dqnykamp (approved):

dqnykamp:
I merged this PR, which closes the original issue. @virginiaMae , could you add another issue for the more general issue of accessibility and collapsible components? Then, we can have the discussion of appropriate aria roles and how to use `<details>` there or with an associated PR.

PR #643, opened 2025-10-20
script to find and download latest build artifact

### LINKED ISSUE #638: Host build cache online for downloading to new development environments
After a push to main, CI/CD should build the project and host the build cache online. New Codespaces can be configured to download this cache and use it for the initial setup process.

StevenClontz:
Closes #638

I feel like I've probably reinvented some wheel, but this script should grab and extract the latest build artifact if it exists.

Once #623 is merged, I'll update this to run on provisioning a new codespace before building things, which should really speed things up.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
             "version": "latest"
         }
     },
-    "postCreateCommand": "npm install && npm run build",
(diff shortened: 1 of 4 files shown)


StevenClontz:
CC @oscarlevin and @siwelwerd -- this pattern would allow us to use GitHub's built-in artifact system to download caches of PreTeXt-generated or CheckIt-generated assets for new codespaces/clones, without needing to upload them elsewhere as is done currently for TBIL.org

PR #659, opened 2025-10-31
remove overflow:hidden from doenetml-applet for pretext

StevenClontz:
Support scrolling when doenet is within a PTX document


dqnykamp:
They must have upgraded rust on the CI, as now rust lint if failing. Will make another PR to fix that.

Issue #652, opened 2025-10-26
improve accessibility of the `<graph>` component


virginiaMae:
Currently, there is no label that alerts the user that they are entering a graph region. If the description attribute is included, this is read off first. Then, the screen reader starts reading off the features of the graph. This includes the x-axis, the x-axis labels, followed by the y-axis, the y-axis labels, followed by features on the graph, followed by the navigation. All of the graphical features are simply read off as either "graphic symbol" or the number (such as an axis label or point coordinate), and there is no information that describes what the feature that is being read off refers to on the graph. Based on a comparison with similar graphical tools on Desmos Amplify classroom activities involving graphs, I recommend:

- Before the alt-text is read off, and before any features are read, the user should first hear "Doenet Interactive Graph. You are now on a Graph."

- Next, the alt-text should be read.

- Following this, there should be an option for the user to provide an interaction to enter the graph. (On VoiceOver for example this is Cntl-Option-Shift-Down Arrow)

- Once the user enters the graph, we need to apply labels to all features of the graph as they are read off. Some of these features should be suppressed if the feature cannot generally be interacted with. For example, instead of reading "Graphic Symbol" for the x-axis followed by listing off the numbers, we should instead have an algorithmic way of substituting logical language proscribed by the doenetML that describes the range. In Desmos, this sounds like "Graph paper. X axis visible from negative 5.9 to 5.9. Y axis visible from negative 0.9 to 10.9. Grid lines are visible...." (It ends with how to interact with an audio trace, which we don't have yet.)

-Another example of features that should have descriptive reading text are objects like points, functions, and curves. Currently, these are just read off as "Graphic Symbol". 

Should we make individual issues on all of these graphical elements other than the axes? 

- Labels of graphical objects appear to be ignored by the screen reader

-Finally, the graph navigation is read off as "selectable text" as the last of the features that are read. It is read as "end dash, o (oh), plus"

[...]

PR #689, opened 2025-11-17
updates to graph renderer for issue #652

virginiaMae:
correcting screen reader performance. 

- removed the "title" attribute from the Board object for JSX graph
- created a div around the board that has aria label of the description state variable supplied by user from the graph component
- added a role of image to the new container div, so that screen readers ignore the JSX graph board objecgt


dqnykamp (approved):
Interesting how `role="img"` was the trick for keeping screen readers from going into the graph.

PR #888, opened 2026-02-28
stream additional diagnostics into CodeMirror

### LINKED ISSUE #885: Warning should propagate to LSP
Right now, only warning from the initial parsing show up in the LSP and create squiggly lines. Warnings generated from within Doenet core do show up in the lower left corner. But, those are easy to ignore. Warnings from Doenet core (e.g., accessibility warning) need to result in squiggly lines in the editor.

dqnykamp:
This PR fixes the additional-diagnostics flow so that warnings and errors generated by the viewer appear in the editor lint panel without replacing existing parser/schema diagnostics.

## Changes

- Migrated diagnostics delivery in the editor to a push/subscription model.
- Added worker diagnostics subscriber support and plugin lifecycle unsubscribe.
- Removed obsolete polling-based diagnostics code.
- Added a normalization/validation of ranges 
- Added lsp unit test
- Added cypress component tests of the diagnostics of `<CodeMirror>`.

Fixes #885

### DIFF
--- a/packages/codemirror/src/extensions/lsp/plugin.ts
+++ b/packages/codemirror/src/extensions/lsp/plugin.ts
@@ -52,6 +52,10 @@ import {
     CompletionContext,
     Completion,
 } from "@codemirror/autocomplete";
+import {
+    getSnippetCursorFromCompletionItemData,
+    type CompletionSnippetCursor,
+} from "@doenet/static-assets/completion-snippet-protocol";
 import type {
     MarkupContent,
     MarkedString,
@@ -92,6 +96,7 @@ type ExtendedCompletion = Completion & {
         start: { line: number; character: number };
         end: { line: number; character: number };
     };
+    _snippetCursor?: CompletionSnippetCursor;
 };
 
 // One language server is shared across all plugin instances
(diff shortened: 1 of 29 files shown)


packages/codemirror/src/extensions/lsp/worker.ts
```diff
@@ -47,6 +64,18 @@ export class LSP {
         }
     }
 
+    onDiagnostics(
+        callback: (params: { uri: string; diagnostics: Diagnostic[] }) => void,
+    ) {
+        this.diagnosticsSubscribers.add(callback);
+        void this.init().then(() => {
```
siefkenj: I really don't like this style of adding uneeded `void`s in places.

dqnykamp: replaced with explicit error handler

packages/codemirror/src/extensions/lsp/worker.ts
```diff
@@ -47,6 +64,18 @@ export class LSP {
...
+            this.ensureDiagnosticsHandlerRegistered();
+        });
+        return () => {
```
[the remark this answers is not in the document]
dqnykamp: refactored with auto-cleanup

PR #959, opened 2026-03-23
Fill out AGENTS.md with repo structure

cqnykamp:
This is what Copilot's `/init` command generated (what it thinks should be in AGENTS.md). I kept the existing info too.


dqnykamp:
Ended up basing it off an /init command from claude

PR #1126, opened 2026-05-13
Add Pretext and Doenet skills and doenet-to-pretext functions

siefkenj:
This PR adds skills for AIs writing doenet and pretext.

It also does some minor work on the doenet-to-pretext conversion and removes some warnings when initializing wasm blobs.

PR #1128, opened 2026-05-13
Pretext Conversion Updates

siefkenj:
This PR depends on #1126 

Adds more pretext export functionality.


packages/doenetml-to-pretext/scripts/create-renderer-stub.ts
```diff
@@ -0,0 +1,201 @@
...
+            console.error(
+                "Invalid value. Please enter 'children' or 'nochildren'.",
+            );
+        }
+    } finally {
+        rl.close();
+    }
+}
+
+async function main() {
```
dqnykamp, line 66: Docstring here or at top of file explaining the purpose of this function.

packages/doenetml-to-pretext/src/renderers/pretext-xml/answer.tsx
```diff
@@ -1,19 +1,10 @@
...
-                Submit
-            </button>
-        </span>
-    );
+type AnswerData = { props: unknown };
+
+export const Answer: BasicComponentWithPassthroughChildren<AnswerData> = ({
+    children,
+}) => {
+    return children;
```
dqnykamp, line 9: Is there a reason why this does not have to be wrapped in a `React.Fragment`? It seems like most of these functions, including `_PassThroughWithoutTag` always return a single React element. Wondering why this one differs.

packages/doenetml-to-pretext/src/renderers/pretext-xml/choice-input.tsx
```diff
@@ -1,6 +1,88 @@
...
+        label,
+        inline,
+    } = node.data.props;
+    const childrenArray: React.ReactNode[] = Array.isArray(children)
+        ? children
+        : [children];
+
+    // If the choiceInput is inline, then the selected choice is shown without any icon.
+    // If there is no selected choice, a fillin is shown instead.
+    if (inline) {
```
dqnykamp, line 33: We do allow inline choice inputs with select multiple. The below would just show the first one that is selected.

packages/doenetml-to-pretext/src/renderers/pretext-xml/abs.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { BasicComponent } from "../types";
+
+type AbsData = { props: { latex: string } };
+
+export const Abs: BasicComponent<AbsData> = ({ node }) => {
+    return <m>{node.data.props.latex}</m>;
```
dqnykamp, line 7: What's the trick so that react knows what `<m>` is?

packages/doenetml-to-pretext/src/renderers/pretext-xml/code-editor.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import { BasicComponent } from "../types";
+
+type CodeEditorData = { props: { value: string } };
+
+export const CodeEditor: BasicComponent<CodeEditorData> = ({ node }) => {
```
dqnykamp, line 6: This is a reasonable first effort. I've been pondering if there is a way, when `showResults` is `true`, to process the nodes from the code editor inside the same Core but somehow create an isolated resolver so that the name resolution is sandboxed. Seems hard to justify the effort and additional complexity, though.

packages/doenetml-to-pretext/src/renderers/pretext-xml/li.tsx
```diff
@@ -8,10 +8,6 @@ export const Li: BasicComponentWithPassthroughChildren<{
 }> = ({ children, node, annotation, ancestors }) => {
     const htmlId = generateHtmlId(node, annotation, ancestors);
     const label = node.data.props.label;
-    return (
-        <li id={htmlId}>
-            <span className="list-label">{label}</span>
-            {children}
-        </li>
-    );
+    console.log("li rendering", node, ancestors);
```
dqnykamp: remove console.log

packages/doenetml-to-pretext/test/pretext-export.test.ts
```diff
@@ -217,4 +357,17 @@ describe("Pretext export", async () => {
...
     });
+
+    it("renders a graph with a point", async () => {
+        source = `<graph><point name="P" x="1" y="2" /></graph>`;
+        expect(await coreRunner.processToFlatDast(source))
+            .toMatchInlineSnapshot(`
+              "<?xml version="1.0" encoding="UTF-8"?>
+              <pretext>
+              <article>
+              <image><prefigure label="prefigure-doenet-id-1" xmlns="https://prefigure.org"><diagram dimensions="(425,425)"><coordinates bbox="(-10,-10,10,10)"><axes axes="all"></axes><point at="point_0" p="(1,2)" style="circle" size="5" fill="#648FFF" stroke="#648FFF" fill-opacity="0.7" stroke-opacity="0.7" thickness="4"></point></coordinates><annotations></annotations></diagram></prefigure></image>
```
dqnykamp, line 381: I remembered why DoenetML always produces an `<annotations>` even if there aren't any. The PreFigure wasm automatically generates annotations if there aren't any, and I want to disable those.

dqnykamp (changes requested):
What is the plan for publishing `doenetml-to-pretext`? Will it be a separate npm module? Just thinking about change sets and this package. I've been trying to create change sets for publicly visible changes. You're currently off the hook since this isn't published. :) But, eventually, we should determine a strategy for publishing and change sets.

siefkenj:
I'm not sure if there is a plan to publish, since I don't think this package works in node (because of the WebWorker, I will have to test...).

I am working on a python package ATM. I assume that will be what is published.

packages/doenetml-to-pretext/src/renderers/pretext-xml/answer.tsx
siefkenj: Not a particular reason.

packages/doenetml-to-pretext/src/renderers/pretext-xml/abs.tsx
siefkenj: The `pretext.d.ts` defines all the pretext elements as valid JSX

packages/doenetml-to-pretext/test/pretext-export.test.ts
siefkenj: Ah. Maybe the answer is to fix prefigure then, so you don't have to include them :-)

packages/doenetml-to-pretext/src/renderers/pretext-xml/code-editor.tsx
siefkenj: Yeah. I say we wait till a user requests it.

dqnykamp (approved):
Looks good. We can merge after running prettier.

PR #1129, opened 2026-05-14
refactor(lsp): load rust core via Comlink sub-worker instead of bundled WASM

dqnykamp:
## Summary

The LSP previously imported `@doenet/doenetml-worker-rust`'s WASM directly, shipping a duplicate copy of the Rust core inside the LSP bundle alongside the inlined core webworker that already embeds it.

This PR plumbs `<CodeMirror doenetWorkerUrl=...>` through the LSP `initialize` request's `initializationOptions`. `getRustCore(url)` now spawns a fresh Comlink-wrapped sub-worker per document — reusing the existing core worker bundle — and the LSP no longer ships its own copy of the WASM.

- `<CodeMirror>` prop `doenetWorkerUrl` is now optional. When omitted, ref/member completions are silently disabled (the existing `rustState: "unavailable"` path).
- `RustResolverCore` interface and `RustResolverAdapter` become async (Comlink calls return Promises), which propagates to `AutoCompleter.isNameAddressable` / `resolveRefMemberContainerAtOffset` / `getCompletionItems` and the LSP `connection.onCompletion` handler.
- `CoreWorker` (in `@doenet/doenetml-worker`) gains an async `resolvePath` method matching the gating pattern of `setSource` / `setFlags`.
- Per-document worker lifecycle: `getRustCore` returns `{ core, terminate }`; `validate.ts` stores the terminator and calls it on `documents.onDidClose`. The close-during-init race (worker spawned after the document was closed) is also handled.

## Bundle size impact

Same ~6.1 MiB drops out of every downstream bundle that pulled in the LSP — that's the duplicated Rust WASM (base64-encoded into the LSP IIFE) being removed. The inline-worker bundle is unchanged because it has always been the *single legitimate* copy of the core worker.

| file | before | after | delta |
|---|---:|---:|---:|
| `packages/lsp/dist/index.js` | 6.79 MiB | 713 KiB | **−6.10 MiB (−89.7%)** |
| `packages/codemirror/dist/index.js` | 8.61 MiB | 2.51 MiB | **−6.10 MiB (−70.9%)** |
| `packages/doenetml/dist/doenetml-*.js` | 15.17 MiB | 9.06 MiB | **−6.10 MiB (−40.2%)** |
| `packages/doenetml/dist/doenetml-inline-worker.js` | 10.51 MiB | 10.51 MiB | ~0 |
| `packages/standalone/dist/doenet-standalone.js` | 30.56 MiB | 24.46 MiB | **−6.10 MiB (−20.0%)** |

## Singleton URL caveat

The LSP is a process-wide singleton, so the first `<CodeMirror>` instance to mount locks in the URL. Later instances passing a different URL get a `console.warn` and have their value ignored. In practice every editor on a page reads from the same `doenetGlobalConfig`, so this is rarely an issue. The contract is documented on the `doenetWorkerUrl` prop's JSDoc.

## Follow-up: missing cross-document isolation test

`packages/lsp/test/language-server.test.ts` has one `it.skip` for cross-document isolation. `@vitest/web-worker` doesn't implement the worker-global `addEventListener` that `Comlink.expose` needs inside the spawned core sub-worker, so the test can no longer run in vitest. The contract it was checking (per-document Rust state isolation) is now structurally provided by `getRustCore` returning a fresh sub-worker per document — but explicit coverage is gone. **Worth adding a Cypress spec** that opens two docs in the same editor session and verifies ref-completions in doc A don't see names from doc B.

## Test plan

- [x] `npx vitest run --root packages/lsp` — 9 passed, 1 skipped (the cross-doc isolation test noted above)
- [x] `npx vitest run --root packages/lsp-tools` — 192 passed (including `rust-resolver-integration` 38 tests against real WASM)
- [x] `npx tsc --noEmit` clean for `lsp`, `lsp-tools`, `codemirror`, `doenetml-worker`
- [x] `npm run build -w packages/lsp -w packages/codemirror -w packages/doenetml-worker -w packages/lsp-tools -w packages/doenetml -w packages/standalone` succeeds
- [ ] Manual smoke in the editor: open a doc with `<point name="P"/> $P.|`, confirm member completions; confirm `$P[0].|` for a `takesIndex` element; confirm `$nope.x` shows none
- [ ] DevTools sanity check: LSP web worker no longer fetches/decodes the Rust WASM bytes


packages/doenetml-worker/src/CoreWorker.ts
```diff
@@ -489,6 +489,36 @@ export class CoreWorker {
         }
     }
 
+    async resolvePath(args: {
```
siefkenj, line 505: docstring

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
```diff
@@ -171,37 +171,38 @@ function getElementNameAttributeValue(el: DastElement): string | null {
 
 /**
  * Minimal interface matching the subset of PublicDoenetMLCore methods needed
- * for path resolution. Consumers provide an initialized WASM-backed instance;
- * lsp-tools does not depend on the WASM package directly.
+ * for path resolution. Consumers provide an implementation that may be backed
+ * by an in-process WASM module or a Comlink proxy to a worker.  All methods
+ * are async because the Comlink-proxied implementation returns Promises.
```
siefkenj: Is this wrapper even needed now that we are instantiating a full core? Consider deleting.

packages/lsp-tools/src/dev-site.tsx
```diff
@@ -79,16 +79,20 @@ function App() {
...
-            { currentPos },
-            sourceObj.elementAtOffsetWithContext(currentPos),
-            "elm2 left",
-            sourceObj.nodeAtOffset(currentPos, { side: "left" })?.type || null,
-            "elm2 right",
-            sourceObj.nodeAtOffset(currentPos, { side: "right" })?.type || null,
-            sourceObj.attributeAtOffset(currentPos),
-            completionObj.getCompletionItems(currentPos),
-        );
+        completionObj.getCompletionItems(currentPos).then((items) => {
```
siefkenj: Use async function with await in the args of console.log instead.

packages/lsp/src/features/validate.ts
```diff
@@ -88,27 +88,41 @@ export function addValidationSupport(
 
         if (info.rustState === "ready" && info.rustAdapter) {
             // Adapter already wired — just resync source.
-            info.rustAdapter.updateSource(info.autoCompleter.sourceObj);
+            info.rustAdapter
+                .updateSource(info.autoCompleter.sourceObj)
+                .catch(() => undefined);
```
siefkenj: This `catch` makes no sense.

siefkenj: Let's make the `onDidChangeContent` function async so we can await inside of it.

packages/lsp/src/features/validate.ts
```diff
@@ -88,27 +88,41 @@ export function addValidationSupport(
...
-                    // document. This avoids cross-document source switching
-                    // complexity and keeps Rust state aligned with this
-                    // document's AutoCompleter mappings.
+                    // Intentionally create a dedicated adapter per document.
+                    // This avoids cross-document source switching complexity
+                    // and keeps Rust-side state aligned with this document's
+                    // AutoCompleter mappings.
                     const adapter = new RustResolverAdapter(sourceObj, {
-                        core: core as RustResolverCore,
+                        core: spawned.core as RustResolverCore,
```
siefkenj: This makes me unahppy. If each document gets its own instance of core, then fine. We don't need an additional layer of abstraction. Store the core instance on the document's global data structure.

packages/lsp/src/rust-core.ts
```diff
@@ -1,152 +1,61 @@
...
+    doenetWorkerUrl: string | undefined,
+): Promise<SpawnedRustCore | null> {
+    if (!doenetWorkerUrl) {
+        return null;
+    }
+
+    const worker = new Worker(doenetWorkerUrl);
+    const remote = Comlink.wrap<CoreWorker>(worker);
+
+    const core: RustResolverCore = {
```
siefkenj: No need for the RustResolverCore any more. Just expose the real core.

packages/lsp/test/utils/init-message-connection.ts
```diff
@@ -4,8 +4,11 @@ import { createMessageConnection } from "@qualified/vscode-jsonrpc-ww";
 /**
  * Initialize a WebWorker that runs a language server. The worker is initialized with
  * `rootUri` set to `file:///` and `workspaceFolders` set to `null`.
+ *
+ * `doenetWorkerUrl`, when provided, is forwarded via `initializationOptions`
+ * so the language server can spawn a rust-core sub-worker for path resolution.
  */
-export async function initWorker(worker: Worker) {
+export async function initWorker(worker: Worker, doenetWorkerUrl?: string) {
```
siefkenj, line 19: Why do we need the extra argument here? Where is `worker` coming from in the first place? Shouldn't `worker` be the thing with `doenetWorkerUrl`?

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
```diff
@@ -246,6 +246,23 @@ export class RustResolverAdapter {
...
     readonly _visibleDescendantNamesCache: Map<string, string[]> = new Map();
 
+    /**
+     * Tail of the in-flight/queued sync chain.  `init` and `updateSource`
+     * append a fresh `_syncSource` to this; query methods `await` it before
+     * reading state.  Without this, a completion request can race a
+     * fire-and-forget `updateSource` and resolve against stale rust source
+     * or stale JS-side index mappings.
+     */
+    _isSyncing: Promise<void> = Promise.resolve();
```
siefkenj: Give this a better name and make a comment that is understandable and less abstract. Don't use "in-flight" jargon.

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
```diff
@@ -246,6 +246,23 @@ export class RustResolverAdapter {
...
+
+    /**
+     * Source text the most recently *executed* `_syncSource` decided to push
+     * (or accept as already-current).  Used to dedup rapid-fire syncs:
+     * fast typing chains many calls but only one actually pays the worker
+     * round-trips; subsequent dequeued syncs see the same source and skip.
+     */
+    _lastSyncedSource: string | null = null;
```
siefkenj: What is with this `synced` language? Is this a standin for `_currentSource`? Is there another source being usd at the same time? Should it not be called `doenetML`?

packages/doenetml-worker/src/CoreWorker.ts
dqnykamp: Added in 60a09d986. While there I also added the `source_set`/`flags_set` guard — `resolvePath` was lazily constructing an empty `PublicDoenetMLCore` when called before `setSource`, so it now throws the same clear error as its siblings `dispatchAction`/`returnDast` instead of resolving against an empty core.

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
dqnykamp: Addressed in 79fa2b450. The snake_case translation shim is gone — the adapter now calls the core's real method names directly.

I kept a minimal interface (renamed `RustResolverCore` → `ResolverCore`) rather than importing `CoreWorker` directly. The adapter's options type is public API, so importing `CoreWorker` would leak that type into `@doenet/lsp-tools`' emitted `.d.ts` and force lsp-tools to build-depend on the heavy `@doenet/doenetml-worker` package. `ResolverCore` is now an honest port — the structural subset of the core API the adapter calls — which the real Comlink-proxied `CoreWorker` satisfies. Happy to delete it outright and take the build-graph dependency if you'd rather.

packages/lsp-tools/src/dev-site.tsx
dqnykamp: Done in 0ef57736f — replaced the `.then()` with an async IIFE that awaits `getCompletionItems` directly in the `console.log` argument list.

packages/lsp/src/features/validate.ts
dqnykamp: Agreed — removed in 9e4c17815. `updateSource` chains its syncs internally and never rejects (the adapter logs and disables itself on a sync failure), so the catch was unreachable dead code.

dqnykamp: Done in 9e4c17815. `onDidChangeContent` is now async: it computes diagnostics first (they derive from the JS-side DAST and must not wait on the rust worker) and then awaits the rust adapter sync. The rust-core bootstrap stays a `void`-marked fire-and-forget IIFE, since diagnostics must not block on the worker spinning up.

packages/lsp/src/features/validate.ts
dqnykamp: Done in 79fa2b450. The extra layer (the snake_case shim) is gone, and the core is now stored directly on the per-document data structure as `documentInfo[uri].rustCore`.

packages/lsp/src/rust-core.ts
dqnykamp: Done in 79fa2b450 — `getRustCore` now returns the real Comlink proxy directly (with one boundary cast to drop Comlink's pessimistic `RemoteObject` return typing); the translation object is gone. See the reply on `rust-resolver-adapter.ts` for why a minimal `ResolverCore` port type still remains in lsp-tools.

packages/lsp/test/utils/init-message-connection.ts
dqnykamp: `worker` and `doenetWorkerUrl` are two genuinely different workers. `worker` is the **language-server** worker — the caller constructs it from the LSP bundle. `doenetWorkerUrl` is the URL of a **separate DoenetML core** worker bundle.

The language server runs *inside* `worker` and needs that URL so it can spawn its own core sub-worker for rust-backed path resolution. The LSP-standard channel for host→server configuration is the `initialize` request's `initializationOptions`, which is exactly what `initWorker` sends — so the URL can't be baked into `worker`; it has to be passed here to be forwarded through the handshake. I clarified the docstring in fcb294465 to spell this out.

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
dqnykamp: Done in 0716a2e12 — renamed to `_pendingSync` (it is a promise, not a boolean) and rewrote the comment in concrete terms without the "in-flight" jargon.

packages/lsp-tools/src/auto-completer/rust-resolver-adapter.ts
dqnykamp: Done in 0716a2e12 — renamed to `_coreSource`: the DoenetML source string the rust core currently holds. It is distinct from `_sourceObj.source` (the latest source the adapter knows); the two differ while a sync is pending, and `_syncSource` compares them to skip redundant syncs. Comment rewritten to say that plainly.

Issue #1139, opened 2026-05-19
DoenetViewer/DoenetEditor use `Math.random()` for the iframe id, breaking SSR hydration


dqnykamp:
## Summary

`DoenetViewer` and `DoenetEditor` in `@doenet/doenetml-iframe` generate a per-instance id with `Math.random()`:

- `packages/doenetml-iframe/src/index.tsx:116` (`DoenetViewer`)
- `packages/doenetml-iframe/src/index.tsx:345` (`DoenetEditor`)

```ts
const [id, _] = React.useState(() => Math.random().toString(36).slice(2));
```

That id is baked into the iframe's `srcDoc`. Under server-side rendering the server and client pick **different** random ids, so React reports a hydration mismatch on the `srcDoc` attribute ("some attributes of the server rendered HTML didn't match the client properties"). The server-rendered iframe keeps id A while the client component listens for `postMessage`s tagged id B — so the viewer never wires up and the example renders blank until a full client-side re-render happens.

## Impact

Any server-rendered host (Next.js, etc.) that embeds these components hits a hydration mismatch on first page load / reload.

This was worked around in `@doenet/docs-nextra` by gating the components behind a client-only mount (`useClientOnly` in `packages/docs-nextra/components/doenet.tsx`) so they render nothing during SSR and mount client-side. That fixes the symptom for the docs site, but the component itself remains SSR-unsafe for other consumers, and the docs examples lose SSR prerendering as a result.

## Suggested fix

Replace `React.useState(() => Math.random().toString(36).slice(2))` with `React.useId()`, which is stable across server and client renders. The `docs-nextra` `useClientOnly` workaround could then be removed.

**Caveat:** `useId()` returns strings containing `:` (e.g. `:r0:`). The id is interpolated into generated HTML/JS as a string literal and used for `postMessage` routing — verify it is not used anywhere as a CSS selector or with `querySelector('#' + id)` (colons must be escaped there; `getElementById` is fine). Sanitize the `useId()` output if colons turn out to be a problem.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1138, opened 2026-05-19
DoenetML worker blob URL triggers spurious `/index.js.map` 404s


dqnykamp:
## Summary

Loading any page that embeds a `DoenetViewer`/`DoenetEditor` produces repeated `GET /index.js.map 404` requests in the browser / dev-server console — one (or more) per embedded document. The 404s are harmless (the worker runs fine without the source map) but are console noise.

## Root cause

- `packages/doenetml-worker/vite.config.ts` builds the worker with `sourcemap: true`, so the built `index.js` ends with `//# sourceMappingURL=index.js.map`.
- `packages/doenetml/src/doenetml-inline-worker.ts` imports that worker as a raw string (`@doenet/doenetml-worker/index.js?raw`) and runs it via `URL.createObjectURL(...)` — a `blob:` URL.
- A worker started from a `blob:` URL has no path component, so the browser resolves the relative `index.js.map` against the page origin → `http://<host>/index.js.map` → 404.

The comment string `//# sourceMappingURL=index.js.map` is embedded verbatim inside the bundled `doenet-standalone.js`, so this happens with both the CDN-published and locally-built standalone.

## Suggested fixes (options)

1. **Preferred:** strip the trailing `//# sourceMappingURL=...` line from `workerSource` in `doenetml-inline-worker.ts` before creating the blob. Keeps the on-disk worker source map for anyone debugging the worker directly; just drops the dangling reference from the blob.
2. `sourcemap: "inline"` in `packages/doenetml-worker/vite.config.ts` — keeps worker debuggability, no separate file to 404 on, but enlarges the bundle.
3. `sourcemap: false` for the worker build — simplest, but loses worker source maps.

## Context

Noticed while wiring the `docs-nextra` dev server to the locally-built standalone bundle (branch `docs-nextra-local-standalone`). Pre-existing and unrelated to that change — filed as a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1140, opened 2026-05-19
Load local standalone in docs-nextra dev server; fix iframe React duplication

dqnykamp:
## Motivation

The `docs-nextra` package loaded `DoenetEditor`/`DoenetViewer` with `doenetmlVersion="dev"`, which resolves to the standalone build published off `main` via CDN. That is correct for the published docs (built in CI off `main`), but in local development it means doc examples cannot exercise a feature that is new in the current PR — which pushes contributors to write docs in a separate follow-up PR rather than the preferred single feature-plus-docs PR.

This PR makes the docs-nextra **dev server** point at the locally-built standalone, and fixes two bugs uncovered along the way.

## Changes

**1. Dev server loads the local standalone**
- `components/doenet.tsx`: in `next dev`, `DoenetEditor`/`DoenetViewer` now load `/bundle/doenet-standalone.js` + `/bundle/style.css` — the locally-built standalone that the existing `build:pre-copy` task already copies into `public/bundle/`. `next build` (published docs) keeps using the CDN `dev` version, unchanged.
- Version autodetection in `@doenet/doenetml-iframe` is left at its default (`true`). A doc example whose source contains an explicit `<document xmlns="…">` version tag therefore still resolves that version from the CDN rather than the local bundle; examples without a version tag (the common case) load the local bundle.

**2. Fix `dispatcher.getOwner is not a function`**
- `@doenet/doenetml-iframe`'s library build only externalized the exact strings `react` and `react-dom`, so the subpaths (`react/jsx-runtime`, `react-dom/client`, …) and `better-react-mathjax` were bundled into `dist/component/index.js`, creating a duplicate React instance that conflicted with the host's React.
- `doenetml-iframe/vite.config.ts` now externalizes those subpaths and `better-react-mathjax` via a predicate; `better-react-mathjax` is added as a peer dependency.
- `docs-nextra/next.config.mjs` adds a `resolve.alias` React-dedup as a belt-and-suspenders guarantee.

**3. Render the iframe components client-only in docs-nextra**
- Fixing `getOwner` exposed a latent SSR bug: the iframe components use a `Math.random()` id baked into the iframe `srcDoc`, so server and client disagree and hydration mismatches — leaving examples blank after a page reload. `components/doenet.tsx` now loads `DoenetEditor`/`DoenetViewer` via `next/dynamic` with `ssr: false`.
- `next/dynamic` cannot import `@doenet/doenetml-iframe` directly — the package's `exports` map only declares the `import` condition, so a dynamic `import()` of it fails webpack's exports-map resolution (`Package path . is not exported`). The dynamic import therefore targets a new thin re-export module, `components/doenet-iframe.tsx`, which does the (working) static package import.
- The root cause in `@doenet/doenetml-iframe` is tracked in #1139.

## Notes
- Doc examples are now client-rendered rather than SSR-prerendered (in both dev and the published build). This is necessary — the `Math.random()` hydration mismatch would otherwise break the published docs once `getOwner` is fixed.
- Follow-ups filed: #1138 (worker source-map 404s) and #1139 (make `doenetml-iframe` SSR-safe with `useId()`).

## Testing
- `@doenet/doenetml-iframe` library build: `dist/component/index.js` now imports `better-react-mathjax` externally (no inlined MathJax); the published `package.json` lists the new peer dependency.
- `docs-nextra` production build (`next build`) succeeds; prod export uses the CDN `dev` version, with no `/bundle/` reference.
- `docs-nextra` dev server serves `/bundle/doenet-standalone.js`; the rendered pages reference the local bundle, not the CDN.
- Verified in-browser: a page with `DoenetEditor`/`DoenetViewer` loads without the `getOwner` error and is no longer blank on reload.

### DIFF
--- a/packages/docs-nextra/components/doenet.tsx
+++ b/packages/docs-nextra/components/doenet.tsx
@@ -6,12 +6,52 @@ import {
 import "@doenet/virtual-keyboard/style.css";
 
 /**
+ * Which standalone DoenetML build the embedded examples load.
+ *
+ * In development (`next dev`), load the locally-built standalone bundle that the
+ * `build:pre-copy` task copies into `public/bundle/`. This lets doc examples exercise
+ * features that are new in the current PR, instead of the version published off `main`.
+ *
+ * In a production build (`next build`, used for the published docs), keep using the
+ * `dev` version from the CDN.
+ */
+const versionProps =
+    process.env.NODE_ENV === "development"
+        ? {
+              standaloneUrl: "/bundle/doenet-standalone.js",
+              cssUrl: "/bundle/style.css",
+              // Force the local bundle even if an example contains a
+              // <document xmlns=".../vX"> tag that would otherwise be auto-detected.
+              autodetectVersion: false as const,
+          }
+        : { doenetmlVersion: "dev" };
+
+/**
+ * `true` only after the component has mounted on the client.
+ *
+ * The iframe components generate a random per-instance id that is baked into the
+ * iframe's `srcDoc`, so they cannot be server-rendered without a hydration mismatch
+ * (server and client pick different ids), which leaves the example blank after a page
+ * reload. Gating on this renders nothing during SSR and the initial hydration pass,
+ * then mounts the real component client-side.
+ */
+function useClientOnly() {
+    const [mounted, setMounted] = React.useState(false);
+    React.useEffect(() => setMounted(true), []);
+    return mounted;
+}
+
+/**
  * Render DoenetML in an iframe so that its styling/state is completely isolated from the page.
  */
 export function DoenetViewer({
     source,
 }: React.PropsWithChildren<{ source: string }>) {
-    return <DoenetViewerOrig doenetML={source} doenetmlVersion="dev" />;
+    const mounted = useClientOnly();
+    if (!mounted) {
+        return null;
+    }
+    return <DoenetViewerOrig doenetML={source} {...versionProps} />;
 }
 
 export function DoenetEditor({
@@ -25,15 +65,18 @@ export function DoenetEditor({
     viewerLocation?: "left" | "right" | "top" | "bottom";
     height?: string;
 }>) {
+    const mounted = useClientOnly();
     return (
         <div className="doenet-editor-container">
-            <DoenetEditorOrig
-                doenetML={source}
-                showFormatter={showFormatter}
-                viewerLocation={viewerLocation}
-                height={height}
-                doenetmlVersion="dev"
-            />
+            {mounted ? (
+                <DoenetEditorOrig
+                    doenetML={source}
+                    showFormatter={showFormatter}
+                    viewerLocation={viewerLocation}
+                    height={height}
+                    {...versionProps}
+                />
+            ) : null}
         </div>
     );
 }
--- a/packages/docs-nextra/components/doenet.tsx
+++ b/packages/docs-nextra/components/doenet.tsx
@@ -20,9 +20,6 @@ const versionProps =
         ? {
               standaloneUrl: "/bundle/doenet-standalone.js",
               cssUrl: "/bundle/style.css",
-              // Force the local bundle even if an example contains a
-              // <document xmlns=".../vX"> tag that would otherwise be auto-detected.
-              autodetectVersion: false as const,
           }
         : { doenetmlVersion: "dev" };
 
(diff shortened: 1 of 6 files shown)


packages/docs-nextra/components/doenet.tsx
```diff
@@ -5,13 +5,53 @@ import {
...
+/**
+ * `true` only after the component has mounted on the client.
+ *
+ * The iframe components generate a random per-instance id that is baked into the
+ * iframe's `srcDoc`, so they cannot be server-rendered without a hydration mismatch
+ * (server and client pick different ids), which leaves the example blank after a page
+ * reload. Gating on this renders nothing during SSR and the initial hydration pass,
+ * then mounts the real component client-side.
+ */
+function useClientOnly() {
```
siefkenj: There should be a built-in nextra/nextjs directive to force a component to be client-only. You don't need to create your own.

packages/docs-nextra/next.config.mjs
```diff
@@ -8,6 +8,15 @@ import {
 } from "./dist/index.js";
 import { getHighlighter, bundledLanguages, bundledThemes } from "shiki";
 import fs from "node:fs";
+import path from "node:path";
+import { createRequire } from "node:module";
+
+const require = createRequire(import.meta.url);
+
+/** Resolve the directory of an installed package (so webpack aliases handle subpaths). */
```
siefkenj: Keep `/**` on its own line.

siefkenj: I don't understand the comment in parenthesis.

packages/doenetml-iframe/vite.config.ts
```diff
@@ -5,8 +5,19 @@ import { createPackageJsonTransformer } from "../../scripts/transform-package-js
...
-// These are the dependencies that will not be bundled into the library.
-const EXTERNAL_DEPS = ["react", "react-dom"];
+// Bare package names that will not be bundled into the library. These become
+// peerDependencies in the published dist/component/package.json.
+const EXTERNAL_PACKAGES = ["react", "react-dom", "better-react-mathjax"];
+
+// Rollup `external` predicate. Unlike a plain string array, this also matches
+// subpath imports (e.g. react/jsx-runtime, react/jsx-dev-runtime, react-dom/client),
+// which must be externalized too so the consuming app provides a single React
+// instance. Bundling them causes a duplicate-React "dispatcher" mismatch.
```
siefkenj: Use docstring comments `/**`

packages/docs-nextra/components/doenet.tsx
dqnykamp: Good call — done in b7a435d94. Replaced the hand-rolled `useClientOnly` hook with `next/dynamic(..., { ssr: false })`.

One wrinkle worth recording: `dynamic()` can't import `@doenet/doenetml-iframe` directly. `next/dynamic(() => import("@doenet/doenetml-iframe"))` fails webpack's exports-map resolution with `Package path . is not exported` (the subpath form `.../index.js` fails the same way), even though the plain static `import` resolves fine — the package's `exports` map only declares the `"import"` condition. So the dynamic import targets a new local re-export module, `components/doenet-iframe.tsx`, which does the (working) static package import; keeping it in its own file also means it's only pulled into the lazily-loaded client chunk.

Note `"use client"` alone wouldn't fix the underlying bug here — `"use client"` components are still server-rendered and hydrated; only `dynamic(..., { ssr: false })` actually skips SSR, which is what's needed to avoid the `Math.random()` id hydration mismatch.

packages/docs-nextra/next.config.mjs
dqnykamp: Done in e05b7de32. Expanded it to a multi-line block and clarified the parenthetical: `packageDir` returns a package's *directory* (not a specific file) because the `resolve.alias` entries below alias each package to that directory, which is what lets subpath imports like `react/jsx-runtime` and `react-dom/client` keep resolving against the aliased copy.

packages/doenetml-iframe/vite.config.ts
dqnykamp: Done in e05b7de32 — converted both `//` comments to `/**` docstring comments.

packages/docs-nextra/components/doenet.tsx
```diff
@@ -1,17 +1,48 @@
...
+ * In a production build (`next build`, used for the published docs), keep using the
+ * `dev` version from the CDN.
+ */
+const versionProps =
+    process.env.NODE_ENV === "development"
+        ? {
+              standaloneUrl: "/bundle/doenet-standalone.js",
+              cssUrl: "/bundle/style.css",
+          }
+        : { doenetmlVersion: "dev" };
```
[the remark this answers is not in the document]
dqnykamp: updated PR description

PR #1146, opened 2026-05-19
Improved DoenetML to PreTeXt

siefkenj:
Improves the DoenetML to PreTeXt functions to allow for embeddable fragments of pretext code.

### DIFF
--- a/packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
+++ b/packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
@@ -28,7 +28,6 @@ class DoenetConverter:
         # 1. Try to import from a Python Deno package (if present)
         try:
             import deno
-
             return deno.find_deno_bin()
         except ImportError:
             pass
--- a/packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
+++ b/packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
@@ -6,7 +6,7 @@ import subprocess
 from pathlib import Path
 from typing import List, Optional
 
-from .exceptions import DenoError, ConversionError, DoenetTimeoutError as DoenetTimeoutError
+from .exceptions import ConversionError, DenoError, DoenetTimeoutError as DoenetTimeoutError
 
 _PACKAGE_DIR = Path(__file__).resolve().parent
 _DIST_INDEX = _PACKAGE_DIR / "js-assets" / "index.js"
@@ -28,6 +28,7 @@ class DoenetConverter:
         # 1. Try to import from a Python Deno package (if present)
         try:
             import deno
+
             return deno.find_deno_bin()
         except ImportError:
             pass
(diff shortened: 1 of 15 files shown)


packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
```diff
@@ -49,13 +49,27 @@ def _resolve_dist_index(self) -> Path:
...
+            "  let output;\n"
+            "  if (input.length === 1) {\n"
+                        "    output = [await doenetMLToPretext(input[0], { fragment: false })];\n"
+            "  } else {\n"
+            "    const converter = new DoenetMLToPretext();\n"
+            "    if (typeof converter.convertMultiple === 'function') {\n"
+            "      output = await converter.convertMultiple(input);\n"
+            "    } else {\n"
+                        "      output = await Promise.all(input.map((s) => doenetMLToPretext(s, { fragment: true })));\n"
+            "    }\n"
```
dqnykamp: The single-item branch here returns a standalone PreTeXt document (with `<?xml…?>` and `<pretext>` wrapper), while the multi-item branch returns fragments with prefixed `xml:id`s. So a caller of `convert_multiple_doenetml_to_pretext(["x"])` gets a document, but `convert_multiple_doenetml_to_pretext(["x", "y"])` gets fragments — same function, different output shape depending on list length.

Suggest dropping the length-1 special case entirely and always invoking `converter.convertMultiple(input)` so the output contract is uniform. `DoenetConverter.convert` (the single-string entry point) should keep using a separate, standalone-mode eval path rather than routing through `convert_multiple`.

While here, the `convertMultiple === 'function'` fallback below silently drops the `prefixIds` step, so collisions would be returned without warning — since the JS bundle ships alongside this wrapper, the runtime check seems unnecessary and could be removed too.

packages/doenetml-to-pretext-python/doenetml_to_pretext/converter.py
```diff
@@ -49,13 +49,27 @@ def _resolve_dist_index(self) -> Path:
...
+            "  }\n"
+            "  let output;\n"
+            "  if (input.length === 1) {\n"
+                        "    output = [await doenetMLToPretext(input[0], { fragment: false })];\n"
+            "  } else {\n"
+            "    const converter = new DoenetMLToPretext();\n"
+            "    if (typeof converter.convertMultiple === 'function') {\n"
+            "      output = await converter.convertMultiple(input);\n"
+            "    } else {\n"
+                        "      output = await Promise.all(input.map((s) => doenetMLToPretext(s, { fragment: true })));\n"
```
dqnykamp: Stray indentation: these two `output = …` lines begin with ~24 spaces of literal whitespace, breaking the otherwise consistent alignment of the f-string fragments. It's not a runtime bug (the strings concatenate fine and the JS parser ignores leading whitespace), but it reads as an editor artifact. Worth reformatting before merge.

packages/doenetml-to-pretext/src/utils/pretext/render-to-pretext.tsx
```diff
@@ -22,7 +22,18 @@ import { normalizeAttrs } from "./normalize-attrs";
...
+         * Whether to render a standalone PreTeXt document. That is, one that starts with `<?xml version="1.0" encoding="UTF-8"?>`
+         * and has a single root `<pretext>` tag. If false, the output may not be valid PreTeXt, but should be a fragment of valid
+         * PreTeXt suitable for embedding inside a larger document.
+         *
+         * Default: `false`.
+         */
+        fragment?: boolean;
+    } = {},
 ): Xast.Root {
+    const { fragment: fragment = false } = options;
```
dqnykamp: Minor: `fragment: fragment` is a no-op rename. `const { fragment = false } = options;` is equivalent and cleaner.

packages/doenetml-to-pretext/src/index.ts
```diff
@@ -7,6 +7,9 @@ import "./index-inline-worker";
 import { renderFlatDastToPretext } from "./utils/pretext/render-to-pretext";
 import { preprocessDastForPretext } from "./utils/pretext/preprocess-dast";
 import { toXml as xastToXml } from "xast-util-to-xml";
+import { prefixIds } from "./utils/pretext/prefix-ids";
+
+export type ConvertOptions = Parameters<typeof renderFlatDastToPretext>[1];
```
dqnykamp: Deriving `ConvertOptions` via `Parameters<typeof renderFlatDastToPretext>[1]` is fragile — any future change to that function's signature (e.g. adding a non-optional second arg, or replacing the inline object literal) silently changes the exported public type, and ripples through `DoenetMLToPretext.convert`'s intersection type. Suggest extracting the options shape into a named exported interface and having both `renderFlatDastToPretext` and its consumers reference it explicitly.

Also note that `_flatDastToPretext` below re-spells the full `Parameters<typeof renderFlatDastToPretext>[1] = {}` rather than reusing this `ConvertOptions` alias — should use the alias once it exists.

dqnykamp (approved):

PR #1235, opened 2026-05-29
Print Support through PreTeXt

siefkenj:
This PR adds printing support to DoenetML activities by converting the source to PreTeXt source and then rendering the pretext source in the browser, via calling pretext with Pyodide.

The PR contains copies of the pretext source, which should not be reviewed (and counts for most of the code). Since PreTeXt often changes its internal API, I suggest keeping the source in-tree rather than having a build script fetch the latest commit on every build. (An alternative could be to pin to a specific commit, or even have a git sub-tree).

The script works by converting DoenetML -> PreTeXt, calling PreTeXt to compile, calling PreFigure to compile <graph> tags, then inlining all figures and CSS.

Much of the code was modified from https://github.com/siefkenj/pretext-book/tree/main/packages/pretext-playground


packages/doenetml-print/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "compilerOptions": {
+        "outDir": "./dist/",
+        "rootDir": "./src"
+    },
+    "include": [
+        "./src/**/*.ts",
+        "./src/**/*.tsx",
+        "./src/**/*.d.ts",
+        "./devs/**/*.ts",
+        "./devs/**/*.tsx"
```
dqnykamp: **Wrong directory name.** `include` lists `./devs/**`, but the directory is `dev/` (not `devs/`). As a result the dev UI (`dev/main.tsx`) is never type-checked. Either fix these to `./dev/**` or drop the two `devs` entries.

packages/doenetml-print/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Building
+
+Before building, refresh the bundled PreTeXt core files to the latest upstream commit:
+
+```bash
+./update_pretext_core.sh
+```
+
+This script downloads the latest PreTeXt repository archive (master, with fallback to main), 
+then updates [pretext_core/schema](pretext_core/schema), [pretext_core/xsl](pretext_core/xsl),
```
dqnykamp: **Docs don't match the script.** This says the script updates `pretext_core/schema`, `xsl`, `COPYING`, `CURRENT_COMMIT`, but `update_pretext_core.sh` actually copies `pretext`, `xsl`, `css` (plus `COPYING` + `CURRENT_COMMIT`). There is no `schema`, and `pretext`/`css` are undocumented.

packages/doenetml-print/src/ptx-compiler.ts
```diff
@@ -0,0 +1,489 @@
...
+                await prefigBrowserApi.initFinished;
+                this.pyodide.registerJsModule(
+                    "prefigBrowserApi",
+                    prefigBrowserApi,
+                );
+            } catch (e) {
+                reject(e);
+            }
+
+            resolve();
```
dqnykamp: **Async-executor anti-pattern + double-settle.** On error this calls `reject(e)` and then unconditionally `resolve()` afterward — the `resolve()` is a no-op once rejected, so harmless today but fragile. More importantly, on failure `pyodideInitPromise` stays a *rejected* promise, so the guard `await this.pyodideInitPromise` (line 68) re-throws on every later `init()` call and the compiler can never recover/retry. Consider clearing `this.pyodideInitPromise` on failure, and prefer a plain `async` method with try/catch over the `new Promise(async …)` executor.

packages/doenetml-print/src/ptx-compiler.ts
```diff
@@ -0,0 +1,489 @@
...
+                const parser = new DOMParser();
+                const doc = parser.parseFromString(indexHtml, "text/html");
+                const redirectElm = doc.querySelector(
+                    "meta[http-equiv=refresh]",
+                );
+                if (!redirectElm) {
+                    throw new Error("No redirect element found");
+                }
+                const redirectUrl =
+                    redirectElm.getAttribute("URL") ||
```
dqnykamp, line 360: **Meta-refresh DOM fallback can't find the URL.** An HTML `<meta http-equiv="refresh">` stores its target in the `content` attribute (`content="0; URL='…'"`), not a `URL`/`url` attribute, so `getAttribute("URL")` here is always `null`. The regex path above (line 350) handles the normal case, so this is a latent dead fallback rather than an active bug — but it won't do what it intends. Parse the redirect out of the `content` attribute instead.

packages/doenetml-print/src/ptx-compiler.ts
```diff
@@ -0,0 +1,489 @@
...
+                    ) {
+                        hastMutateToEmptyString(node);
+                    }
+
+                    // Inline svg images creatd by PreFigure. Thee have a class of `ChemAccess-element`
+                    // The name of the to-be-inlined svg is in the `data-src` attribute
+                    if (hastElementContainsClass(node, "ChemAccess-element")) {
+                        const dataSrc = node.properties["dataSrc"];
+                        if (typeof dataSrc === "string") {
+                            console.log("Inlining svg for", dataSrc);
```
dqnykamp, line 249: **Stray `console.log`/`console.error` in library code.** These "Inlining svg…" / "Inlining stylesheet…" logs (also lines 293, 311) will spam consumer consoles. Note the init `messageCallback` was deliberately silenced (line 105) for the same reason — gate these behind a debug flag or remove them for consistency.

packages/doenetml-print/vite.config.ts
```diff
@@ -0,0 +1,105 @@
...
+            dts({ rollupTypes: true }),
+            viteStaticCopyPyodide(),
+            vitePluginPrefigure(),
+            arraybuffer(),
+        ],
+        worker: {
+            format: "es",
+        },
+        server: {
+            host: "0.0.0.0",
```
dqnykamp, line 72: **Dev server binds `0.0.0.0`.** This exposes the dev server (which loads/executes WASM Python) to the local network. Probably fine for a dev-only tool, but worth being a deliberate choice rather than a default.

dqnykamp (commented):
Inline comments for the correctness and minor issues found during review. The two I'd fix before merge are the `tsconfig` `devs`→`dev` typo and the README/script mismatch; the rest are non-blocking cleanups.

packages/doenetml-print/dev/main.tsx
```diff
@@ -0,0 +1,208 @@
...
+                                const pretextOutput =
+                                    await doenetMLToPretextInstance.convert(
+                                        source,
+                                        {
+                                            fragment: false,
+                                            throwOnError: false,
+                                        },
+                                    );
+                                await pretexCompiler.init();
+                                await pretexCompiler.setMainPtx(pretextOutput);
```
dqnykamp: `setMainPtx` is synchronous, so the `await` here is harmless but misleading — drop it.

packages/doenetml-print/python/pretext_wasm/__init__.py
```diff
@@ -0,0 +1,208 @@
...
+
+    ptx.html(
+        ptx_source,
+        publication_source,
+        {},
+        "",
+        "html",
+        None,
+        None,
+        os.path.join(ptx_source, "../out"),
```
dqnykamp: **Path-building reads oddly.** Joining a *file* path with `"../out"` works only via `..` traversal from the file node. `os.path.join(os.path.dirname(ptx_source), "out")` is clearer (same for `"../out/generated"` in `compile_prefigure`). Not a bug. Also: "direcotry" typo at line 123.

packages/doenetml-print/update_pretext_core.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+DEST_DIR="$SCRIPT_DIR/pretext_core"
+TMP_ROOT="$SCRIPT_DIR/tmp"
+TMP_DIR=""
+ZIP_PATH="$TMP_DIR/pretext.zip"
```
dqnykamp: **Dead assignment.** `ZIP_PATH="$TMP_DIR/pretext.zip"` runs while `TMP_DIR=""`, then is correctly reassigned at line 23. Harmless leftover that can be removed.

.prettierignore
```diff
@@ -17,3 +17,5 @@ extension/build
 generated-assets
 multi-reporter-config.json
 jsxgraph-distrib
+doenetml-print/pretext_core
+pretext_wasm/core
```
dqnykamp: **Why CI prettier still fails.** Prettier ignore patterns follow `.gitignore` semantics: a pattern that contains a `/` anywhere but the end is **anchored to the directory of the ignore file** — here the repo root. So `doenetml-print/pretext_core` only matches `<root>/doenetml-print/pretext_core`, and `pretext_wasm/core` only matches `<root>/pretext_wasm/core`. Neither exists — the real paths are under `packages/`, so the `.scss` files in `pretext_core/css/**` are never ignored and CI (`prettier . --check`) flags them.

I verified locally: with these patterns prettier still warns on `packages/doenetml-print/pretext_core/css/**/*.scss`; adding the `packages/` prefix makes it pass (`All matched files use Prettier code style!`).

Fix — anchor the patterns to the actual locations:

```suggestion
packages/doenetml-print/pretext_core
packages/doenetml-print/python/pretext_wasm/core
```

(The second line is the symlink to `pretext_core`; an alternative single-line fix is just `**/pretext_core`, since a leading `**/` un-anchors the pattern to match at any depth.)

packages/doenetml-print/src/ptx-compiler.ts
siefkenj: We'll keep this for now.

packages/doenetml-print/vite.config.ts
siefkenj: This is needed to be able to access the dev server when running in a devcontainer.

dqnykamp (approved):

PR #1254, opened 2026-06-01
use whitespace delmiters (to match how attribute lists are actually used)

StevenClontz:
In today's workshop we got bit because the context menu suggested using `f, g` rather than `f g` for `functionSymbols` attribute


StevenClontz:
@djnykamp to confirm, every list that comes from an attribute in Doenet is whitespace-delimited (not comma-delimited) correct?

dqnykamp (approved):

dqnykamp:
> @djnykamp to confirm, every list that comes from an attribute in Doenet is whitespace-delimited (not comma-delimited) correct?

That's right. Good catch!

PR #1257, opened 2026-06-02
Add more to tags to pretext export

siefkenj:
This adds components up through P to the converter


packages/doenetml-to-pretext/src/renderers/pretext-xml/math-input.tsx
```diff
@@ -1,14 +1,18 @@
...
     const characters = 8;
+    const label = node.data.props.label;
 
     return (
-        <m>
-            <fillin characters={characters} />
-        </m>
+        <React.Fragment>
+            {label ? label + " " : null}
+            <m>
```
dqnykamp, line 29: In Doenet, a label can be inside an `<m>`, without an `<m>`, or mixed. For example, this is valid:
```
<mathInput>
  <label>hello <m>x^2</m></label>
</mathInput>
```
in which case the `label` prop will be `hello \(x^2\)`. The `label` prop already has the `\( \)` delimiters to determine which parts are latex.

I don't know if it easier to replace `\(` with `<m>` and `\)` with `</m>` or grab the `<label>` component and renderer it here rather than using the `label` state variable.

This is true for the labels of all the inputs, not just math input.

packages/doenetml-to-pretext/test/pretext-export.test.ts
```diff
@@ -125,6 +125,13 @@ describe("Pretext export", async () => {
         );
     });
 
+    it("mathInput renders its label", async () => {
+        source = `<answer><mathInput><label>My Label</label></mathInput></answer>`;
+        expect(await coreRunner.processToFlatDastAsFragment(source)).toContain(
+            `My Label <m><fillin characters="8"></fillin></m>`,
```
dqnykamp, line 131: In this case `My label` should not be in an `<m>`. The label should be in an `<m>` only if it has an explicit `<m>` (or similar, like `<math>`), such as `<label><m>x^2</m></label>`. It is possible for just part of a label to be in an `<m>`, as mentioned in the comment in the math input renderer.

packages/doenetml-to-pretext/test/pretext-export.test.ts
```diff
@@ -400,4 +444,175 @@ describe("Pretext export", async () => {
         expect(firstRefMatch![1]).toBe(firstIdMatch![1]);
         expect(secondRefMatch![1]).toBe(secondIdMatch![1]);
     });
+
+    it("<booleanInput> renders as math fillin", async () => {
```
dqnykamp, line 455: Same issue for the `<m>`s in these input renderers and mentioned with the math input renderer.

packages/doenetml-to-pretext/test/pretext-export.test.ts
```diff
@@ -400,4 +444,175 @@ describe("Pretext export", async () => {
...
+    it("<matrixInput> renders as math fillin", async () => {
+        source = `<matrixInput><label>Enter a matrix</label></matrixInput>`;
+        expect(
+            await coreRunner.processToFlatDastAsFragment(source),
+        ).toMatchInlineSnapshot(
+            `"Enter a matrix <m><fillin characters="8"></fillin></m>"`,
+        );
+    });
+
+    it("<orbitalDiagramInput> renders as math fillin", async () => {
```
dqnykamp, line 482: I have no idea what we should do for `<orbitalDiagramInput>`. I guess a blank is as good as anything. Or just don't renderer it.

dqnykamp (changes requested):
Look good. The only concern is rendering math and/or text inside a `<label>` in an input.

packages/doenetml-to-pretext/test/pretext-export.test.ts
siefkenj: I think a blank is better than nothing. When we have chemistry users printing, the can give us feedback about what it should look like in print.

packages/doenetml-to-pretext/test/pretext-export.test.ts
siefkenj: In other words, this test is correct?

dqnykamp: Ah, right!  My eyes didn't read where the `<m>`s were correctly.

packages/doenetml-to-pretext/src/renderers/pretext-xml/math-input.tsx
siefkenj: So, I'm not really sure a _good_ way to solve this issue. I could parse and put in <m> tags. But, as it is, when you use pretext to print, it does actually render the math appropriately!

dqnykamp: That's surprising. So, it's not urgent for our print use-case. We should also make sure it works for the scenario where the resulting pretext is converted to other outputs beside latex. The pretext html output would seem to be an unlikely target of this pretext conversion, as that would require someone to convert the doenetml explicitly to pretext and then include the pretext markup directly. But there are other static outputs of pretext that might not render the `\( latex \)` as formatted math.

I'm torn about the idea of just rendering the `<label>` child directly and then having the `<label>` component render its children directly when it is not in a graph. That would solve this issue, but then the `<label>` would have more capabilities when it is outside a graph. Inside a graph, we rely on jsxgraph and prefigure to render the latex within their labels. Other components like `<em>` are ignored in a graph, but would start working outside a graph.

siefkenj: Pretext-print uses the html output (mathjax is configured to automatically find the `\(..\)` pairs). Nevertheless, I did some string manipulation to force an `<m>` tag.

dqnykamp (approved):

PR #1266, opened 2026-06-03
Edits made by the devcontainer automatically upon provision

StevenClontz:
Currently the devcontainer makes two changes automatically upon provision:

- Reformat spacing in `devcontainer-lock.json`
- Change permissions for `convert.js`

We should either make these the canonical state of the repo (so contributors don't suggest these edits by mistake), or fix the codespace to not make these changes upon provision.


siefkenj:
These changes are very weird. Why are they happening?

dqnykamp:
Any progress or idea on what to do about this? PR #1275 includes these changes as well, and I'm not sure what to do about them. Should I revert them or just merge them in, making this PR obsolete?

StevenClontz:
Simplest solution is to just switch to these versions. Unless someone else
has automated settings that will try to revert them back. Certainly
whatever the devcontainer is set up to do is that should be in the repo (so
we fix the files or we fix the devcontainer).



Steven Clontz
https://clontz.org - ***@***.***

siefkenj:
Let's commit this PR, so we have a better idea of what caused the file changes.

siefkenj (approved):

PR #1270, opened 2026-06-03
Ignored testCode.doenet

StevenClontz:
This PR git ignores the testCode file so developers can tweak it as they please without accidentally commiting changes to it.


siefkenj:
@dqnykamp  isn't using the devcontainer, so we need a solution that works for him too.

StevenClontz:
This is unrelated to the devcontainer and should work for anyone.

siefkenj:
If I am understanding the code correctly, anyone who does a git clone and doesn't run the devcontainer won't have a `testCode.doenet` and so the dev server will fail.

StevenClontz:
`npm run dev` creates the `testCode.doenet` if it doesn't exist, no codespace involved

siefkenj:
> `npm run dev` creates the `testCode.doenet` if it doesn't exist, no codespace involved

Oh, I see! Can you add this to the `build:dev` as well?

siefkenj:
Actually, maybe it's best to handle this with wireit so the dependencies are clear. Let me know if you have trouble figuring out how to make wireit work.

StevenClontz:
I haven't done that before so it would take nonzero time to understand how to implement it. But I'll take a crack when I can

dqnykamp:
The codebase has 5 testCode.doenet files in 5 different packages. When we figure out the right approach, we should apply it to all 5 cases.

packages/doenetml-print/package.json
```diff
@@ -20,7 +20,7 @@
         }
     },
     "scripts": {
-        "dev": "vite",
+        "dev": "wireit",
```
StevenClontz, line 23: This is just a sample comment on a particular line of code because I'm demoing this active PR for a workshop today.

StevenClontz: We would have a conversation here and in the comments, and hopefully eventually resolve the conversation as I'll do momentarily.

StevenClontz:
This PR is updated so that all testCode files are ignored and populated by tracked defaults.

PR #1275, opened 2026-06-03
IFF + Implies Boolean Operators

sfrezza:
This adds two new Boolean Operator tags: <iff> and <implies>; includes updated test files and user documentation. Includes updates to two README.MD files to help others be able to build and test the documentation code.


siefkenj:
@sfrezza Thanks for the contribution! It looks like you need to apply prettier to some of the files.

dqnykamp:
That for this PR, @sfrezza. I'm curious what our author community thinks about adding the components `<iff>` and `<implies>`, so I posted [this discussion question](https://community.doenet.org/t/feedback-on-two-new-component-ideas-iff-and-implies/134).

We've been talking about having an RFC requirement for new components, attributes, or properties, but we don't currently having anything in place. We'll see if we can start some discussion on your proposed components. Feel free to post on that discussion, as you can probably explain them better than I did!

sfrezza:
The rationale for these changes is to provide a more complete set of boolean operators that map to the types of assignments used in Discrete Mathematics courses.

packages/docs-nextra/pages/reference/iff.mdx
```diff
@@ -0,0 +1,109 @@
...
+whether enclosed `<boolean>{:dn}` are true "if and only if." The biconditional is true in two cases, where either all statements are either all 'true' or all are 'false'. 
+
+---
+
+<a id="award-ex"></a>
+### Example: `<iff>{:dn}` in an `<award>{:dn}`
+
+
+```doenet-editor-horiz
+<p>Move the point to either be 1) near/on the intersection of the circle and 2) near/on the line and 3) in the first quadrant, or not anywhere near either the line or on the circle. (Scroll down to check work.)</p>
```
dqnykamp: If conditions 1, 2, and 3 are not satisfied, then all must be false. In particular, if the point is not anywhere near the line or the circle, then it also needs to avoid the first quadrant. (This last condition is missing in the statement.)

One option is to just avoid the third condition `$firstQuad` so the logic is clearer. Though it would be good to have one example (maybe the last one) address the more complicated case of having three booleans in the `<iff>`.

packages/docs-nextra/pages/reference/iff.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+An `<iff>{:dn}` component is used to construct a compound boolean from three referenced `<boolean>{:dn}` components 
+within an `<award>{:dn}` (e.g., all 'true' or all 'false').
+
+
+---
+
+
+<a id="alternative-ex"></a>
+### Example: An alternative `<iff>{:dn}` use in an `<award>{:dn}`
```
dqnykamp: I find this last example confusing, given that the correctness of the answer does not correspond to the state condition.

I have a suggestion: delete this last example and then break up the previous example into two: one with just two conditions (e.g., ignoring the first quadrant) and then a second variation of it adding a third condition (e.g., add back the first quadrant condition). Then, the description/title of the last example can focus on how `<iff>` can take more than two arguments.

packages/docs-nextra/pages/reference/iff.mdx
```diff
@@ -0,0 +1,109 @@
...
+       <boolean name="check3">$notN = 7</boolean>
+  </or></boolean>
+
+<p>Either both conditions are true, or both are false: 
+  <boolean name="bothTrueOrBothFalse">
+    <iff>$nCorrect $notNCorrect</iff>
+  </boolean></p>
+```
+
+The word `iff` may be typed directly within any component that takes a boolean
```
dqnykamp: I'm OK deleting this last paragraph. I know it was copied from another example, but I think the idea of having three arguments to `<iff>` is more interesting.

packages/docs-nextra/pages/reference/implies.mdx
```diff
@@ -0,0 +1,80 @@
...
+<a id="boolean-ex"></a>
+
+### Example: `<implies>{:dn}` in a `<boolean>{:dn}`
+```doenet-editor-horiz
+<p><mathInput name="n"><label>Enter a number ≤ 10, or equal to 15</label></mathInput></p>
+      
+<boolean name="check1">$n > 10</boolean>
+<boolean name="check2">$n = 15</boolean>
+      
+<p>You have entered a number ≤ 10 or = 15: ):
```
dqnykamp: Delete the orphaned `):`.

packages/docs-nextra/pages/reference/implies.mdx
```diff
@@ -0,0 +1,80 @@
...
+whether enclosed `<boolean>{:dn}` are true when P implies Q. This is true when either the first argument is 'false' OR the second argument is 'true'.
+
+---
+
+<a id="award-ex"></a>
+### Example: `<implies>{:dn}` in an `<award>{:dn}`
+
+
+```doenet-editor-horiz
+<p>Move the point to either be 1) not on the circle or 2) near/on the line a. (Scroll down to check work.)</p>
```
dqnykamp: On my computer, I don't have to scroll down to see the check work. Does it require scrolling on yours?

packages/docs-nextra/pages/reference/implies.mdx
```diff
@@ -0,0 +1,80 @@
...
+```
+
+An `<implies>{:dn}` component is used to construct an ordered binary boolean from two referenced `<boolean>{:dn}` components 
+within an `<award>{:dn}` (e.g., first argument is 'false' or second is 'true').
+
+
+---
+
+
+The tag `implies` may be typed directly within any component that takes a boolean
```
dqnykamp: Delete this paragraph.

packages/docs-nextra/pages/reference/iff.mdx
dqnykamp: Coming back to this example, I noticed that I don't need to scroll down to see the check work for this example, either.

dqnykamp (changes requested):
The components look good. I'm just requesting changes for the documentation.

I'm not sure what is going on with devcontainer-lock.json and webwork-to-doenetml/convert.js, but it looks like folks are discussing in in #1266. If you can revert those, that's great. Otherwise, I'll just revert them before committing, and let them figure out what to do with those in #1266.

dqnykamp:
I think all the copilot comments make sense. It caught issues I missed.

dqnykamp:
@sfrezza Would you like any help addressing the review comments?

sfrezza:
Yes - Help finishing this would be appreciated. I think i fixed and committed changes to address the issue.
I'm not sure what my next steps are. These changes seem reasonable... i think. Is it that I have to approve them somehow?
STF

dqnykamp:
I see that your new commit addressed the copilot comments about the test names. I clicked the "Resolve Conversation" button so those comments are now closed. 

Copilot also had four comments about the documentation files. They all look like typos and copy-and-paste errors. Could you fix those and the click the "Resolve conversation" button when you push the fix? Or, if you disagree, you can just type a reply to the specific comment so I know that you have a different opinion.

My comments, which are above the copilot comments, are all about the documentation files. One is a duplicate about the same typo copilot found. The rest are requests to change the examples from the documentation. If you agree with the change I propose, you can make the change, optionally reply, and then resolve the conversation. If you disagree, let me know how you disagree by replying to the comment.

When you've finished that, I can submit another review and approve the changes if everything looks good. Once the changes are approved, I can merge this PR into the main branch.

packages/docs-nextra/pages/reference/implies.mdx
```diff
@@ -0,0 +1,80 @@
...
+<boolean name="check1">$n > 10</boolean>
+<boolean name="check2">$n = 15</boolean>
+      
+<p>You have entered a number ≤ 10 or = 15: ):
+  <boolean>
+    <implies>$check1 $check2</implies>
+  </boolean></p>
+```
+
+The `<implies>{:dn}` component functions as the boolean logical biconditional operator `imp` or `→`, and checks to see
```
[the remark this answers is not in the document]
sfrezza: binary is the term i was looking for.

sfrezza:
Check the two recent commits - if i missed something, please just fix it. These are all appropriate changes.

dqnykamp:
I rebuilt the schema and updated the documentation pages.

dqnykamp (approved):

PR #1313, opened 2026-06-13
Drive doenetml-prototype from the JS core via FlatDast updates

### LINKED ISSUE #1310: Send FlatDast updates from the JavaScript core to doenetml-prototype
Groundwork for driving the `doenetml-prototype` renderers from the `doenetml-worker-javascript` core, communicating exclusively through the new FlatDast format. This is one more step toward the ultimate goal of converting to `doenetml-worker-rust`.

The initial render already works (`returnFlatDastFromJS` / `flatDastFromJS`); this issue adds the **update** path so that, after a user interaction, the prototype renderers re-render from JS-core state.

### Initial render (already done)
- `CoreWorker.returnFlatDastFromJS()` ([packages/doenetml-worker/src/CoreWorker.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker/src/CoreWorker.ts)) spins up the JS core, captures the one-shot `updateRenderersCallback({ init: true })` payload, and calls `flatDastFromJS(documentToRender, updateInstructions)`.
- `flatDastFromJS` ([packages/doenetml-worker/src/flatDastFromJS.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker/src/flatDastFromJS.ts)) turns the JS `documentToRender` + `rendererStatesToUpdate` into a `FlatDastRootWithErrors` (a sparse `elements` array indexed by `componentIdx`, each with `data.props`), applying per-component JS→Rust fixups (`textJsToRust`, `sectionJsToRust`, `pointJsToRust`, `refJsToRust`, `coords`→`math`).

### How the prototype consumes updates (Rust path today)
- [packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts)
  - `retrieveDast` → `worker.returnDast()` → `_setFlatDastRoot(flatDast)`.
  - `dispatchAction` → `worker.dispatchAction(action)` → returns an `ActionResponse`; dispatches `processElementUpdates(updates.payload)`.
- [packages/doenetml-prototype/src/state/redux-slices/dast/slice.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-prototype/src/state/redux-slices/dast/slice.ts)
  - `processElementUpdates` reducer iterates `payload` entries `[componentIdx, update]`, then for each element applies `Object.assign(elm.data.props, update.changedState)` and, if present, `elm.children = update.newChildren`.
- Payload type: `ActionResponse["payload"]` = `HashMap<ComponentIdx, FlatDastElementUpdate>` where `FlatDastElementUpdate = { new_children?: FlatDastElementContent[]; changed_state?: ForRenderProps }` (serialized to camelCase `newChildren` / `changedState` on the wire).
- The Rust analogue that produces this is `DocumentRenderer::get_flat_dast_updates` ([to_flat_dast.rs](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker-rust/lib-doenetml-core/src/core/document_renderer/to_flat_dast.rs)).

### The fundamental mismatch — pull vs. push
- **Rust core**: `dispatch_action(action)` **returns** the update map. The prototype's `dispatchAction` thunk is built around this pull model.
- **JS core**: `requestAction()` resolves with only `{ actionId }`. The actual renderer updates are **pushed** asynchronously — possibly multiple times and even with no triggering action (animations, async state resolution) — through the `updateRenderersCallback` ([RendererInstructionBuilder.updateRendererInstructions](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker-javascript/src/core/RendererInstructionBuilder.ts), ~L313), as:
  ```
  updateInstructions = [{
    instructionType: "updateRendererStates",
    rendererStatesToUpdate: [
      { componentIdx, stateValues, rendererType, childrenInstructions? }, ...
    ],
    sourceOfUpdate,
  }]
  ```
  Note the initial render emits the singular `"updateRendererState"`; later batches emit plural `"updateRendererStates"`. `flatDastFromJS` ignores the string and reads `updateInstructions[0].rendererStatesToUpdate` directly.

Reconciling this push model with the prototype's pull-based thunk is the core of the work.

### Part A — a pure converter: `flatDastUpdateFromJS`
New function alongside `flatDastFromJS` (same file or a sibling `flatDastUpdateFromJS.ts` in `packages/doenetml-worker/src/`):

```
flatDastUpdateFromJS(
    updateInstructions: UpdateInstruction[],
): Record<number, FlatDastElementUpdate>
```

Behavior, mirroring `get_flat_dast_updates` + the existing initial converter:
1. Flatten `updateInstructions[*].rendererStatesToUpdate`.
2. For each `rendererState`:
   - `changedState` ← `rendererState.stateValues`, **after** running the same per-component JS→Rust fixups used in `flatDastFromJS`. Factor those fixups out of `flatDastFromJS` into a shared helper keyed on the element/renderer type so both code paths stay in sync. (The element `name`/type is needed to pick the fixup; carry `rendererType`/`componentType` through, or look it up from the already-built `elements` map — see Part C.)
   - `newChildren` ← only when `rendererState.childrenInstructions` is present; convert each child instruction into `FlatDastElementContent` (`{ id, annotation: "original" }` for components, raw `string` for text), reusing the child-handling logic currently inlined in `flatDastFromJS`.
3. Key the result map by `componentIdx`.

Refactor note: the child-instruction → `FlatDastElementContent` logic and the per-type fixups currently live inside `flatDastFromJS`. Extract them into shared helpers so the initial and update paths cannot drift.

### Part B — bridging push → pull in `CoreWorker.ts`
Two viable strategies; **B1 is recommended** for the smallest, most faithful first step.

**B1 (recommended): buffer-and-return.** Add `dispatchActionJavascriptFlat(actionArgs)` to `CoreWorker`:
1. Install an internal `updateRenderersCallback` (set up once during the JS-core init used by `returnFlatDastFromJS`) that appends every pushed `updateInstructions` batch to a buffer on `CoreWorker`.
2. On dispatch: clear the buffer, `await this.javascriptCore.requestAction(...)`, then drain the buffer, run each batch through `flatDastUpdateFromJS`, merge the per-`componentIdx` maps (later batches win / are `Object.assign`-merged), and **return** the merged `Record<number, FlatDastElementUpdate>`.
3. The prototype thunk treats this exactly like the Rust `dispatchAction` payload — no prototype change required for the common case.

Caveats to document in code:
- Updates that arrive with **no** in-flight action (animations, deferred async resolution) are not delivered by a return-based API. Acceptable for the initial milestone (the few prototype components are static-ish); note it as a known gap.
- `requestAction` resolving does not guarantee all async renderer pushes have flushed. For the first pass, drain synchronously-available batches; revisit if a component needs post-resolution updates.

**B2 (later/optional): true push channel.** Have the prototype register a `Comlink.proxy(updateRenderersCallback)` (as the `doenetml` package does in [DocViewer.tsx](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml/src/Viewer/DocViewer.tsx) ~L1251), with `CoreWorker` converting each batch via `flatDastUpdateFromJS` and pushing it across. The prototype dispatches `processElementUpdates` on each message. This handles animations/async but requires a new worker→main message path and prototype wiring. Defer until B1 is proven.

### Part C — element-type lookup for fixups
The update path only has `componentIdx` + `stateValues`, but the JS→Rust fixups are chosen by element type. Options:
- Reuse the `elements` map already produced by the initial `flatDastFromJS` (the prototype keeps the full `flatDastRoot`), or have `CoreWorker` retain a `componentIdx → name` map built during initial render and pass it into `flatDastUpdateFromJS`. Recommended: maintain the `componentIdx → name` map in `CoreWorker` so the converter stays pure.
- `rendererStatesToUpdate` also carries `rendererType`; confirm whether that is sufficient to select the fixup, which would avoid the extra map.

### Part D — end-to-end Cypress harness for the prototype
There is no Cypress harness for `doenetml-prototype` today (the package has an unused `cypress` devDependency only; `test-viewer`'s prototype support is a manual React `usePrototype` checkbox that specs cannot drive). **Extend the existing `@doenet/test-cypress` package rather than building a new harness.** Two signals show this is the intended home: its wireit `build` already declares `../doenetml-prototype:build` as a dependency (latent — the prototype is not yet imported in `src`), and its vite config already copies `@doenet/doenetml-worker/index.js` to `/doenetml-worker/`, exactly the path the prototype's `getWorkerUrl()` resolves. All the hard infra (cypress config, grep groups, a11y tasks, multi-reporter, preview server on `:4173`, and the `window.onmessage`/`postMessage` content channel in `CypressTest.tsx`) is reused as-is.

Wiring:
- In [packages/test-cypress/src/CypressTest.tsx](https://github.com/Doenet/DoenetML/blob/main/packages/test-cypress/src/CypressTest.tsx), import `DoenetML` from `@doenet/doenetml-prototype` (plus its stylesheet) and render it instead of `DoenetViewer` when a flag is set. Gate it so existing `@doenet/doenetml` specs are untouched — either a new `postMessage` field (e.g. `e.data.usePrototype`) or a dedicated route/query param (e.g. `?prototype=1`).
- Specs drive content with the existing `cy.window().then(win => win.postMessage({ doenetML, usePrototype: true }, "*"))` mechanism.

`textInput` is the ideal first interactive component: the prototype renderer ([text-input.tsx](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-prototype/src/renderers/doenet/text-input.tsx)) already dispatches `updateImmediateValue` / `updateValue` actions on DOM events, and both the Rust core and (post-plan) the JS core implement those actions plus a `for_render` `immediateValue`/`ImmediateValue` prop. A spec that types into the input and asserts a bound `$ti.immediateValue` echo updates is therefore **green today against the default Rust core**, and becomes the JS-core validation once the prototype is pointed at the JS core (`setCoreType("javascript")` + `dispatchActionJavascriptFlat`). Caveat: under the B1 buffer-and-return bridge this only verifies action-driven updates; animation/async-only pushes need B2.

## Implementation steps
1. **Refactor `flatDastFromJS`** to extract: (a) child-instruction → `FlatDastElementContent`, and (b) per-type JS→Rust prop fixups, into shared helpers. No behavior change; covered by existing `doenetml-to-pretext` output.
2. **Add `flatDastUpdateFromJS`** (Part A) using those helpers.
3. **Wire the JS-core init** in `CoreWorker` so a persistent `updateRenderersCallback` buffers update batches and a `componentIdx → name` map is retained (Part C). Reuse/share the init currently inside `returnFlatDastFromJS` rather than duplicating it.
4. **Add `dispatchActionJavascriptFlat`** (Part B1) returning the merged update map.
5. **Prototype wiring**: allow the prototype to select the JS core (`worker.setCoreType("javascript")` instead of `"rust"` in [core/thunks.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts) `_loadWorker`) and route `dispatchAction` to `dispatchActionJavascriptFlat` when the JS core is active. Keep behind a flag/config so the Rust path stays default.
6. **Add prototype render path to `@doenet/test-cypress`** (Part D): import the prototype `DoenetML` in [CypressTest.tsx](https://github.com/Doenet/DoenetML/blob/main/packages/test-cypress/src/CypressTest.tsx), gated by a `postMessage` flag (`e.data.usePrototype`) or `?prototype=1` route so existing specs are unaffected.
7. **Tests**:
   - Unit (Vitest) for `flatDastUpdateFromJS` mirroring the existing `flatDastFromJS` coverage, including a `childrenInstructions` case (→ `newChildren`) and a `stateValues`-only case (→ `changedState`).
   - Worker-level (Vitest) against one prototype-supported component (e.g. `text`): initial FlatDast + an action that mutates a `forRenderer` state value, asserting the returned update map matches what `processElementUpdates` needs.
   - **Cypress e2e** (`packages/test-cypress/cypress/e2e/prototype/textInput.cy.js`, tagged into a grep group): post a `<textInput name="ti"/>` document with a bound `$ti.immediateValue` echo, type into the input, and assert the echoed DOM updates. Add it green against the Rust core first; flip the harness flag to the JS core once Part B1 lands so the same spec validates the JS-core FlatDast update round-trip.

## Known gaps / follow-ups
- No-action / animation pushes (needs B2 push channel).
- Error elements in updates (`processElementUpdates` already throws "Updating errors is not yet implemented" — out of scope here).
- Attribute updates (`changedAttributes`) — also unimplemented on the consumer side; out of scope.
- `effectiveIdx` vs `componentIdx` divergence (same XXX caveat already noted in `flatDastFromJS`).

## Touch list
| File | Change |
|------|--------|
| [packages/doenetml-worker/src/flatDastFromJS.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker/src/flatDastFromJS.ts) | Extract shared helpers (child content, per-type fixups) |
| `packages/doenetml-worker/src/flatDastUpdateFromJS.ts` (new) | `flatDastUpdateFromJS` converter |
| [packages/doenetml-worker/src/CoreWorker.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-worker/src/CoreWorker.ts) | Persistent update-buffer callback, `componentIdx→name` map, `dispatchActionJavascriptFlat` |
| [packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts](https://github.com/Doenet/DoenetML/blob/main/packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts) | Optional JS-core selection + route to flat action dispatch |
| [packages/test-cypress/src/CypressTest.tsx](https://github.com/Doenet/DoenetML/blob/main/packages/test-cypress/src/CypressTest.tsx) | Render prototype `DoenetML`, gated by a `postMessage`/route flag |
| `packages/test-cypress/cypress/e2e/prototype/textInput.cy.js` (new) | Interaction→DOM-change e2e (Rust core now, JS core post-B1) |
| Vitest specs | Coverage for the new converter and a worker-level text update |

---

dqnykamp:
Completes the second half of #1310. PR #1311 landed the worker-side groundwork — the pure converter `flatDastUpdateFromJS` and the push→pull bridge `dispatchActionJavascriptFlat`. This PR wires the prototype to actually use them, then hardens the result: along the way it surfaced and fixed several real bugs in driving the `doenetml-prototype` renderers from the JavaScript core.

## Prototype JS-core selection + flat dispatch (issue step 5)

`packages/doenetml-prototype`

- Add a `coreType` (`"rust" | "javascript"`, default `"rust"`) field to the core redux slice + `_setCoreType` reducer.
- Thread `coreType` through `DoenetML` → `PageViewer` → `setSourceAndStartWorker`, which sets it in the store **before** `setSource` starts the worker so `_loadWorker` calls `worker.setCoreType(coreType)`.
- The dev page persists the selected core in `?core=` and performs the documented full-page reload when switching cores, ensuring the cached worker is recreated with the selected core. The rust core remains the default.
- Route the `dispatchAction` thunk to `dispatchActionJavascriptFlat` when the JS core is active, dispatching the returned update map straight into `processElementUpdates`. The rust pull path is untouched and remains the default.
- `EditorViewer` also accepts `coreType` and forwards it to its viewer panel (the editor-only PreTeXt/Markdown export features read the FlatDast from the store and are intentionally left on the rust path).

## test-cypress prototype render path (issue step 6)

`packages/test-cypress`

- Render the prototype `DoenetML` when a message carries `usePrototype: true`, with an optional `prototypeCoreType`. Gated so existing `@doenet/doenetml` specs are unaffected.
- The prototype renderer (and its stylesheet) is `React.lazy`-loaded via a `PrototypeRenderer` wrapper, so the prototype's Tailwind-preflight CSS is only applied when a spec actually renders the prototype — otherwise it leaks global resets into the `@doenet/doenetml` viewer and breaks unrelated specs (e.g. `graph.cy.js` pixel-width assertions).
- Declare `@doenet/doenetml-prototype` as a workspace dependency (`"*"`, since the prototype is unpublished).

## Bug fixes surfaced while testing/reviewing

1. **Duplicated section title** (`sectionJsToRust`) — when a `<section>`'s `<title>` references a value, a JS-core update left `xrefLabel.label` set to the full title text (the fixup only cleared it when the section's children happened to be materialized in the batch), so the header rendered the title twice. Now cleared whenever the section has a title child.
2. **Graph interactions were silent no-ops** — the prototype dispatches the rust action names `move` / `changeBoundingBox`, but the JS core registers `movePoint` / `changeAxisLimits`, so dragging a point or panning the graph did nothing (core state never changed; referencing components never updated). Added a `jsCoreActionNames` translation applied in `dispatchActionJavascriptFlat`, keyed by component type. Audited the full rust action surface — these two graph actions are the only divergences; args are compatible.
3. **`<m>` rendered empty under the JS core** — the prototype `M` renderer was written for the rust core's child-`<text>` representation; the JS core (like production `@doenet/doenetml`) supplies the content in a `latex` prop. The renderer now prefers `latex`, falling back to children, so it works against both cores.

## Tests

- **Worker (Vitest):** converter unit tests, action-name translation unit tests, and live-JS-core integration tests covering `textInput`, point `move` → `$P.x`, `changeBoundingBox` → bound `$xmax`, point-in-text `coordsLatex`, `$P.x` math, and `<m>` latex (initial + update). The graph tests derive the component type from the initial-render-seeded `componentIdx→name` map, exercising the same lookup the bridge performs.
- **Cypress (both cores):** `textInput` bound echoes, the section-title duplication regression, and `<m>` rendering.

## Known gaps / follow-ups

- **B1 bridge limitation** (carried over from #1311): only action-driven updates are delivered; no-action/animation pushes need the B2 push channel.
- **No e2e graph drag/pan spec** — JSXGraph drag automation is flaky, so the graph action path is covered deterministically at the worker level instead.
- **#1314** — pre-existing rust-core bug: `<m>` with a reference renders as `[object Object]` (orthogonal to this PR; filed as a follow-up).
- **Newly introduced JS-core children** — if an action reveals or generates a component absent from the initial FlatDast, the current return-based update bridge may not provide a full element definition for that id. This is a follow-up limitation, not implemented in this PR.

Closes #1310.


packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -37,7 +41,8 @@ export const coreThunks = {
             // We need to load a new worker
 
             const worker = createWrappedCoreWorker();
-            await worker.setCoreType("rust");
+            const { coreType } = selfSelector(getState());
+            await worker.setCoreType(coreType);
             const key = workerCache.length;
             workerCache[key] = { worker };
             dispatch(_coreReducerActions._setWorkerCacheKey(key));
```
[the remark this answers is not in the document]
dqnykamp: Fixed in 26bc60600. The worker cache is now keyed by `coreType`, and `_loadWorker` reuses a cached worker only when it is running the requested core — otherwise it creates a fresh worker and applies `setCoreType`. (In practice the dev page still switches cores via a full page reload, but this makes the redux/worker layer correct on its own.)

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,23 +4,28 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
     const errors = useAppSelector(errorsSelector);
     const isInErrorState = errors.length > 0;
 
     React.useEffect(() => {
         dispatch(globalActions.watchForDarkModePreferenceChange());
-        dispatch(dastActions.setSourceAndStartWorker({ source, flags }));
+        dispatch(
+            dastActions.setSourceAndStartWorker({ source, flags, coreType }),
+        );
     }, [source]);
```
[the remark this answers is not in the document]
dqnykamp: Fixed in 26bc60600. `coreType` is now included in the effect's dependency array, so a `coreType` change re-runs `setSourceAndStartWorker` and re-initializes the worker for the selected core.

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,24 +4,29 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
     const isInErrorState = errors.length > 0;
 
     React.useEffect(() => {
         dispatch(globalActions.watchForDarkModePreferenceChange());
-        dispatch(dastActions.setSourceAndStartWorker({ source, flags }));
-    }, [source]);
+        dispatch(
+            dastActions.setSourceAndStartWorker({ source, flags, coreType }),
+        );
+    }, [source, coreType]);
```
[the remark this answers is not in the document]
dqnykamp: Fixed in ba1e8ee8f. The dark-mode `matchMedia` watcher now runs in its own one-time `useEffect(..., [])`, separate from the `setSourceAndStartWorker` effect, so it no longer registers duplicate listeners on `source`/`coreType` changes.

packages/doenetml-prototype/src/renderers/doenet/m.tsx
```diff
@@ -4,16 +4,35 @@ import { BasicComponentWithPassthroughChildren } from "../types";
...
+        latex != null && latex !== ""
+            ? latex
+            : Array.isArray(children)
+              ? children.join("")
+              : String(children);
+
     if (onServer) {
-        return <span className="process-math">{children}</span>;
+        return <span className="process-math">{content}</span>;
     }
```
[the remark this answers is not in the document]
dqnykamp: Fixed in ba1e8ee8f. The server-render branch now renders the React `children` directly when there is no `latex` prop, preserving the rust core's referenced child element instead of stringifying it to `[object Object]`. The JS core's `latex` prop is still used when present, and the client (MathJax) branch keeps the string fallback. (The client-branch rust-core reference case remains the separate, pre-existing issue tracked in #1314.)

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -29,17 +37,21 @@ export const coreThunks = {
     _loadWorker: createLoggingAsyncThunk(
         "core/loadWorker",
         async (_: void, { dispatch, getState }) => {
-            const { workerCacheKey } = selfSelector(getState());
-            if (workerCacheKey != null && workerCache[workerCacheKey]?.worker) {
-                // There's already a worker loaded
+            const { workerCacheKey, coreType } = selfSelector(getState());
+            const cachedWorker = workerCache[workerCacheKey ?? -1];
```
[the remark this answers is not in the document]
dqnykamp: Fixed in e79604ace. When the currently-selected worker runs a different core, `_loadWorker` now searches the cache for an existing worker running the requested core and reuses it (updating `workerCacheKey`), only allocating a new worker when none exists. This avoids duplicate workers when switching cores back and forth.

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,24 +4,35 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
         dispatch(globalActions.watchForDarkModePreferenceChange());
-        dispatch(dastActions.setSourceAndStartWorker({ source, flags }));
-    }, [source]);
+    }, []);
+
+    React.useEffect(() => {
+        dispatch(
+            dastActions.setSourceAndStartWorker({ source, flags, coreType }),
+        );
+    }, [source, coreType]);
```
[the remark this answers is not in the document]
dqnykamp: Fixed in ce5e7bd81. The effect now depends on a serialized key of the flag values (`JSON.stringify(flags)`) in addition to `source`/`coreType`, so it re-runs when a flag value actually changes and propagates the new flags to the worker — without re-running on every render just because the parent rebuilt the `flags` object.

packages/test-cypress/src/CypressTest.tsx
```diff
@@ -157,6 +167,8 @@ export function CypressTest() {
         let newAnswerResponseCounts: Record<string, number> | undefined =
             undefined;
         let newDoenetMediaUrl: string | undefined = undefined;
+        let newUsePrototype: boolean | undefined = undefined;
+        let newPrototypeCoreType: "rust" | "javascript" | undefined = undefined;
```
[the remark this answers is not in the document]
dqnykamp: Fixed in 3416b2cf9. `newUsePrototype` / `newPrototypeCoreType` now default to the current state values and are only overwritten when the message explicitly includes `usePrototype` / `prototypeCoreType`, so the selected renderer stays stable across incremental messages (matching how `attemptNumber` is preserved). Verified the prototype specs still pass and a standard viewer spec is unaffected.

packages/doenetml-prototype/dev/main.tsx
```diff
@@ -1,14 +1,90 @@
...
+            <div
+                style={{
+                    flex: "0 0 auto",
+                    display: "flex",
+                    alignItems: "center",
+                    gap: "12px",
+                    padding: "8px 12px",
+                    borderBottom: "1px solid #ccc",
+                    background: "#f5f5f5",
+                }}
```
siefkenj: Create a CSS file and put styles in there instead.

packages/doenetml-prototype/src/renderers/doenet/m.tsx
```diff
@@ -4,19 +4,45 @@ import { BasicComponentWithPassthroughChildren } from "../types";
...
 
-export const M: BasicComponentWithPassthroughChildren = ({ children }) => {
+/**
+ * The two cores represent `<m>` differently: the rust core renders its content
+ * as a child `<text>` element (consumed here via passthrough children), while
+ * the JavaScript core (like production `@doenet/doenetml`) puts the rendered
+ * LaTeX in a `latex` prop with no children. Prefer `latex` when present and
+ * fall back to children, so a single renderer works against both cores.
+ */
+type MData = { props: { latex?: string } };
```
siefkenj, line 19: I don't know how I feel about this. We need everyone to use the same props in the end. So the JsToDast function will have a bunch of special cases for a while to convert the Dast so we always have the assumed types.

HOWEVER, I don't think we know yet what are the "right" types. Answering that question will involve mapping out all the state vars/props and looking to see how we can make them consistent.

packages/doenetml-prototype/dev/main.tsx
```diff
@@ -1,14 +1,90 @@
 import React from "react";
 import { createRoot } from "react-dom/client";
-import { DoenetML, EditorViewer } from "../src/index";
+import { EditorViewer } from "../src/index";
 
 // @ts-ignore
 import doenetMLstring from "./testCode.doenet?raw";
-import { ResizablePanelPair } from "../src/editor/components/resizable-panel-pair";
+
+type CoreType = "rust" | "javascript";
```
siefkenj: Don't define this type multiple times. Export it from the main lib and then import it.

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -99,11 +124,37 @@ export const coreThunks = {
...
+
             try {
-                const updates = await worker.dispatchAction(action);
-                dispatch(
-                    _dastReducerActions.processElementUpdates(updates.payload),
-                );
+                if (coreType === "javascript") {
+                    // The JavaScript core pushes renderer updates asynchronously
+                    // rather than returning them; `dispatchActionJavascriptFlat`
+                    // bridges that to the same `Record<componentIdx, update>`
```
siefkenj: This comment is confusing. What does "bridges" mean?

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
```diff
@@ -99,11 +124,37 @@ export const coreThunks = {
...
+                    // shape the rust core returns in `ActionResponse["payload"]`.
+                    const updates = await worker.dispatchActionJavascriptFlat({
+                        actionName: action.actionName,
+                        componentIdx: action.componentIdx,
+                        args: action.args ?? {},
+                    });
+                    // The JS-core update map mirrors the rust
+                    // `ActionResponse["payload"]` shape that
```
siefkenj: Why doesn't it return the correct type already?

This seems more like an `// XXX: June 2026, the type returned by dispatchActionJavascriptFlat should be the same as ActionResponse["payload"]. Remove cast when this is the case`

packages/doenetml-worker/src/jsRustConversions/section.ts
```diff
@@ -40,13 +40,23 @@ export function sectionJsToRust(
...
             props.title = titleChildIdx;
 
+            // A title child means the section renders its title separately (via
+            // `props.title`), so the xrefLabel must not also carry the title
+            // text — otherwise the renderer shows it twice. Clear it whenever a
+            // title child exists, independent of whether `element.children` is
+            // materialized in this batch: the update path (`flatDastUpdateFromJS`)
+            // builds the synthetic element with only the children carried by the
+            // current update, which for a section state-only update is empty.
+            props.xrefLabel.label = "";
```
siefkenj: This doesn't make sense to me. What is going on? The renderer should know what to do an deal with this...

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
```diff
@@ -109,14 +109,15 @@ export function seedInstructionMaps(
...
  *
- * Known limitation: when a batch omits `childrenInstructions` (common when only
- * state changed), the synthetic element is given an empty `children` array. A
- * fixup that consults children (currently only `sectionJsToRust`, which removes
- * a `titleChild` and clears `xrefLabel.label`) can therefore compute a slightly
- * wrong `changedState` for such an update. This does not affect the components
- * this milestone drives (e.g. `textInput`); a correct fix needs the last-known
- * children retained per component, which belongs with the deferred prototype
- * update wiring where `section` updates are actually exercised.
+ * Note on children-dependent fixups: when a batch omits `childrenInstructions`
```
siefkenj: `batch`?

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
```diff
@@ -109,14 +109,15 @@ export function seedInstructionMaps(
...
+ * (common when only state changed), the synthetic element is given an empty
+ * `children` array, so a fixup that consults children operates without them.
```
siefkenj: `fixup`?

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
```diff
@@ -109,14 +109,15 @@ export function seedInstructionMaps(
...
+ * The one such fixup, `sectionJsToRust`, no longer depends on the children
+ * being present for its `changedState` output: it clears `xrefLabel.label`
+ * whenever the section has a title child (keyed off `titleChildName`), and only
+ * the optional splice of the title child out of `element.children` is skipped
+ * when children are absent — which is harmless, since an update that does not
+ * carry `childrenInstructions` also does not replace the consumer's children.
```
siefkenj: I don't understand what this comment is trying to say.

packages/doenetml-worker/src/jsCoreActionNames.ts
```diff
@@ -0,0 +1,43 @@
...
+ * rust core's action names) with the names the JavaScript core registers.
+ *
+ * The prototype renderers were written against the rust core, so most action
+ * names already match (e.g. `textInput`'s `updateValue` /
+ * `updateImmediateValue`). The graph components are the exceptions: the rust
+ * core (and thus the prototype) uses `move` / `changeBoundingBox`, while the JS
+ * core registers `movePoint` / `changeAxisLimits`. The args are compatible in
+ * both cases (JS `movePoint` accepts `{x, y, z}`, a superset of `{x, y}`; both
+ * graph actions take `{xMin, xMax, yMin, yMax}`), so only the name needs
+ * translating.
```
siefkenj, line 12: Differences in names should only be temporary. When we've decided on names, there will only be _one_ name in both cores (though since the JS core will eventually be replaced by Rust, the old name may stick around internally for a while).

packages/doenetml-prototype/dev/main.tsx
dqnykamp: Done in c1ea33407 — the dev toolbar/layout styles now live in a new `dev/main.css`.

packages/doenetml-prototype/src/renderers/doenet/m.tsx
dqnykamp: Agreed this is transitional. I've added a note to the doc block calling it temporary scaffolding: the real fix is making the cores agree on the props each component exposes (likely normalized in the JS→Dast conversion) so the renderer can assume one representation. Until that prop reconciliation lands, the renderer tolerates both forms. (f50136355)

packages/doenetml-prototype/dev/main.tsx
dqnykamp: Fixed in f50136355 — `CoreType` is now exported from the prototype lib's `index.ts` and imported from there in `dev/main.tsx` instead of being redefined.

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
dqnykamp: Reworded in f50136355 to drop the "bridges" metaphor. It now states plainly: the JS core doesn't return updates from an action — it pushes them through a callback, and `dispatchActionJavascriptFlat` collects those pushed updates into the same `Record<componentIdx, update>` map the rust core returns.

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
dqnykamp: Good call — replaced the long justification with a terse `XXX (June 2026)` note: `dispatchActionJavascriptFlat` returns `Record<number, FlatDastElementUpdateFromJS>`, which is shaped like `ActionResponse["payload"]` but is a distinct type; remove the cast once they share a type. (f50136355)

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
dqnykamp: Renamed in f50136355 — "batch" → "update instructions" throughout the doc/comments in this file.

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
dqnykamp: Renamed in f50136355 — "fixup" → "converter" throughout the doc/comments in this file.

packages/doenetml-worker/src/flatDastUpdateFromJS.ts
dqnykamp: Rewrote it in f50136355. It now reads: a state-only update carries no `childrenInstructions`, so the synthetic element gets an empty `children` array; the only converter that consults children (`sectionJsToRust`) still produces correct `changedState` from `titleChildName` alone, and skipping the title-child splice is harmless because such an update doesn't replace the consumer's children.

packages/doenetml-worker/src/jsCoreActionNames.ts
dqnykamp: Agreed — added a note that this whole module is temporary scaffolding: the differing names are an artifact of the cores being developed separately, and once names are unified there will be a single name in both cores (the old JS name may linger internally since the JS core will eventually be replaced by Rust), at which point this module can go away. (f50136355)

packages/doenetml-worker/src/jsRustConversions/section.ts
dqnykamp: Fair pushback — let me give the context, since this lives inside an existing hack.

The whole `sectionJsToRust` function is a best-effort reconstruction of the rust core's `xrefLabel`. The rust core produces a structured `xrefLabel` (the section *type*/number, e.g. "Section 1", rendered as `displayName`) separately from the title child. The JS core doesn't expose that structure, so `sectionJsToRust` guesses it from the title string (see the `// Hack to try to get xrefLabel info...` comment above). The `Division` renderer then renders `{displayName} {title}`.

The duplication happens because, when there's a title child, the guessed `xrefLabel.label` ends up holding the same text the title child already renders → "My cool section My cool section". The line you flagged clears `xrefLabel.label` in exactly that case so the renderer's `displayName` doesn't collide with the title child.

So the renderer *is* doing the right thing already — the defect is that the JS→Rust conversion feeds it a bad `xrefLabel`. Options:

- **A. Keep the worker-side clearing (current).** It's part of the existing `sectionJsToRust` hack; minimal and contained, but yes, it's a hack on top of a hack.
- **B. Move the dedup into the `Division` renderer** (suppress `displayName`/label when a title child is present). Cleaner per your "the renderer should know what to do" point, but it changes the renderer shared with the rust path, so it risks regressing rust rendering, and it papers over the real issue (the bad `xrefLabel` from conversion) rather than fixing it.
- **C. File a follow-up issue** to give the JS core a proper structured `xrefLabel` (or otherwise normalize section props) so the entire `sectionJsToRust` guesswork — including this clearing — can be deleted. Keep the minimal fix for now.

My inclination is **C**: this is one more symptom of the JS/Rust prop mismatch that needs a real reconciliation pass, not a renderer tweak that only masks it. How do you want to proceed?

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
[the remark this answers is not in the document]
siefkenj: What is going on here? You cannot index an array with `-1`. Are you attempting to avoid a typescript error?

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,24 +4,42 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
     darkMode: string;
+    coreType?: CoreType;
 }) {
     const dispatch = useAppDispatch();
     const errors = useAppSelector(errorsSelector);
     const isInErrorState = errors.length > 0;
 
+    // Register the dark-mode watcher once. It adds a `matchMedia` listener
+    // that is never removed, so it must not run on every `source`/`coreType`
+    // change (that would register duplicate listeners).
```
siefkenj: Remove comment

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,24 +4,42 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
     React.useEffect(() => {
         dispatch(globalActions.watchForDarkModePreferenceChange());
-        dispatch(dastActions.setSourceAndStartWorker({ source, flags }));
-    }, [source]);
+    }, []);
+
+    // Re-run when the source, core, or any flag *value* changes. The parent
+    // rebuilds the `flags` object on every render, so depend on a serialized
+    // key of its values rather than its identity to avoid re-initializing the
+    // worker on unrelated re-renders while still picking up real flag changes.
```
siefkenj: Remove comment. This is too verbose. Every comment is a call of attention to the code reader. Comment the important things that are not obvious or must be done a very specific way. This is just explaining what useEffect does...

packages/doenetml-prototype/src/viewer/page-viewer.tsx
```diff
@@ -4,24 +4,42 @@ import { useAppDispatch, useAppSelector } from "../state/hooks";
...
+    const flagsKey = JSON.stringify(flags);
+    React.useEffect(() => {
+        dispatch(
+            dastActions.setSourceAndStartWorker({ source, flags, coreType }),
+        );
+        // `flags` is intentionally tracked via `flagsKey`, not by object identity.
```
siefkenj: Add the sentence here explaining why

packages/doenetml-worker/src/jsRustConversions/section.ts
siefkenj: I still don't understand what you are saying.

Do not add "hacks upon hacks". It is okay if a ref from JS core doesn't work correctly in the rust core. Add an issue for it and keep the code clean.

packages/doenetml-worker/src/CoreWorker.ts
```diff
@@ -638,8 +639,24 @@ export class CoreWorker {
 
             this._javascriptUpdateBuffer = [];
             this._capturingJavascriptUpdates = true;
-            const actionResult =
-                await this.javascriptCore.requestAction(actionArgs);
+
+            // Translate the rust/prototype action name to the JS core's name
+            // when they differ (e.g. point `move` -> `movePoint`). The component
+            // type is looked up from the lookup map seeded during the initial
+            // render; an unknown component or unmapped action passes through.
```
siefkenj: Too much comment

packages/doenetml-prototype/src/state/redux-slices/core/thunks.ts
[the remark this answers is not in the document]
dqnykamp: Changed to a less strange approach:
```
            const cachedWorker =
                workerCacheKey != null ? workerCache[workerCacheKey] : undefined;
```

packages/doenetml-worker/src/jsRustConversions/section.ts
dqnykamp: Reworked to remove both hacks (commit 0a7c2c3d0).

You were right that the displayed-`title` regex plus the later label-clearing was a hack on a hack. The root cause: the JS core already computes the section's heading prefix in `titlePrefix` (with all of its auto-name/auto-number rules baked in), so there was no need to reverse-engineer it from the rendered title string.

`sectionJsToRust` now maps directly: the heading is `titlePrefix` when there's a title child (which the renderer draws separately via `props.title`) and `title` otherwise — no regex, no clearing. This structurally prevents the duplicated-title bug, since the child's text is never copied into the heading label.

On the `xrefLabel`/ref point: agreed it shouldn't be hacked in. We only borrow `xrefLabel.label` to render the heading; real cross-reference (`<xref>`) resolution from the JS core isn't wired up, and I've filed #1318 to do that properly during the prop-reconciliation pass rather than faking it here.

Verified with `sectionTitleUpdate.cy.js` (passes on both cores) and the worker unit tests.

dqnykamp: OK, Copilot got a little eager and replied to this thread again. What happened is that, when I originally made this `section.ts` converter last year for the static version, I created a hack to grab the label for xrefs from titles that look like `Section 5`. Apparently, that trick wasn't robust for the dynamic case, as it caused titles to get duplicated, and Copilot did some tricks on top of that to remove that duplication.

For now, we just simplified the code by removing my original hack. I haven't checked what this does to references, but we just added the issue to address it properly later.

dqnykamp:
@siefkenj Does this now meet your approval?

siefkenj (approved):

PR #1322, opened 2026-06-15
Add prototype renderers

siefkenj:
This PR adds renderers to doenetml-prototype. The goal is to make it so these renderers can totally replace the current doetnet renderers.


dqnykamp:
Could you move the changes to `dev/main.tsx` to its own little PR? We'll presumably want to copy that for other dev servers, so it'd be easier if it were isolated.

siefkenj:
> Could you move the changes to `dev/main.tsx` to its own little PR? We'll presumably want to copy that for other dev servers, so it'd be easier if it were isolated.

Done. This PR now depends on #1339

packages/doenetml-prototype/src/renderers/doenet/_button.tsx
```diff
@@ -0,0 +1,83 @@
...
+        };
+        dispatch(coreActions.dispatchAction(action));
+    }, [dispatch, node.name, node.data.id, clickAction]);
+
+    const displayLabel = processLabelWithMath(
+        node.data.props.label,
+        node.data.props.labelHasLatex,
+    );
+
+    console.log(
```
dqnykamp: Remove this console.log

packages/doenetml-prototype/src/renderers/doenet/angle.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
```
dqnykamp, line 1: For the point renderer, we used the convention `graph-point.tsx` for the version inside the graph and `text-point.tsx` for the version outside the graph. Do we want to keep this convention or only make a special prefix for the graph versions? To keep this convention, this should be `text-angle.tsx`; otherwise, we should rename `text-point.tsx` to `point.tsx`.

packages/doenetml-prototype/src/renderers/doenet/aside.tsx
```diff
@@ -0,0 +1,134 @@
+import React from "react";
```
dqnykamp: Is there a reason to keep `aside.tsx` separate from `division.tsx`? I notice that `division.tsx` now has a `collapsible` prop, which I thought was the difference between an aside and other sections (though it makes sense to add a collapsible attribute to all sections).

dqnykamp:
```
<booleanInput>
  <label>Hello there!</label>
</booleanInput>
```
is rendered with the label `!Hello there`.

dqnykamp:
> is rendered with the label `!Hello there`.

I had to ask AI to figure out what is going on here. Here's what Claude said:

The culprit is this rule in `boolean-input.css`:
   
   ```css
   &.input-label-right {
       direction: rtl;
       input[type="checkbox"] {
           direction: ltr;
       }
   }
   ```
   
   When `labelPosition` is `"right"`, the layout trick sets the whole `<label>` to `direction: rtl` so the checkbox (which comes *after* the text in DOM order) gets visually reversed to the left. The checkbox is then reset to `direction: ltr` so it doesn't render mirrored.
   
   **The problem:** the text span `.input-label` is *not* reset to `ltr`, so it inherits `direction: rtl`. The letters in "Hello there" are strong LTR characters, so they stay in order — but the trailing `!` is a **neutral/weak** character under the Unicode bidi algorithm. In an RTL context, trailing neutral punctuation gets reordered to the *logical end*, which is the left side. Hence `!Hello there`.
 
   
   ## Fix options
   
   **Option A — reset direction on the text span (minimal):**
   
   ```css
   &.input-label-right {
       direction: rtl;
       .input-label {
           direction: ltr;
       }
       input[type="checkbox"] {
           direction: ltr;
       }
   }
   ```
   
   **Option B — avoid the `rtl` hack entirely, use flexbox (more robust):**
   
   ```css
   label {
       display: inline-flex;
       align-items: center;
       &.input-label-right {
           flex-direction: row-reverse;
       }
   }
   ```
   
   This reverses the visual order without touching text directionality, so punctuation in any label renders correctly.
   
   I'd recommend **Option B** — the `direction: rtl` approach will bite any label ending in punctuation (`?`, `.`, `:`, parentheses, etc.), not just `!`.

packages/doenetml-prototype/src/renderers/doenet/aside.tsx
siefkenj: I kept it different because of the different functionality, but I can merge if you want.

packages/doenetml-prototype/src/renderers/doenet/angle.tsx
siefkenj: At this point, I think it makes sense to have the text renderers unprefixed and the graph ones start with `graph-`

siefkenj:
@dqnykamp Should be ready!

packages/doenetml-prototype/src/renderers/doenet/angle.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ResolvedStyleDefinition } from "@doenet/utils";
+import { _ServerSafeMath } from "./_server-safe-math";
+import { useTextRendererStyle } from "../../utils/use-renderer-style";
+
+type AngleData = {
+    props: { latex: string; selectedStyle: ResolvedStyleDefinition };
+};
+
+export const Angle: BasicComponent<AngleData> = ({ node, htmlId }) => {
+    const style = useTextRendererStyle(node.data.props.selectedStyle);
```
dqnykamp, line 12: The angle works fine with the Javascript core. If I try to render a document with `<angle>2pi</angle>` and select the rust core, then the viewer crashes with the error:
```
style.ts:1122 Uncaught TypeError: Cannot read properties of undefined (reading 'textColor')
    at textRendererStyle (style.ts:1122:29)
    at useTextRendererStyle (use-renderer-style.ts:8:12)
    at Angle (angle.tsx:12:19)
```
Do we care about that now? Ideally, such an error should bring down the whole viewer, but maybe that's OK for now.

siefkenj: I don't think we worry about Rust core right now.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
+    hr: { component: _PassThroughWithTag },
+    interval: { component: M },
+    ion: { component: M },
+    ionicCompound: { component: M },
+    lcm: { component: M },
+    li: { component: Li, passthroughChildren: true },
+    lq: renderAs("“"),
+    lsq: renderAs("‘"),
+    m: { component: M, passthroughChildren: true },
+    math: { component: Math },
```
dqnykamp: `math` can use `M` as well, as it behaves identically to components like `max`.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
-    xref: { component: Xref, passthroughChildren: true },
-    ol: { component: Ol, passthroughChildren: true },
+    triggerSet: { component: Button },
     ul: { component: Ul, passthroughChildren: true },
-    li: { component: Li, passthroughChildren: true },
-    em: { component: Em, passthroughChildren: true },
+    updateValue: { component: Button },
+    variance: { component: M },
+    wrapNumberPeriodic: { component: M },
+    xref: { component: Xref, passthroughChildren: true },
```
dqnykamp: Doenet does not have a `<xref>` component; the similar component is `<ref>`.

packages/doenetml-prototype/src/renderers/doenet/_button.tsx
```diff
@@ -0,0 +1,77 @@
...
+        const action: Action = {
+            component: node.name,
+            componentIdx: node.data.id,
+            actionName: clickAction,
+            args: {},
+        };
+        dispatch(coreActions.dispatchAction(action));
+    }, [dispatch, node.name, node.data.id, clickAction]);
+
+    const displayLabel = processLabelWithMath(
```
dqnykamp: In the legacy renderers, if no label is supplied, the label becomes "Button". It doesn't matter what word it says, but it should have some word so that the button looks like a button rather than thin line. Or we could add a minimum height.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
+    asList: { component: AsList, passthroughChildren: true },
+    atom: { component: M },
+    attr: passThroughWithRenamedTag("code"),
+    blockQuote: { component: BlockQuote, passthroughChildren: true },
+    boolean: { component: Boolean },
+    booleanInput: { component: BooleanInput },
+    br: { component: _PassThroughWithTag },
+    c: passThroughWithRenamedTag("code"),
+    callAction: { component: Button },
+    ceil: { component: M },
```
dqnykamp, line 171: I imagine choiceInput hasn't been looked at yet, but it does not work with the Javascript core.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
+    electronConfiguration: { component: M },
+    ellipsis: renderAs("…", { wrapInSpanWithId: true }),
+    em: { component: Em, passthroughChildren: true },
+    evaluate: { component: M },
+    extractMath: { component: M },
+    floor: { component: M },
+    function: { component: M },
+    gcd: { component: M },
+    graph: { component: Graph },
+    hr: { component: _PassThroughWithTag },
```
dqnykamp, line 192: At least on my computer, the `<hr>` is almost imperceptible.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
+    interval: { component: M },
+    ion: { component: M },
+    ionicCompound: { component: M },
+    lcm: { component: M },
+    li: { component: Li, passthroughChildren: true },
```
dqnykamp, line 197: The numbers and bullets are invisible to me.

packages/doenetml-prototype/src/renderers/renderers.ts
```diff
@@ -56,40 +74,177 @@ export type RendererObject = Record<
...
+    lq: renderAs("“"),
+    lsq: renderAs("‘"),
+    m: { component: M, passthroughChildren: true },
+    math: { component: Math },
+    matrix: { component: M },
+    max: { component: M },
```
dqnykamp, line 203: Already on core, we have these components tagged with `static rendererType = math`. Not for this PR, but it seems we could simplify this code by using `rendererType` rather than `componentType` for the component names in the flat dast. Then, you wouldn't need as large of a translation table.

packages/doenetml-prototype/src/renderers/doenet/text-input.tsx
```diff
@@ -5,14 +5,40 @@ import { useAppDispatch, useAppSelector } from "../../state/hooks";
 import { renderingOnServerSelector } from "../../state/redux-slices/global";
 import "./text-input.css";
 import { coreActions } from "../../state/redux-slices/core";
+import { _ServerSafeMath } from "./_server-safe-math";
```
dqnykamp, line 8: It doesn't need to be with this PR, but for consistency, we should rename `text-input` to `input` and `text-point` to `point` if we are changing conventions.

dqnykamp (changes requested):
Sorry, this took me so long. Just a set of minor issues.

packages/doenetml-prototype/src/renderers/renderers.ts
siefkenj: This is for PreTeXt compatability

packages/doenetml-prototype/src/renderers/doenet/_button.tsx
siefkenj: Added a min height

packages/doenetml-prototype/src/renderers/doenet/text-input.tsx
siefkenj: `text-input` is actually `<textInput>`, but I renamed `text-point`

packages/doenetml-prototype/src/renderers/renderers.ts
siefkenj: I think it's nice to have everything listed explicitly. Also, core shouldn't have to care about what renderers things use.

packages/doenetml-prototype/src/renderers/renderers.ts
siefkenj: Fixed

siefkenj:
@dqnykamp Should be ready when the tests pass

packages/doenetml-prototype/src/renderers/renderers.ts
dqnykamp: We could move the `rendererType` logic from core to the renderers. I always associated the two because it affected the choice of the `forRenderer` state variables. Since the rust core doesn't have the notation of `rendererType`, I guess we are moving in that direction.

packages/doenetml-prototype/src/renderers/doenet/text-input.tsx
dqnykamp: Oh, right, I over-generalized the `text-*` pattern!

packages/doenetml-prototype/src/renderers/renderers.ts
dqnykamp: The `inline` attribute does not yet work with choiceInput, but that's OK for this PR.

dqnykamp (approved):

PR #1323, opened 2026-06-16
Change default styling of an open <endpoint> to have an opaque interior

### LINKED ISSUE #184: Change default styling of an open `<endpoint>` to have an opaque interior

KhoiUna:
**What does this PR do?** Fixes the default styling of an open `<endpoint>` by ensuring its interior renders as opaque rather than transparent.

**Why was this PR needed?** Issue #184 reported that background grid lines and other graphic elements leak through open endpoints, causing visual clutter. Investigation of `point.tsx` revealed that the rendering logic lacked a fallback opaque fill rule for open symbols during both initialization and dynamic state changes.

**What are the relevant issue numbers?** Closes #184

Does this PR meet the acceptance criteria?:
- [x] Verified fix via manual testing.
- [x] Verified cross and plus marker styles remain unaffected.
- [x] Follows codebase styling guidelines.


dqnykamp:
Great, @KhoiUna, this is a good start. Indeed for a point that starts open, such as `<endpoint open>(2,6)</endpoint>`, it now looks good.

The next step is address the case where the openness of the point changes dynamically. The easiest way to see that is to add the `switchable` attribute to the endpoint. For example `<endpoint switchable>(2,6)</endpoint>`. Then, when you click on the point, it toggles between open and closed. For this example, since it starts closed, the endpoint will initialize with some intermediate opacity, as it should. But then, when you click it to make it open, the opacity does not change to `1`, so it still translucent.

To determine how to change it dynamically, you can look at the code starting on line 504 of `point.tsx`. There, you can find code that changes stroke and fill opacity when the state variable `markerOpacity` changes. You'll want to add your special condition to modify the fillopacity change there as well. That won't fix the `<endpoint switchable>` case, but it will fix cases where one changes `markerOpacity` in other ways.

The case invoked by the `switchable` attribute can probably be built as an `else if` from the condition starting on 504. To do this, considering making a ref like `previousUseOpenSymbol` so that `else if` can check if `useOpenSymbol` has changed from its previous value. If it has changed, then you'll need to update fill opacity.

Note: when modifying a jsxgraph property like `pointJXG.current.visProp.fillopacity`, you have use the lowercase term `fillopacity`. (You can initialize it as `fillOpacity`, but you have to change it as `fillopacity`. Go figure!)

dqnykamp:
Also, could you assign issue #184 to yourself?

Then, edit the PR description to give a short description of what the PR does. At the end of the description, add the line `Closes #184` so that github will automatically link this PR with that issue.

dqnykamp (changes requested):
See the above comments for the requested changes.

KhoiUna:
> Also, could you assign issue #184 to yourself?
> 
> Then, edit the PR description to give a short description of what the PR does. At the end of the description, add the line `Closes #184` so that github will automatically link this PR with that issue.

@dqnykamp I do not think I have the permission to assign the issue to myself.

dqnykamp:
> @dqnykamp I do not think I have the permission to assign the issue to myself.

I see. I assigned it to you.

KhoiUna:
@dqnykamp I have made some tweaks. Could you please review?

dqnykamp (approved):
Looks good!

KhoiUna:
Awesome!

dqnykamp:
Thanks for the PR!

KhoiUna:
You're welcome!

Issue #1334, opened 2026-06-18
[for-prototype] <booleanList> sugar


siefkenj:
`booleanList` in a renderable context could be wrapped in a `<asList>` when streaming out to FlatDAST. This would avoid the need to check weird things like `SV._compositeReplacementActiveRange` in the new renderers.

PR #1338, opened 2026-06-19
Wrap composite list replacements in <asList> parents in FlatDast

dqnykamp:
Addresses #1334, generalized beyond `<booleanList>`.

## Problem
The `doenetml` renderers (e.g. `p.tsx`) call `addCommasForCompositeRanges` to insert commas between the replacements a composite contributes to a parent, deciding where to do so from `SV._compositeReplacementActiveRange`. The `doenetml-prototype` renderers should not have to re-implement that check (issue #1334).

## Approach
Move the comma decision out of individual renderers and into the FlatDast that the JS core bridge produces for the prototype: when a renderable parent reports a `_compositeReplacementActiveRange`, wrap the relevant replacement range in a synthetic `<asList>` element. Nested composite groups are materialized as FlatDast wrappers only when needed for grouping; they are not pre-trimmed, matching the production helper's actual behavior for grouped nested composites. Whitespace-only replacement slots are treated as spacing around list items, not as separate comma-delimited items.

`_compositeReplacementActiveRange` is produced for **most multi-replacement composites**, not just `booleanList`, so this is implemented generally: under the conditions where a `doenetml` renderer would add commas via `addCommasForCompositeRanges`, the FlatDast now adds the `<asList>` parent. This holds for **both** the initial FlatDast and FlatDast updates. The prototype `<asList>` renderer and its registration are supplied by the already-merged renderer PR (#1322); this PR only changes the worker-side FlatDast wrapping logic and the prototype update reducer's upsert path.

## What changed
- **`compositeListWrapping.ts`** (new, `@doenet/doenetml-worker`): a port of the renderer recursion in `addCommasForCompositeRanges`. It emits an `<asList>` wrapper when an as-list range has multiple non-blank list items, and a passthrough `<_fragment>` only when a nested non-list composite must be grouped as a single list item so the enclosing list delimits it correctly. Top-level non-list composites are left inline. Wrappers borrow the composite's own (otherwise-unused) `componentIdx` as a stable id, and child slots stay aligned with the child-instruction index space (null placeholders preserved) so range indices line up.
- **`flatDastFromJS.ts`**: wraps ranges on the initial render, adding wrapper elements to `elements[]`.
- **`flatDastUpdateFromJS.ts`**: wraps on updates (before the JS→Rust fixups, so `section`'s title-splice stays index-aligned); ships each wrapper as a full element via a new `newFlatDastElement` field.
- **`doenetml-prototype` `dast/slice.ts`**: `processElementUpdates` upserts `newFlatDastElement`.

## Tests
- `compositeListWrapping.test.ts`: simple list, omitted `potentialListComponents` fallback, single item, blank-string spacing, non-eligible components, non-list composite, null-placeholder alignment, nested-grouping cases, and production-matching nested-wrapper blank handling.
- `flatDastUpdateFromJS.test.ts`: `<asList>` parent on the initial path and the update path (incl. the upsertable wrapper element), and absence of a wrapper for a non-list composite.
- `flatDastUpdateFromJS.coreIntegration.test.ts`: a **real** JS-core `<booleanList>` produces an `<asList>` parent in the initial FlatDast.

Validation:
- `CI=true npx vitest run --root packages/doenetml-worker` (35 tests pass)
- `npx tsc --noEmit -p packages/doenetml-worker/tsconfig.json`
- `npx prettier --check .changeset/composite-aslist-flatdast.md packages/doenetml-worker/src/compositeListWrapping.ts packages/doenetml-worker/src/compositeListWrapping.test.ts`

The production `doenetml` renderers are unchanged, since they do not use this FlatDast path.


packages/doenetml-prototype/src/renderers/doenet/as-list.tsx
```diff
@@ -0,0 +1,105 @@
+import React from "react";
```
siefkenj, line 1: I already wrote this in #1322 The don't need to be written again.

dqnykamp: Yes, the AI flagged adding this renderer as a feature in a round of revision, which made me realize I need to wait on merging #1322 before doing anything more here. Hope to have some time to review #1322 late today or tomorrow.

Issue #1332, opened 2026-06-17
[for-prototype] Pass tabular styling down to <cell>s


siefkenj:
Currently, the styling of a tabular cell happens at several levels. Some of the attributes, like the top border, are on the `table` itself, some are on the `td`, and some are on the `tr`. If they were all consolidated and passed down to cell, they could all be on the `td` using border collapse to make it look right.

PR #1339, opened 2026-06-19
Make the dev environment save the doenetml in localStorage

siefkenj:


packages/doenetml-prototype/dev/main.tsx
```diff
@@ -53,12 +86,20 @@ function App() {
...
+                </button>
             </div>
             <div className="dev-viewer">
                 <EditorViewer
-                    key={coreType}
-                    doenetML={doenetMLstring}
+                    key={resetKey}
+                    doenetML={resetKey === 0 ? initialSource : doenetMLstring}
                     coreType={coreType}
+                    onChange={saveSource}
```
dqnykamp, line 102: This `onChange` will never be called. You need to include the changes from #1332 to editor-viewer.tsx that call `onChange` when there is a change event.

dqnykamp (changes requested):

packages/doenetml-prototype/dev/main.tsx
siefkenj: #1332 seems like the wrong reference...

dqnykamp: Err, #1322

siefkenj: I think this can be left as a template and #1322 can just be merged. Considering the other dev environments don't use prototype and the onChange event already exists for them.

dqnykamp (approved):

PR #1407, opened 2026-07-02
fix: add textEdit ranges to close-tag completions for VSCode compatibility

dqnykamp:
Without explicit `textEdit` ranges, VS Code duplicates typed characters when accepting close-tag completions. Typing `</` inside `<text>` and accepting the completion `/text>` produced broken results such as `<text><//text>`, and after typing farther into the tag name (`</te|`) VS Code could keep using a stale replacement range and leave the already-typed suffix behind. The same stale-range behavior could also affect `<`-triggered snippet completions that stay visible while the user keeps typing.

## Root cause

The CodeMirror LSP plugin keeps track of the trigger start across keystrokes, but VS Code's native LSP client only asks the server for updated items while a completion list is marked incomplete. The close-tag item therefore needed both an explicit `textEdit`/`filterText` and a way to force the server to refresh the item as the user keeps typing. Snippet completions that replace the whole `<...` prefix need the same refresh behavior for their replacement ranges.

## Fix

**`packages/lsp-tools/src/auto-completer/methods/get-completion-items.ts`**

1. Close-tag completions now always carry explicit `textEdit` ranges in the relevant flows, including partially typed close tags, so accepting the item replaces the whole `</...` prefix.
2. Close-tag items set `filterText = "</tag>"` so they remain visible in VS Code as the typed prefix grows from `<` to `</` to `</te`.
3. Regular `<`-triggered snippets keep their `<`-prefixed `filterText`, and the dynamic prefigure `annotations-skeleton` snippet now uses the same filtering path so it stays visible as the tag name is typed.
4. The per-instance `<module copy|extend>` attribute refresh now runs only in the attribute-name completion branch, avoiding unrelated resolver work during close-tag completion requests.

**`packages/lsp/src/features/completions.ts`**

5. Completion responses that include a close-tag item or a `<`-triggered snippet item are now returned as `CompletionList { isIncomplete: true }`, so VS Code re-requests completions on subsequent keystrokes and gets an updated replacement range.

## Tests

- Updated `packages/lsp-tools/test/doenet-auto-complete.test.ts` assertions and snapshots for the explicit close-tag `textEdit`/`filterText` behavior.
- Added a focused dynamic-snippet regression in `packages/lsp-tools/test/doenet-auto-complete.test.ts` covering the prefigure `annotations-skeleton` filter behavior while typing `<ann`.
- Added focused `packages/lsp/test/completions-feature.test.ts` regressions covering the `isIncomplete` responses used by VS Code for both close-tag and `<`-triggered snippet completion refresh.
- Added a changeset describing the user-visible editor fix.

### DIFF
--- a/.changeset/vscode-close-tag-completion.md
+++ b/.changeset/vscode-close-tag-completion.md
@@ -1,0 +1,15 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Editor: Fix close-tag autocomplete in VS Code.
+
+When typing a closing tag in the editor (for example `</te|` inside `<text>`),
+the close-tag completion now stays in sync with the full partially typed prefix
+and accepting it replaces that whole prefix. This avoids VS Code/native-LSP
+flows that could previously duplicate the `/` or leave the already-typed suffix
+behind when completing a close tag.
--- a/.changeset/vscode-close-tag-completion.md
+++ b/.changeset/vscode-close-tag-completion.md
@@ -6,10 +6,14 @@
 "doenet-vscode-extension": patch
 ---
 
-Editor: Fix close-tag autocomplete in VS Code.
+Editor: Fix stale VS Code tag/snippet autocomplete ranges.
 
 When typing a closing tag in the editor (for example `</te|` inside `<text>`),
 the close-tag completion now stays in sync with the full partially typed prefix
 and accepting it replaces that whole prefix. This avoids VS Code/native-LSP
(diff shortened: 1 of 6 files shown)


dqnykamp:
Additional VS Code completion fixes added to this branch:

1. **Snippet completions disappearing after `<ans`** — VS Code uses the text in the `textEdit` range as the filter prefix, so the filter is `<ans` (not `ans`). Changed snippet `filterText` from `"answer-labeled"` to `"<answer-labeled"` so it starts with `<ans`. The codemirror LSP plugin strips the `<` before its own substring filter, so `"ans"` still matches there.

2. **Attribute name completions not auto-appearing** — VS Code only sends LSP completion requests on configured trigger characters. Typing letters inside an open tag never triggered suggestions. Added `" "` (space) to `triggerCharacters` so typing `<graph ` (after the space) immediately shows the attribute list.

PR #1421, opened 2026-07-05
feat: add positiveOnly/negativeOnly options for displayXAxis/displayYAxis

### LINKED ISSUE #355: feature request: options to display only positive or negative x and y axes
per office hours discussion, proposed change would be
`<graph displayXAxis="positiveOnly">`
to the existing `displayXAxis` and `displayYAxis` attributes, add the following specifications:
"full" / "true" (default)
"none" / "false"
"positiveOnly"
"negativeOnly"

musaddikchoudhury:
## What does this PR do?
Adds `positiveOnly` and `negativeOnly` as valid values for the `displayXAxis` and
`displayYAxis` attributes on `<graph>`, alongside the existing `full`/`true` and
`none`/`false`. An axis can now render only its positive or only its negative half.
The attribute becomes a multi-value text attribute and stays fully backward
compatible, including when bound to a `<booleanInput>` (e.g. `displayXAxis="$b"`).
A negative-only axis is defined toward the negative direction so its arrow head
stays inside the graph border, and its tick labels keep the correct negative sign.

## Why was this PR needed?
Closes #355. There was no way to display only half of an axis, which is useful for
math contexts like absolute value and distance. The attribute was a boolean that
only distinguished full/none, so any other value was treated as "hidden."

## What are the relevant issue numbers?
Closes #355

## Does this PR meet the acceptance criteria?
- [x] Tests added for new/changed behavior
- [x] All tests passing (full CI green)
- [x] Follows project style guide
- [x] No breaking changes (legacy true/false preserved)
- [x] Changeset added


dqnykamp:
I have not reviewed this PR yet, as it is still marked as draft. But, I did run the CI, and three jobs failed. Since you changed attributes, you'll need to rebuild the schema with `npm run build:schema --workspace packages/static-assets`. Let me know if you need any assistance with addressing the tests that now fail.

musaddikchoudhury:
Hi @dqnykamp — thanks for the earlier guidance and for running CI. I've addressed
the three failing checks (regenerated the schema, and fixed two boolean→text
axis-visibility checks in prefigure/graph.ts and useGridAndAxesSync.ts), synced
with the latest main, and added a changeset. Full CI is green on my fork and I
verified all four axis modes plus the booleanInput binding in the dev viewer.
Ready for review when you have time!

packages/doenetml-worker-javascript/src/core/StateVariableDefinitionFactory.ts
```diff
@@ -1667,6 +1667,31 @@ function validateAttributeValue({
...
     }
 
+    // Map a boolean-valued attribute source onto its text equivalent via
+    // `valueForTrue`/`valueForFalse`. Literal `"true"`/`"false"` authored
+    // strings are already mapped during DAST conversion (see
+    // `createInitialComponentFromAttribute`), but a value bound to a boolean
+    // component (e.g. `displayXAxis="$someBooleanInput"`) arrives here as the
+    // adapted text `"true"`/`"false"` and would otherwise be rejected by the
+    // `validValues` check below and fall back to the default.
+    if (
```
dqnykamp, line 1677: Good catch. I never realized that this didn't work, but now it does:
```
<booleanInput name="simp" />
<math simplify="$simp">x+x</math>
```

packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts
```diff
@@ -67,25 +67,28 @@ export default function useGridAndAxesSync({
             displayYAxisTicks: SVs.displayYAxisTicks,
         });
 
-        const displayXAxisChanged = SVs.displayXAxis
+        const showXAxis = SVs.displayXAxis !== "none";
+        const showYAxis = SVs.displayYAxis !== "none";
+
+        const displayXAxisChanged = showXAxis
```
dqnykamp, line 73: These conditions do not catch all the ways in which the x-axis display mode could change. It catches a change between "none" and any other mode, but not a change, for example, between "full" and "positiveOnly". Switching between axis display modes without going through "none" will lead to unexpected results.

To detect these changes, you could track the previous display-mode string in a ref  (like `previousXaxisWithLabelRef`) and remove/recreate the axis when it changes.

Here's some DoenetML that you can use to see the bug:
```
<choiceInput name="disp">
  <choice>full</choice>
  <choice>none</choice>
  <choice>positiveOnly</choice>
  <choice>negativeOnly</choice>
</choiceInput>

<graph displayXAxis="$disp">
  
</graph>
```
If you switch between "none" and any of the others, it works correctly. But, if you switch between the other modes, strange things happen.

packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
```diff
@@ -103,52 +103,74 @@ export function applyAxisTickHeights({
...
+        const scale = this.evalVisProp("scale") || 1;
+        const coord = tick.usrCoords[coordIndex];
+        if (Math.abs(coord) < 1e-12) {
+            return "0";
+        }
+        return this.formatLabelText(coord / scale);
+    };
+}
+
+export function createXAxis({
```
dqnykamp, line 134: The fact that you changed the order of `createXAxis` and `createYAxis` it makes the git diff much larger than it needs to be. If you change them back to the original order, that will create less churn and make it easier to see what actually changed.

(I have no idea why `createYAxis` is before `createXAxis`, as it makes more sense the order you made it. But, if we want to change that, we should do that in a separate PR that makes no other change so that it is easier to review what actually changed.)

dqnykamp (changes requested):
This PR looks well-constructed. Thanks for contributing it! And you fixed a latent bug in the process.

I did find a couple flaws. One is later comment. Another flaw is that, if the axis display is `negativeOnly`, the axis labels are sometimes in the wrong spots. For example, try the DoenetML:
```
<graph displayXAxis="negativeOnly" yLabelPosition="bottom" displayYAxis="negativeOnly">
  <xLabel>x</xLabel>
  <yLabel>y</yLabel>
</graph>
```

I had to use AI to figure out what is going on there. It isn't a flaw you introduced but a latent flaw the label position logic. It looks like you are using AI as well, so it can tell you as well as I can what how to fix it. If your AI can figure out a reasonable regression test for that, all the better.

musaddikchoudhury:
Hi @dqnykamp — thank you for the feedback! All three items from your review have been addressed, along with regression tests and a rebase onto latest `main`:

### 1. Fixed Axis Label Positioning for `negativeOnly` Axes
- **Issue**: For `negativeOnly` axes, custom label positions (e.g. `yLabelPosition="bottom"`, `xLabelPosition="left"`) placed labels at the wrong ends due to the reversed vector definition `[0, 0] -> [-1, 0]` / `[0, -1]`.
- **Fix**: Inverted the JSXGraph label anchor option mapping in both `packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts` and `useGridAndAxesSync.ts`:
  - `negativeOnly` X-axis: `xLabelPosition="left"` maps to `"rt"` (arrowhead), default (`"right"`) maps to `"lft"` (origin).
  - `negativeOnly` Y-axis: `yLabelPosition="bottom"` maps to `"rt"` (arrowhead), default (`"top"`) maps to `"lft"` (origin).
- **Verification**: Added unit tests in `packages/doenetml/src/Viewer/renderers/utils/axisLabelPlacement.test.ts` covering all anchor mappings for `full`, `positiveOnly`, and `negativeOnly`.

---

### 2. Dynamic Axis Mode Reconciliation (`useGridAndAxesSync.ts`)
- **Issue**: Switching dynamically between non-`"none"` modes (e.g. `full` ↔ `positiveOnly` ↔ `negativeOnly` via `<choiceInput>`) did not recreate the axis, causing rendering anomalies.
- **Fix**: Added `previousDisplayXAxisRef` and `previousDisplayYAxisRef` to track the active mode string. Whenever the mode transitions between any of the valid values, the JSXGraph axis is cleanly removed and re-instantiated. If the mode has not changed, in-place attribute updates are preserved.
- **Verification**: Added unit tests in `useGridAndAxesSync.test.ts` and tag integration tests in `packages/doenetml-worker-javascript/src/test/tagSpecific/graph.test.ts` cycling through all mode combinations.

---

### 3. Restored Function Ordering in `jsxgraph.ts`
- Restored `createYAxis` before `createXAxis` to eliminate unnecessary git diff churn, keeping the diff focused strictly on the feature changes.

---

### 4. Synced with Latest Main & Windows Worker Fix
- Rebased cleanly onto the latest upstream `Doenet:main`.
- Normalized static-copy paths for the worker bundle in `vite.config.ts` for cross-platform compatibility.

All unit, worker, and schema test suites are passing locally. Ready for re-review!

dqnykamp:
Sorry for the delay @musaddikchoudhury. Finally getting to this.

I asked my claude to review the changes. If found some flaws. I hope it is OK if I just have it post the comments directly rather than me summarizing. I'll do that now.

dqnykamp (commented):
Reviewed `71aa3a725` against merge base `dfe0f13`.

## Status of the three earlier review comments

| Earlier comment | Status |
|---|---|
| Axis labels in the wrong spots for `negativeOnly` | **Half-fixed.** The y-axis mapping is now correct, but the same inversion was applied to the x-axis where it was already correct, so `displayXAxis="negativeOnly"` is now broken the way the y-axis used to be. |
| Mode changes between non-`none` values don't recreate the axis | **Fixed in the production code** (`previousDisplayXAxisRef`/`previousDisplayYAxisRef` plus the removal). I traced the full/none/positiveOnly/negativeOnly transition matrix including the `drawZero` cross-recreation and found no hole. The regression test for it, however, is a no-op — see finding 3. |
| Restore `createYAxis` before `createXAxis` | **Fixed.** |

## Findings

**1. `packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts:363` — high. The x-axis label inversion for `negativeOnly` is wrong and regresses `xLabelPosition`.**

`const position = isLeft === isNegativeOnly ? "rt" : "lft"` assumes JSXGraph resolves `"lft"`/`"rt"` by `point1`->`point2` order. It doesn't. `JXG.Line.getLabelAnchor` (jsxgraph 1.12.2, `jsxgraphsrc.js:72133`) resolves them by comparing *screen x*:

```
case "lft": if (c1[1] < c2[1] + eps) use c1 else c2
case "rt":  if (c1[1] > c2[1] + eps) use c1 else c2
```

For a horizontal axis that makes `"rt"` the visually-right end regardless of point order, so no inversion is needed. The y-axis case genuinely is different: for a vertical line the two screen x values are equal, the `+eps` tie-break sends `"lft"`->`point1` and `"rt"`->`point2`, so the inversion at line 251 *is* correct.

Concrete failure: `<graph displayXAxis="negativeOnly"><xLabel>x</xLabel></graph>`. Default `xLabelPosition="right"` is documented as "Place the x-axis label at the right end of the axis"; for a negative-only axis that end is the origin. This now yields `"lft"` -> the clipped left border, and combined with `anchorx: "right"` and `offset: [5, 10]` the label lands right-aligned ~17px from the left edge and runs off-canvas. `xLabelPosition="left"` is likewise moved from the arrowhead to the origin.

Fix: revert to the pre-PR `isLeft`-only mapping for the x-axis and keep the new inversion for y.

**2. `packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts:144` — high.**

Same inverted x mapping in the in-place label-update path, so it also applies to a `negativeOnly` x-axis whose label changes without the axis being recreated. Line 185, the y-axis counterpart in the same file, is correct.

**3. `packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.test.ts:63` — high. The new regression test does not exercise the code it claims to test.**

The file imports `useGridAndAxesSync` at line 15 and never calls it. `runSyncHelper` is a hand-copied re-implementation of the effect body, so all three tests assert against the test file's own copy of the logic.

Verified empirically: with `useGridAndAxesSync.ts` reverted to `main`, all 3 tests still pass. This is exactly the behavior the earlier review asked for a regression test on, and it currently has none.

Two secondary tells that this isn't the real path: line 238 passes `displayXAxis: "positiveOnly"` in camelCase, a value the lower-cased attribute can never produce; and line 85 is a real `tsc` error, since `GraphSVs` requires `hidden: boolean`.

**4. `packages/doenetml/src/Viewer/renderers/utils/axisLabelPlacement.test.ts:213` — medium.**

These tests do call the real `createXAxis`, but they assert the wrong mapping from finding 1 (line 213 expects `"rt"` for left+negativeOnly, line 245 expects `"lft"` for default+negativeOnly), which locks the regression in. They also only check the option object handed to `board.create`, never what JSXGraph resolves that option to, which is where the bug lives.

**5. `packages/doenetml-worker-javascript/src/components/Graph.js:226` — medium. Undeclared breaking change to a public state variable.**

`displayXAxis`/`displayYAxis` are `public: true` and now yield the text `"full"`/`"none"` instead of booleans. `<p>$g.displayXAxis</p>` renders `full` where it used to render `true`, and any `<boolean>$g.displayXAxis</boolean>` or `<conditionalContent condition="$g.displayXAxis">` in an existing document changes behavior. The PR body's "No breaking changes (legacy true/false preserved)" holds for the *attribute*, not for the property. Worth calling out in the changeset.

**6. `packages/docs-nextra/pages/reference/graph2.mdx:84` — medium.**

The docs still say the attributes set rendering "to `true` or `false`", and the property table at line 373 still shows `$g.displayXAxis` without the new text type. Neither the new values nor the type change is documented, and this is the only place authors would learn `positiveOnly`/`negativeOnly` outside autocomplete.

**7. `packages/static-assets/src/generated/doenet-schema.json` — medium (merge readiness).**

The branch is 74 commits behind `main`, and `main` has since rewritten both generated schema files (~76k inserted lines) plus `Graph.js` (-366) and `StateVariableDefinitionFactory.ts` (+194). The PR's own schema diff is correct and minimal, but it will need regenerating after the rebase (`npm run build:schema -w packages/static-assets`) or the schema check will fail. GitHub reporting `MERGEABLE` only covers textual conflicts.

**8. `packages/doenetml/vite.config.ts:33` — low.**

The `normalizePath(...)` wrapper for the Windows worker-asset copy is unrelated to this feature and isn't covered by the changeset. It belongs in its own PR.

**9. `packages/doenetml-worker-javascript/src/test/tagSpecific/graph.test.ts:2479` — low.**

The comment "Initial choice is full" isn't right: no choice is selected initially, so `$disp` resolves to `""`, fails the `validValues` check, emits a `doenet-i0048` diagnostic, and falls back to the default, which happens to be `"full"`. The assertion passes for the wrong reason and would keep passing if `"full"` were removed from `validValues`. Selecting a choice explicitly before the first assertion would make it a real check.

## Checked and fine

- The `valueForTrue`/`valueForFalse` mapping in `validateAttributeValue` is correctly placed after `toLowerCase`/`trim` and before the `validValues` check, and `preprocessAttributesObject` already lower-cases `validValues`/`valueForTrue`, so the renderer's `"positiveonly"` comparisons line up.
- `grid` is unaffected, since it has no `createStateVariable`.
- `reverseTickLabelSigns` is sound: `getDistanceFromZero` really does sign by `point1`->`point2`, `evalVisProp`/`formatLabelText` exist on `Ticks` in 1.12.2, the equidistant-tick path passes `value === undefined` so dropping that parameter is safe, and the post-create `setMinorTicks` -> `fullUpdate()` -> `updateRendererLabels()` -> `setText` sequence means there's no wrong-signed first paint.
- The `drawZero` cross-recreation logic and `useAxisTickSpacingSync`'s identity-keyed signature both survive the new recreation path.
- `packages/doenetml-worker-javascript` graph + prefigure tests (87) pass, and prettier is clean.

Findings 1-3 look blocking to me; the rest are smaller.

---

dqnykamp:
Back to me again. Do you have bandwidth to address those changes, or at least the high priority ones?

Also, in my month delay, the `main` branch has moved a lot. If you can rebase on or merge in `upstream/main`, that'd be great. (You may need to regenerated the schema after that.) Otherwise, I can do that myself since you gave me permission to modify your branch.

Thanks again for creating this feature!

PR #1446, opened 2026-07-09
fix: coexist with a host-provided MathJax instead of clobbering it

### LINKED ISSUE #1433: DoenetViewer always injects its own MathJax and clobbers window.MathJax, racing host-loaded MathJax (breaks PreTeXt embeds)
## Context

Reported by PreTeXt users: Doenet activities embedded in PreTeXt books (e.g. Active Calculus) intermittently fail to load — the Doenet logo appears, then the iframe collapses (see companion issue #1434) or the activity never renders. Failures are load-order dependent (often reported as "Chrome-specific", but it is timing, not a browser bug). Thread: https://groups.google.com/g/pretext-dev/c/JAmt-9qhXuI

## Root cause

`DoenetViewer`/`DoenetEditor` wrap content in `better-react-mathjax`'s `<MathJaxContext config={mathjaxConfig} version={4}>` (`packages/doenetml/src/doenetml.tsx` ~lines 302, 499). `better-react-mathjax`'s loader **unconditionally**:

1. assigns `window.MathJax = config` (a plain config object), then
2. appends a `<script src="https://cdn.jsdelivr.net/npm/mathjax@4.1.0/tex-mml-chtml.js">` to `<head>`.

It never checks whether MathJax is already loaded or loading on the page. So when the host page loads its own MathJax (PreTeXt's generated `-if.html` pages load a **deferred, floating `mathjax@4`** script):

- If the host's MathJax initializes first, step 1 **overwrites the live MathJax engine object** with a config object, breaking anything that uses it.
- If Doenet's copy wins, the host's deferred script runs into an already-defined `MathJax` global with a different version.
- Either way two engines (each of which also spawns its own worker in v4) race, and which interleaving you get depends on network/cache timing — hence the intermittent, per-navigation-different behavior.

### Why this surfaced in July 2026

PreTeXt's pages load *floating* `mathjax@4`, which began resolving to **4.1.3 on 2026-07-03** — days before the failure reports — while Doenet pins `4.1.0`. Every published book silently changed MathJax versions without a rebuild. (`@doenet/standalone@latest` has been 0.7.20 since 2026-06-01, so the Doenet side was not the trigger.)

## Proposed fix

Make the viewer coexist with a host-provided MathJax:

1. **Detect an existing engine**: if `window.MathJax` exists and looks initialized (or a MathJax script tag is present/loading), reuse it instead of injecting a second copy — and never overwrite `window.MathJax` with a config object. `better-react-mathjax` has no built-in support for this, so it likely means wrapping/replacing its context provider with our own loader.
2. **Make the source configurable**: expose the MathJax URL (and/or a `useExistingMathJax` flag) through the viewer props / standalone `renderDoenetViewerToContainer` config / `data-doenet-*` attributes, so hosts like PreTeXt can control exactly one load.
3. Document the supported MathJax version range, since reusing a host engine means the host's version governs typesetting behavior.

Even absent a host conflict, the current behavior double-loads MathJax in embedding scenarios (extra engine + extra MathJax worker per iframe), so this also removes a fixed per-iframe memory/startup cost.

## Immediate host-side workaround (for reference)

Hosts embedding the standalone viewer should not load their own MathJax in the same document (PreTeXt: drop the MathJax `<script>` from the Doenet-only `-if.html` pages), and should pin exact versions rather than floating tags.

## Validation

A test page that loads MathJax v4 (deferred) plus a standalone viewer, reloaded repeatedly / with throttled network: the activity must render on every interleaving, `window.MathJax` must remain the host's engine, and only one MathJax engine may exist in the realm.

dqnykamp:
## Problem

`DoenetViewer` / `DoenetEditor` wrapped content in `better-react-mathjax`'s `<MathJaxContext>`, whose loader **unconditionally**:

1. assigns `window.MathJax = config` (a plain config object), then
2. appends its own `<script src="…mathjax@4.1.0/tex-mml-chtml.js">`.

It never checks whether the host page already loaded MathJax. So when a Doenet activity is embedded in a page that loads its own MathJax (e.g. PreTeXt books, whose generated `-if.html` pages load a **deferred** `mathjax@4`):

- if the host engine initializes first, step 1 overwrites the live engine object with a config object, breaking it;
- if Doenet's copy wins, the host's deferred script runs into an already-defined `MathJax` global;
- either way two engines (each spawning its own worker in v4) race, and the interleaving depends on network/cache timing — hence the intermittent, per-navigation failures reported.

Closes #1433.

## Fix

Replace that loader with one that **coexists** with a host-provided MathJax:

- **Reuse a live engine** if `window.MathJax` is already loaded — and never overwrite `window.MathJax`.
- **Wait for an existing/deferred MathJax `<script>`** instead of injecting a second copy.
- **Only inject our own** when the page provides no MathJax, staging `window.MathJax = config` **only** when that global is unclaimed.

The promise is memoized on `window`, so every viewer/editor/keyboard-tray in the realm shares a single MathJax (this also removes the previously double-loaded engine + extra worker per embedded activity).

New knobs, exposed on `DoenetViewer` / `DoenetEditor` props, the standalone `renderDoenet*ToContainer` config, and `data-doenet-*` attributes:

| Attribute | Prop | Meaning |
| --- | --- | --- |
| `data-doenet-mathjax-url` | `mathjaxUrl` | MathJax script URL to load when the page provides none |
| `data-doenet-use-existing-mathjax` | `useExistingMathjax` | Force reuse of a host MathJax even when it is not yet detectable |

Supported versions documented (MathJax 3.x–4.x for reuse; injects 4).

### Implementation

- `packages/utils/src/math/mathjax-loader.ts` — framework-agnostic singleton `loadMathJax()` with the coexistence logic.
- `packages/utils/src/math/MathJaxContext.tsx` (new `@doenet/utils/mathjax` subpath) — drop-in provider feeding the same `MathJaxBaseContext` the `<MathJax>` renderers consume.
- `doenetml.tsx` and `unique-keyboard-tray.tsx` use it in place of `better-react-mathjax`'s context. (Non-production harnesses — docs site, `test-main` files, `doenetml-prototype` — are intentionally left as-is; they don't ship in the standalone/iframe bundles.)

## MathJax version bump

A second commit bumps the bundled MathJax **4.1.0 → 4.1.3** — both the URL Doenet injects when a page provides none (`DEFAULT_MATHJAX_SRC`) and the VS Code preview's Content-Security-Policy allowlist. The `4.1.x` line is bug-fix only (4.1.3 fixes semantic-enrichment/speech infinite-loop crashes and a Safari `overflow: auto` rendering bug; 4.1.1/4.1.2 improved dark-mode contrast and accessibility; 4.1.2 corrected the LaTeX size macros). This also aligns the injected version with the floating `mathjax@4` tag that PreTeXt books load, so typesetting is consistent whether Doenet loads MathJax itself or reuses a host engine.

## Validation

- **9 unit tests** (`mathjax-loader.test.ts`) covering every branch: reuse-no-clobber, wait-for-deferred-script, self-inject, host-config-not-clobbered, single-promise memoization, `useExistingMathjax`, script-load-error rejection, and timeout rejection when a promised host MathJax never appears.
- **Real-browser (Playwright/Chromium)** reproduction of the issue's scenario: with a deferred host MathJax, Doenet reused the host engine, `window.MathJax` stayed the host's, exactly one engine loaded, and Doenet injected no script; with no host MathJax, it injected exactly one.
- Full `@doenet/standalone` build passes; bundle inspection confirms the shipped loader and that the old clobbering path is gone.

### DIFF
--- a/.changeset/bump-mathjax-4-1-3.md
+++ b/.changeset/bump-mathjax-4-1-3.md
@@ -1,0 +1,22 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Update the bundled MathJax from 4.1.0 to 4.1.3.
+
+Doenet now loads MathJax `4.1.3` (from 4.1.0) for the copy it injects when a
+page provides none, and the VS Code preview's Content-Security-Policy allowlist
+is bumped to match. The `4.1.x` line is bug-fix only: 4.1.3 notably fixes
+infinite-loop crashes in the semantic-enrichment/speech code, a Safari rendering
+bug for math in `overflow: auto` containers, and assorted TeX edge cases; 4.1.1
+and 4.1.2 improved dark-mode contrast and accessibility. This also aligns the
+version Doenet injects with what host pages that ship a floating `mathjax@4`
+tag (e.g. PreTeXt books) now load, so typesetting is consistent whether Doenet
+loads MathJax itself or reuses a host-provided engine.
+
+Note: MathJax 4.1.2 corrected the LaTeX size macros (`\large`, `\tiny`, etc.) to
+use standard LaTeX sizes.
--- a/.changeset/coexist-with-host-mathjax.md
+++ b/.changeset/coexist-with-host-mathjax.md
@@ -1,0 +1,37 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Viewer: coexist with a MathJax that the host page already provides.
+
+`DoenetViewer` / `DoenetEditor` previously wrapped content in
+`better-react-mathjax`'s `MathJaxContext`, which unconditionally assigned
+`window.MathJax = config` and appended its own MathJax `<script>` — with no
+check for a MathJax the host page had already loaded. When a Doenet activity was
+embedded in a page that loads its own MathJax (e.g. PreTeXt books), this clobbered
+the host's live engine with a plain config object and/or raced a second engine,
+causing intermittent, load-order-dependent failures to render.
+
+Doenet now loads MathJax through a coexisting loader: if a live MathJax engine
+is already present it is reused and `window.MathJax` is never overwritten; if a
+MathJax `<script>` is already on the page (including a deferred one) Doenet waits
+for it instead of injecting a second copy; only when no MathJax is present does
+Doenet load its own. This also removes the duplicate engine (and its extra
+worker) that was previously loaded per embedded activity.
+
+Two new controls are exposed on `DoenetViewer` / `DoenetEditor` (and, for the
+standalone build, as `data-doenet-mathjax-url` / `data-doenet-use-existing-mathjax`
+attributes and `renderDoenet*ToContainer` config keys):
+
+- `mathjaxUrl` — the MathJax script URL to load when the page provides none.
+- `useExistingMathjax` — force reuse of a host-provided MathJax even when it is
+  not yet detectable (e.g. the host loads it after Doenet mounts).
+
+Reusing a host engine means the host's MathJax version governs typesetting;
+MathJax 3.x–4.x are supported for reuse.
+
+Closes #1433.
--- a/packages/doenetml/src/doenetml.tsx
+++ b/packages/doenetml/src/doenetml.tsx
@@ -8,7 +8,7 @@ import React, {
     useState,
 } from "react";
 import { DocViewer } from "./Viewer/DocViewer";
-import { MathJaxContext } from "better-react-mathjax";
+import { MathJaxContext } from "@doenet/utils/mathjax";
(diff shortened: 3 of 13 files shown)


dqnykamp:
Added a second, separately-scoped commit that bumps the bundled MathJax **4.1.0 → 4.1.3** (`chore: bump bundled MathJax…`). It's on this PR rather than a standalone one because the injected default lives in `DEFAULT_MATHJAX_SRC`, a constant introduced by this PR's new loader — a separate branch off `main` couldn't reference it until this merges.

The bump updates two things that must stay in lockstep: the URL the loader injects, and the VS Code preview's CSP allowlist (`mathjax@4.1.0/` → `mathjax@4.1.3/`).

Why 4.1.3: it's bug-fix-only within `4.1.x` and notably fixes infinite-loop crashes in the semantic-enrichment/speech code and a Safari `overflow:auto` rendering bug. It also aligns the version Doenet injects with the floating `mathjax@4` tag PreTeXt-style hosts now resolve (4.1.3), so typesetting matches whether Doenet loads MathJax or reuses a host engine.

Validation (real Chromium, 4.1.0 vs 4.1.3, same expressions): all expressions typeset with no errors in both. The only behavior change is the documented 4.1.2 size-macro correction — measuring a single glyph's width: `\large`/`\Large`/`\huge` render a bit **smaller** (closer to true LaTeX), `\tiny`/`\small` a bit **larger**, and normal-sized math (the vast majority of content) is unchanged.

Issue #1441, opened 2026-07-09
Meta: reduce DoenetML memory footprint and startup time across embedding scenarios


dqnykamp:
Tracking issue for the coordinated effort to reduce DoenetML's memory footprint and startup time, especially for applications that embed many DoenetML documents at once (PreTeXt textbooks, Doenet assignment pages). Consolidates the analysis and links the individual issues across four repositories: DoenetML, assignment-viewer, DoenetApps, and math-expressions.

## Problem and measured baseline

Applications embedding DoenetML via `@doenet/standalone` / `@doenet/doenetml-iframe` pay a large per-instance cost, and pages with many instances reach multiple GB (#874 reports 1.7–3.6 GB on Active Calculus pages).

Measured with the checked-in harness (`packages/memory-benchmark`, added in #1442; PSS over the whole browser process tree, current 0.7.20 bundles). The `repeat-S` scenarios render **one** viewer with a generated `<repeatForSequence>` document of ≈4S components to expose document-scaling costs:

| Scenario | Total | Headline |
|---|---|---|
| Blank page | 207 MB | — |
| 1 viewer, small ~5-component doc (shared realm) | 552 MB | first instance ≈ +345 MB |
| 4 viewers (shared realm) | 944 MB | ≈ +131 MB per additional instance |
| 4 viewers (one iframe each, as doenetml-iframe/PreTeXt do) | 1,153 MB | ≈ +200 MB per additional instance |
| 1 viewer, `repeat-25` (~100 components) | 766 MB | ≈ +214 MB over the small doc |
| 1 viewer, `repeat-250` (~1000 components) | 2,046 MB | ≈ **+1,494 MB** over the small doc |

Where a ~130–200 MB small-document instance goes:
1. **~120 MB: the dedicated core worker** — parses the 11 MB worker bundle (≈4.9 MB of it base64 WASM), compiles WASM, initializes math-expressions/mathjs, and runs both the JS and Rust cores. One per instance, no sharing.
2. **~50–70 MB: the iframe realm** (iframe embedding only) — re-parses the 28.9 MB standalone bundle (which embeds the entire worker bundle as a string), 5 eager schema decompressions, its own MathJax engine + MathJax worker, 3.3 MB CSS.
[...]

PR #1465, opened 2026-07-11
Externalize the @doenet/standalone core worker to cut per-embed memory

dqnykamp:
_#1464 has merged; this branch is rebased directly onto `main`, so the whole diff belongs to this PR._

## Summary

`@doenet/standalone` embedded the entire ~15 MB core worker as an inline string, so every embedded viewer realm on a page held its own copy. This ships the worker as a separate `doenetml-worker/` directory next to the bundle and loads it from there via a new `@doenet/doenetml/doenetml-external-worker` entry.

The deployments that motivate #1441 both load `@doenet/standalone` **cross-origin from jsdelivr** — doenet.org via `@doenet/doenetml-iframe` (a `srcdoc` iframe), and PreTeXt via a same-origin `…-if.html` page that `<script src>`s the CDN bundle and calls `renderDoenetViewerToContainer`. In both, `new Worker(crossOriginUrl)` is blocked. The entry handles this by wrapping the worker in a tiny **same-origin `importScripts` bootstrap** (allowed for the classic worker `createCoreWorker` uses): the realm holds ~100 bytes instead of ~15 MB, while the worker code loads in the worker thread. Because the worker is resolved relative to the bundle URL, it always matches the bundle's version — no drift even with PreTeXt's floating `latest` tag.

When the bundle is instead loaded from a blob/data URL (the `@doenet/doenetml-iframe` component tests, `src/test-main.tsx`), "next to the bundle" doesn't exist: the entry falls back to `<origin>/doenetml-worker/index.js` against `document.baseURI` (the legacy `getWorkerUrl()` convention), and finally leaves the `global-config` default in place — so worker-URL resolution can never break evaluation of the whole bundle.

## Result (`packages/memory-benchmark`, cross-origin scenario, paired same-machine runs)

| cross-origin scenario | inline (before) | external (this PR) | Δ |
|---|---|---|---|
| iframe-xorigin-1 | 571 | 500 | **−71 MB** |
| iframe-xorigin-4 | 1204 | 1090 | **−114 MB** |
| marginal / added iframe | ~211 | ~197 | −14 MB |
| main-realm JS heap | 82 | 51 | −31 MB |

~70 MB less per embedded instance, realized on the actual cross-origin CDN deployment (a 20-instance textbook page saves >1 GB). The bundle itself also drops ~28 MB → ~13 MB.

## Compatibility

- **`@doenet/doenetml-iframe` benefits automatically** — it loads `@doenet/standalone`, and the measured cross-origin scenario *is* a `srcdoc` loading standalone cross-origin. No change needed there.
- **New requirement:** the `doenetml-worker/` directory must be served alongside the bundle. Automatic on npm/CDN (jsdelivr serves the whole package) and normal `dist/` deploys; a host that serves only `doenet-standalone.js` in isolation must now also serve `doenetml-worker/` next to it. The **`doenetml-inline-worker.js` entry remains available** for fully self-contained embeds.
- CSP unchanged for the cross-origin CDN deployments — the external path still creates the worker from a blob (the same-origin `importScripts` bootstrap). (A same-origin deploy loads the worker from a real `<origin>/doenetml-worker/index.js` URL, so it needs `worker-src 'self'` rather than only `blob:`.)

## Verification

- Added an `iframe-xorigin-*` benchmark scenario (a second CORS-enabled origin) modelling the CDN deployment; all viewers boot and initialize cross-origin, and the numbers match the same-origin externalized case.
- Added `workers-parse-*` / `workers-init-*` benchmark scenarios that boot N raw core workers with no document, isolating the per-worker fixed floor a shared/multiplexed worker (#1466) would collapse to one copy per page; plus an `--only <regex>` scenario filter for running a subset.
- The `@doenet/doenetml-iframe` Cypress component specs load the bundle from a blob URL and thus exercise the fallback path (worker served same-origin by the dev server); 14/14 pass.
- Full build green; worker files co-located in `dist/doenetml-worker/`.

Part of the memory-reduction effort (#1441), stream C. Background on the deployment mechanics is in [this #1441 comment](https://github.com/Doenet/DoenetML/issues/1441#issuecomment-4942063110).


dqnykamp:
## CI failure root-caused and fixed; benchmark floor scenarios added

**The `doenetml-iframe-cypress` failure was a real bug in this PR, now fixed** (`13e870b` + the middleware). The component specs load the standalone bundle from a **Blob URL** (`test/cypress/component/helpers.ts`, same trick as `src/test-main.tsx`). The external-worker entry resolved the worker with `new URL("./doenetml-worker/index.js", import.meta.url)` at module top level — and relative resolution against a `blob:` base **throws**, killing evaluation of the entire bundle. That's why the 7 boot-dependent specs failed while the 2 style-only ones passed.

Fix (product, not just tests): worker-URL resolution is now a chain that can never break bundle evaluation —

1. next to the bundle (`import.meta.url`) — the normal served-URL case, unchanged;
2. on an opaque base (blob:/data:), fall back to `<origin>/doenetml-worker/index.js` against `document.baseURI` (in a srcdoc iframe this inherits the embedding page's base URL — same convention as the legacy `getWorkerUrl()` default);
3. on total failure, keep the global-config default (never throw).

The Cypress dev server now serves the built worker at `/doenetml-worker/`, so the fallback resolves same-origin in the tests: **14/14 tests pass locally**, with the previously-failing specs running in ~2 s instead of timing out. The changeset documents the blob-URL host requirement. Benchmark re-check after the fix: `iframe-1` / `iframe-xorigin-1` boot and match prior numbers.

**Also pushed** (`0f0bf74`): `workers-parse-N` / `workers-init-N` benchmark scenarios that measure the idle-worker fixed floor — the sizing measurement for stream E, now filed as #1466. Headline: the per-worker floor is **~104 MB/instance** (53% of the iframe marginal), of which only ~1 MB is WASM (V8 already shares the compiled module across a page's workers). Details in [this #1441 comment](https://github.com/Doenet/DoenetML/issues/1441#issuecomment-4942631708).

PR #1478, opened 2026-07-12
Preserve preview window scroll position across source refreshes (VS Code extension)

anuragkatyal:
## Problem

In the VS Code extension, every refresh of the Doenet Preview panel — saving the document, pressing **Force Refresh**, or switching to another Doenet editor — rebuilds the rendered activity and resets the preview's scroll position to the very top. When authoring a long activity, you lose your place on every save.

## Fix

When the webview receives a `setSource` message, it records the preview container's scroll position and re-applies it on each animation frame while the viewer tears down and re-renders, clamped to the current content height (the viewer rebuilds asynchronously, so content collapses and grows back over an unknown time span). Restoration stops as soon as:

- the position sticks and the window elapses (4 s), or
- the user scrolls or interacts with the preview (wheel / touch / mouse / keyboard), so it never fights the user.

If the new document is shorter than the old scroll position, the view lands clamped at the bottom instead of jumping to the top.

All changes are confined to the preview webview (`src/preview-window/App.tsx`); no extension-host or viewer changes.

## Verification

Tested headlessly by driving the **built** preview-window bundle in Chromium (Playwright), posting `setSource` messages exactly as the extension host does:

| Scenario | Unpatched | Patched |
|---|---|---|
| Refresh after scrolling to 3023 px | resets to 0 | restored to 3023 |
| User wheel-scrolls during the restore window | — | user position wins; restore cancelled |
| New document much shorter than scroll position | — | clamps to bottom (no jump to top) |

Also verified: `tsc --noEmit` clean, Prettier clean.

## Notes

This is the first step of preview/editor position sync; scrolling the preview to the *edited element* (via `lsp-tools`' `elementAtOffset`) and click-to-navigate in both directions could build on top of this.

### DIFF
--- a/.changeset/preview-scroll-restore.md
+++ b/.changeset/preview-scroll-restore.md
@@ -1,0 +1,14 @@
+---
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+VS Code extension: Keep the preview window's scroll position when the source is refreshed.
+
+Previously, every refresh of the preview (saving the document, pressing Force
+Refresh, or switching editors) rebuilt the rendered activity and reset the
+preview's scroll position to the top. The preview now records its scroll
+position when new source arrives and re-applies it while the viewer re-renders,
+clamped to the content height (so a shorter document lands at its bottom rather
+than jumping to the top). Restoration stops as soon as the user scrolls or
+interacts with the preview, or after a few seconds.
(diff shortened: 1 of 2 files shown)


dqnykamp (approved):
Looks good.

Since the helper is all DOM with no VS Code coupling, we could easily reuse it for the CodeMirror editor (EditorViewer). We could create a follow-up to generalize it.

Issue #1560, opened 2026-07-26
Code the PreFigure curve piecewise-child warnings


dqnykamp:
Three warnings in the PreFigure curve conversion are the last messages in `packages/doenetml-worker-javascript/src/utils/prefigure` still reaching authors only in English. #1559 coded the other sixteen and deliberately left these, because they cannot be coded without first changing how curve conversion describes its own recursion.

## The messages

All three are in `parsedFormulaPiecesFromDefinition` (`packages/doenetml-worker-javascript/src/utils/prefigure/components/curve.ts`):

- `curve.ts:915` — `` `${warningPrefix}: ${pieceRole} piecewise child ${i + 1} failed to build curve pieces (type '${typeLabel}'); child skipped.` ``
- `curve.ts:916` — `` `${warningPrefix}: ${pieceRole} piecewise child ${i + 1} has unsupported function definition type '${typeLabel}'; child skipped.` ``
- `curve.ts:938` — `` `${warningPrefix}: ${pieceRole} piecewise child ${i + 1} has no pieces overlapping the active domain/bounds; child skipped.` ``

They are built inline with `diagnostics.push({ type: "warning", ... })` rather than through `pushWarning`, on purpose: `pushWarning` now takes a code, and routing these through it would hide them from the `lint:i18n` burn-down. Inline, they are counted among the constructions still holding a literal.

## Why they are blocked

`pieceRole` is an English fragment composed **recursively**. It starts as one of three words:

```
curve.ts:1140    pieceRole: "function",
curve.ts:1207    pieceRole: "x",
curve.ts:1217    pieceRole: "y",
```

and every level of piecewise nesting appends to the level above:

```
curve.ts:909     pieceRole: `${pieceRole} piecewise child`,
```

So the value reaching a message is `"function"`, `"x"`, `"y"`, `"x piecewise child"`, `"x piecewise child piecewise child"`, and so on, and the message then appends `" piecewise child "` plus an index once more.

Passing that to the catalog as an argument would nail the English into the record — the thing the migration exists to stop (see `packages/i18n/README.md`, and the "never pass an English phrase as an argument" note on #1518). A translator would receive a sentence with an untranslatable English noun phrase wedged into the middle of it, and no way to reorder around it, which for a language that puts the qualifier the other way round is worse than leaving the whole message in English.

[...]

Issue #1089, opened 2026-05-09
Schema lacks a way to exclude properties (state variables)


dqnykamp:
## Problem

The schema generator supports excluding components and attributes via `excludeFromSchema: true`, but there is no equivalent mechanism for properties (public state variables). As a result, properties that are defined for runtime needs but are not author-facing leak into the schema and surface in places authors see — most visibly the new context-sensitive help (#1085) and autocomplete (#1088) features.

A concrete instance: `<booleanInput>` (and other inputs that inherit from `_input`) expose a `collaborateGroups` property that is currently unused (the attribute itself is already marked `excludeFromSchema: true` in `Input.js`, but its companion state variable — created via `createStateVariable: "collaborateGroups"` — still appears as a property in the schema). Typing `$myInput.` in the editor offers `collaborateGroups` as a completion option even though there's nothing useful for an author to do with it.

## Proposed scope

- Honor `excludeFromSchema: true` when building properties from attributes that have `createStateVariable` set. This is the narrow case that covers `collaborateGroups` and similar attribute-derived state variables.
- Add a way to mark a directly-defined state variable (one not derived from an attribute) as `excludeFromSchema: true`, mirroring the attribute-level mechanism. This handles the broader case where a state variable is exposed to the runtime but should not appear in the author-facing schema.
- Audit existing components for properties that should be excluded under the new mechanism (some are likely lurking — there are ~18 attributes already marked `excludeFromSchema: true`, several of which have `createStateVariable`).

## Files involved

- `packages/static-assets/scripts/get-schema.ts` — `buildPropertiesForType()` is where exclusion should be applied.
- `packages/doenetml-worker-javascript/src/components/abstract/Input.js` — `collaborateGroups` definition.

## Related

- #1085 (context-sensitive help) — surfaces property descriptions in a side panel
- #1088 (autocomplete descriptions) — shows property completions with descriptions; this is where the leak is most visible

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #861, opened 2026-02-13
Internationalize DoenetML


cqnykamp:
Setup DoenetML such that it can be localized to other languages.

dqnykamp:
Two limitations in the style-description pipeline, both surfaced while reviewing #1604 and both predating it:

- #1605 — the border clause is appended with a hardcoded space, the one join the catalog does not own. Chinese renders `带圆点的蓝色填充圆 和粗边框` with a space it would not write.
- #1606 — adjectives receive a gender token but no grammatical role, so German and Russian must be correct in either `borderStyleDescription` or the embedded `mit einem …` / «с …» clause, and wrong in the other. `background` is wrong in both directions in German.

Neither blocks the seeded catalogs; the German, Russian and Chinese `content.ftl` files document the workaround at the point they make it and now cite these issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1520, opened 2026-07-20
i18n Phase 5: authored-document translation tooling (XLIFF sidecar)


dqnykamp:
Part of #861 (internationalization umbrella). This is **Phase 5**: tooling to translate *authored DoenetML documents* — the activity content itself — while preserving math, structure, variants, and scoring.

## Model decision: single source doc + sidecar catalog (recommended)

Two models were compared:

**A. Separate document per language** (plain copies, or a `doenet:` `extend` overlay). No new tooling and full freedom to restructure culturally — but translators must be DoenetML authors, drift is unmanaged (nothing flags the Spanish copy when the English source's questions change), and cross-language variant/answer equivalence is unverifiable.

**B. Single source doc + sidecar XLIFF catalog, merged at load time.** Translators edit *only prose* in a purpose-built web editor (translation platforms speak XLIFF natively) and cannot break markup; variants, scoring, names, and macros are identical across languages *by construction*; per-segment source hashes give precise drift detection ("3 segments need review"); the library hosts one activity + N catalogs.

**Recommendation: B as the primary model** — it is the only one compatible with a community-volunteer translation workflow and the only one that manages drift. A remains the honest escape hatch for genuinely divergent cultural adaptations (a fork, attributed via the existing `doenet:` extend multi-source provenance in `packages/parser/src/expand-external-references.ts`).

## Why this is feasible with existing infrastructure

- DAST (`packages/parser/src/types.ts`, xast-based) has cleanly separable `DastText` leaves — including inside attribute values — with `position` + `source_doc` provenance, distinct from elements, math, and macros.
- Variant selection is seeded/structural (by index, not text), so in-place text replacement preserves variant identity exactly.
- `packages/lsp-tools` already has `DoenetSourceObject` (offset↔position mapping, node-at-offset) and `doenet-to-markdown` (a prose-extraction prototype walking the DAST).
- Round-trip exists via `toXml()` + `prettyPrint()` (normalizing, not byte-preserving — hence merge operates on the DAST/offsets, not by reserializing untouched source).

### New package `packages/doenetml-translation` → `@doenet/doenetml-translation`

Node CLI + browser-safe merge runtime, depending on `@doenet/parser` and `@doenet/lsp-tools`.

[...]

PR #1566, opened 2026-07-26
Localize the words the core computes into a document

### LINKED ISSUE #1519: i18n Phase 4: localize component-generated content words
Part of #861 (internationalization umbrella). This is **Phase 4**: localize the remaining words that components generate as *document content* (computed in the worker, in `documentLocale`).

### Booleans
`packages/doenetml-worker-javascript/src/components/Boolean.js` and `BooleanInput.js` render literal `"true"`/`"false"`. Split value from display: parsed/serialized values stay `true`/`false` (they are DoenetML syntax); only the display text localizes via `content.ftl` (`boolean-true`/`boolean-false`). Text-answer matching against boolean values should accept both the localized and English forms.

### List joiners
~12 worker files hand-roll `" and "` / `" or "` when composing phrases (`BooleanOperators.js`, `PiecewiseFunction.js`, `Extrema.js`, `Line.js` equation-format messages, `utils/answer.js`, `FractionInput.js`, …). Replace with `Intl.ListFormat(locale, {type: "conjunction" | "disjunction"})` — built into every runtime, zero catalog burden, correct per-locale comma/conjunction conventions.

### Numbers in prose
`packages/doenetml-worker-javascript/src/utils/numberDisplay.js` does locale-independent formatting. Prose counts (attempt counts, list lengths) go through `Intl.NumberFormat`; **math-context numbers explicitly keep `en` formatting** (decimal point) until the deferred math-notation phase — enforced by a documented flag, not left implicit.

### Language-aware text components
- `<intComma>` → `Intl.NumberFormat` grouping.
- `<pluralize>` stays English-only initially (correct pluralization of arbitrary nouns in other languages requires per-language dictionaries — out of scope); it emits a coded warning (Phase 3 system) when used in a non-`en` document.
- `<lorem>` stays English/Latin by design.

### Default answer labels
`submitLabel` ("Check Work") and `submitLabelNoCorrectness` ("Submit Response") defaults in `packages/doenetml-worker-javascript/src/utils/answer.js` resolve from `content.ftl` in `documentLocale`. Author-supplied values always pass through verbatim — only *defaults* localize.

## Testing

- `en` output byte-identical (default-locale invariance keeps existing Vitest/Cypress assertions green).
- Unit tests per item with a non-`en` locale (`es`): boolean words, `Intl.ListFormat` joiners, grouped integers, localized default submit labels.

## Size / dependencies

**M.** Depends on Phase 0; the `pluralize` warning depends on Phase 3's code registry (can land with a plain warning first).

dqnykamp:
Closes #1519. Part of #861 (i18n Phase 4).

**#1564, which this was stacked on, has merged.** This is rebased onto `main` and is a single commit — the whole diff is this PR.

The families #1519 lists of words a component generates as *content* rather than as a message about content. Everything here answers to `documentLocale`, and English comes out character-for-character as before.

Not the last English the worker composes into a document, though — a sweep of the whole worker turned up four groups outside #1519's scope, none of them touched here: a section's auto-generated name and the `"<name> <number>: <title>"` frame around it (`SectioningComponent.js`), `<hint>`'s default "Hint" and `<solution>`'s "Solution"/"Answer", `<table>`/`<figure>`'s "Table 1"/"Figure 1", and `<paginator>`'s "Previous"/"Next"/"Page" defaults — which are structurally the same problem as `submitLabel` and would resolve the same way. Chemistry is a fifth and larger one: `[Invalid Chemical Symbol]`, and English nomenclature and element data that are a translation problem rather than a string-literal one. These want a Phase 5 issue rather than a bigger Phase 4.

## Booleans

`<boolean>` and `<booleanInput>` render "verdadero"/"falso" in a Spanish document. The **value** is untouched — `true`/`false` are DoenetML syntax, what an author writes, what `<award>` compares against, and what saved state holds — so only `text` moved. `text` is computed, never essential, so nothing localized is written into saved state and a document saved in one language reopens correctly in another.

That split means a boolean now has two spellings and both have to be accepted where one is read back in (`$b.text` bound to an input hands back the word the reader saw; an author's own `<updateValue newValue="true">` hands back the syntax). `booleanFromWord` takes either, which is why it exists rather than a `text === "true"` at each site. It lower-cases and does nothing else, exactly as the English-only comparison it replaced did, so `" true "` still names no boolean and is still refused — there is a test for that, because the alternative is a boolean that flips on whitespace it used to ignore.

## Default submit labels

An answer's button says "Revisar" instead of "Check Work"; likewise a section-wide check-work button. Only the *default* is translated — `submitLabel="Ready?"` is the author's own wording and passes through verbatim in every language, **including when it happens to match the English default**. `usedDefault` is the only thing that can tell those two apart, so the attribute stores its raw value beside the resolved one: the attribute writes a `…PreLocalize` state variable, and a state definition resolves the author-facing name from it.

That raw variable is **not public**, which is the same shape `showCorrectness` has used all along. Not public is the stronger statement than hiding the companion variable from the schema with `stateVarExcludeFromSchema`, which is what `colorCorrectness` used to do: that left `$a.colorCorrectnessPreliminary` resolving, a reference no author should be able to write. The schema output is identical either way, since a non-public state variable produces no property to hide.

`colorCorrectness` is moved onto the same footing here (it predates this PR, #1089), and the resolved values behind both correctness settings are now **public in their own right** — `$a.showCorrectness`, `$a.colorCorrectness` and their section equivalents report the value after any enclosing section's setting, hand-grading and the activity-wide flag are taken into account. Previously neither resolved name worked, while `$a.colorCorrectnessPreliminary` did. Those are the schema additions in this PR: `showCorrectness` and `colorCorrectness` properties on the 33 elements that carry the attributes. Nothing is removed. `stateVarExcludeFromSchema` now has no users, though the mechanism and its test remain.

`<document>` re-takes the two definitions with `ownLocale: true`: an ancestor dependency skips the component it runs on (that is what lets a nested `<document>` detect that it is nested), so a root `<document lang="es">` would otherwise label its own button against the host's locale. Tested for the root and for a `<document lang="es">` nested inside an English one, where the inner document, the answer inside it, and the answer outside it each get the language they sit in.

## `<intComma>`

Groups by the document's conventions — `25.236.501,35` in Spanish or German, `12,34,567` in Hindi.

It **re-punctuates rather than re-formats**: `formatDecimalString` carries the fraction digits across rather than reformatting them, because handing the whole literal to `Intl.NumberFormat` *under its default options* rounds to three fraction digits and drops a trailing zero — `"1000.50"` would come back `"1,000.5"` in English, where nothing was supposed to change. (`maximumFractionDigits` lifts the rounding but stops at 100 digits; carrying the fraction has no ceiling.)

Carried across, but not verbatim. Bengali, Persian, Nepali and Burmese count in their own digits, so both halves are transliterated into them — otherwise one number ends up written half in one script and half in another (`"২,৫২,৩৬,৫০১.35"`). Grouping is also uniform where CLDR is not: Spanish, Italian, Polish, Hungarian, Bulgarian, Slovenian, Latvian, Estonian and European Portuguese all leave a bare four-digit number ungrouped, though German does not, and `<intComma>` has grouped from a thousand up since it was written. Placing the separators by hand is what keeps that true — the locale is asked where a boundary falls, never whether this number gets one.

English is byte-identical, including a leading-zero run, because the separators are placed against a **pattern** read out of `Intl` — its group separator, its minus sign, its bidi marks, and how long a group is counting from the right — rather than by handing `Intl` the value. A formatter shown the value gets two things wrong. It normalizes a leading zero away, so `<intComma>007</intComma>` would render `7`. And it overflows to `∞` above `Number.MAX_VALUE`, so an integer of about 309 digits would stop rendering at all — where the old regex loop had no ceiling. Neither is a formatting question: where a locale puts its separators depends on how many digits a number has, not on which ones or on how many of them fit in a double.

So `0001234` still groups as `0,001,234`, `<intComma>007</intComma>` still renders `007`, and a 400-digit integer groups the whole way across. Checked four ways, three of them differentially:

- The whole `<intComma>` definition against the pre-PR regex loop, in English, over **266,927 inputs** — every code point below U+0300, structural cases (negatives, exponents, prose, a bare `.`, a trailing `.`, Arabic-Indic and astral digits), every all-zero string up to 500 digits signed and unsigned, zeros-then-digits for every combination up to 70 + 70, every length up to 1,200, 250,000 random literals weighted towards leading zeros and long fractions, and inputs up to 100,000 characters: **zero divergences.**
- `formatDecimalString` against an implementation that hands the decimal *string* to `Intl` directly — exact, but only where `Intl` is: **every one of the 242 locales this ICU has number data for**, at every length from 1 to 45 digits, signed and unsigned (21,780 comparisons), plus 267 locale/numbering-system pairs (`hanidec`, astral-digit `adlm`, `fullwide`, `nkoo`, India's secondary grouping, Arabic and Persian negatives) out to the 308 digits `Intl` can still represent (29,370 comparisons): **zero divergences.**
- Where `Intl` cannot be the oracle — leading zeros, and integers past `Number.MAX_VALUE` — 42,780 checks against two independent ones: leading zeros re-derived by formatting the same number with its zeros written as ones and mapping them back, and 309-to-5,000-digit integers checked digit-for-digit against their localized digits with the rightmost 300 grouped as `Intl` groups 300. **Zero divergences.**
- Mutation-tested: 19 mutants of `intl.ts` (grouping sizes read off the wrong end of the probe, a four-digit probe, code-unit instead of code-point digit counting, a hardcoded separator or decimal point, a dropped bidi prefix, an unanchored literal pattern, no tag validation) — **16 killed, 3 survivors, all three inert rather than untested**: shortening the probe from 16 digits to 8, `useGrouping: true` → `"auto"` (the probe is long enough that every locale groups it either way — what makes a four-digit number group is the placement loop, not this option), and dropping the `suffix` accumulator, which is empty in all 300 locale/numbering-system pairs checked and is kept only for symmetry with the bidi `prefix` that is not.

Only three locale grouping shapes exist across all 242: `3-3-3…`, India's `…2-2-3`, and one locale that groups in twos throughout. The 16-digit probe reveals all three.

## `<pluralize>` and `<lorem>`

`<pluralize>` runs an English part-of-speech model and cannot generalize — a correct plural elsewhere needs that language's morphology and often its dictionary. Outside English the model does not run.

`pluralForm` still does: that is the author supplying their own language's plural, and `basedOnNumber` chooses between the two forms by arithmetic, neither of which needs a model. Only a `<pluralize>` with no plural to fall back on passes its text through and raises `doenet-w0111` — which is what makes the warning's own advice ("set it with the `pluralForm` attribute") true.

`<lorem>` stays Latin in every language and needs no warning: nothing was asked for that wasn't delivered.

## Numbers in mathematics

`MATH_NOTATION_LOCALE` is where the policy is written down, so English formatting is a decision with a name rather than an omission. A decimal comma is real and wanted, but it has to arrive on the input side at the same time — until then it would change what a grader compares, not just how it looks.

It has no callers, and its doc comment says so: `Intl.NumberFormat` and `toLocaleString` appear nowhere in the worker, the viewer or `@doenet/utils`, so no math number reaches a locale-aware formatter and there is no site to pass a locale to. It is there for the phase that adds those sites.

## One scope correction

The issue lists "~12 worker files hand-roll `" and "` / `" or "`". **They are not in the tree.** The words the core composes into content itself are the three `<piecewiseFunction>` writes around a branch's domain — `\text{if } x < 1 \text{ or } x > 2`, and `\text{otherwise}` for the branch that catches what the others leave — and all three are localized here, since translating the conjunction alone would render one expression half in each language. Everything else that reads as "a, b, and c" is a *diagnostic*, and diagnostics were given `Intl.ListFormat` joining in Phase 3 via `DiagnosticListArg`.

`Intl.ListFormat` also turned out to be the wrong tool for `<asList>`: its `type: "unit"` is CLDR's measurement-unit list, so Russian and Japanese come back space-separated (`"a b c"`) and Chinese with no separator at all (`"abc"`). `<asList>` keeps its comma.

Relatedly, `listFormatFor` moves out of `diagnostics.ts` into a shared `intl.ts`, alongside the invalid-tag check `translator.ts` was keeping its own copy of.

## Documentation

The reference pages for the three components whose rendered output changes are updated: `intComma` (grouping follows the document's language, digits are regrouped and never rounded), `pluralize` (English-only, with `pluralForm` as the cross-language route), and `boolean` (`text` localizes, `value` does not).

The `KNOWN_UNTRANSLATED` list in the pseudo-locale sweep (`i18n.cy.js`) keeps its `Check Work` / `Submit Response` entries, with a corrected reason: those labels *are* translated now, but against `documentLocale`, and the sweep pseudo-localizes only `uiLocale` — the pseudo-locale is derived from the chrome catalog and the content side has no equivalent, so a `documentLocale` of `en-XA` negotiates straight back to English. Deleting those entries needs a pseudo content catalog, not another extraction.

## Verification

- `lint:i18n`: 297 keys, 184/184 codes covered, **2 constructions still holding a literal** — unchanged, still the deferred `<curve>` pair (#1560)
- New: 24 worker tests (`contentWordsLocale.test.ts`) and 17 `@doenet/i18n` tests (`intl.test.ts`), each item asserted in both `en` and `es`
- Green: worker suites matching `answer` / `boolean` / `booleanInput` / `contentWordsLocale` / `mathdisplay` / `piecewisefunction` / `document` / `sectioning` / `text` / `problem` / `pluralize` / `lorem` / `diagnostics` / `baseComponent` / `list` — 41 files, 691 passing, 11 skipped; `@doenet/i18n` 157; `@doenet/lsp` 46 (1 skipped); `@doenet/lsp-tools` 626; `@doenet/static-assets` 42; `@doenet/doenetml` 88
- Language-server bundle guard re-run after a rebuild: 1044 KiB of 1128 KiB, no message catalogs
- `doenet-schema.json` / `doenet-relaxng-schema.json` regenerated, and `build:schema` re-run afterwards is a no-op. No element or attribute is gained or lost; the properties gained are `showCorrectness` and `colorCorrectness` on the 33 elements carrying those attributes, described above. What moves: the `submitLabel` / `submitLabelNoCorrectness` property entries on the 33 elements that carry them lose `fromAttribute: true` and `groupName` and change position, because they are now defined as state variables rather than derived from a same-named attribute — the attribute keeps its own `groupName`, which is where the docs pipeline reads a property's group from anyway, and `fromAttribute` has no consumer. Plus the description changes this PR makes: `text` on the 14 components that inherit it from `<boolean>` (`<boolean>`, `<booleanInput>`, `<not>`, `<and>`, `<or>`, `<xor>`, `<iff>`, `<implies>`, `<isInteger>`, `<isNumber>`, `<isBetween>`, `<when>`, `<hasSameFactoring>`, `<matchesPattern>`), `<pluralize>`'s summary, and `<intComma>`'s summary and `value`.

### DIFF
--- a/.changeset/i18n-content-words.md
+++ b/.changeset/i18n-content-words.md
@@ -1,0 +1,23 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Translate more of the words the core computes into a document: boolean words, the default submit-button labels, and the `if`, `or` and `otherwise` a piecewise function writes around its branch conditions.
+
+A `<boolean>` or `<booleanInput>` in a Spanish activity now reads "verdadero" and "falso" where an author interpolates `$b.text` into their prose. The *value* is untouched: `true` and `false` are DoenetML syntax, so an `<award>` comparing against them, and saved state holding them, work the same in every language. Where a boolean is read back out of text — `$b.text` bound to an input — both spellings are accepted.
+
+An answer's submit button says "Revisar" instead of "Check Work" in a Spanish activity, and the same for a section-wide check-work button. Only the *default* is translated: `submitLabel="Ready?"` is the author's own wording and passes through verbatim in every language, including when it happens to match the English default.
+
+`<intComma>` groups by the document's own conventions rather than always in English — `25.236.501,35` in Spanish or German, `12,34,567` in Hindi. It still groups rather than rounds, so a value written with trailing zeros keeps them.
+
+`<pluralize>` works by running an English model over its text, and there is no equivalent for an arbitrary language. In a document written in another language it now leaves the text alone and says so, rather than silently doing nothing — unless the author supplied a `pluralForm`, which needs no model and is used in every language, with `basedOnNumber` choosing between the two forms. `<lorem>` stays Latin in every language, which is what placeholder text is for.
+
+Numbers inside mathematics keep `.` as their decimal separator in every language. A decimal comma is a real and wanted feature, but it has to arrive on the input side at the same time — until then, changing it would change what a grader compares rather than only how it looks.
+
+An answer's or section's `showCorrectness` and `colorCorrectness` are now properties an author can reference as `$a.showCorrectness` and `$a.colorCorrectness`, reporting the resolved values after any enclosing section's setting, hand-grading and the activity-wide flag are taken into account. The raw attribute values behind them, and the raw value behind a submit label, are no longer reachable under their internal names.
+
+With no locale configured, every one of these reads exactly as it did before.
--- a/packages/docs-nextra/pages/reference/boolean.mdx
+++ b/packages/docs-nextra/pages/reference/boolean.mdx
@@ -312,7 +312,10 @@ The `value` property of a named `<boolean>{:dn}` returns its value of `true` or 
 
 
 
-The `text` property of a named `<boolean>{:dn}` returns the text value of `true` or `false`.
+The `text` property of a named `<boolean>{:dn}` returns the text value of `true` or `false`,
+written in the language the document is in — `verdadero` or `falso` under `lang="es"{:dn}`.
+The `value` property does not change: `true` and `false` are DoenetML syntax, so an
+`<award>{:dn}` comparing against them behaves the same in every language.
 
 
 
--- a/packages/docs-nextra/pages/reference/intComma.mdx
+++ b/packages/docs-nextra/pages/reference/intComma.mdx
@@ -10,6 +10,12 @@ import { ComponentDisplay, AttrPropDisplay } from "../../components"
 `<intComma>{:dn}` is a [Math Operator](../concepts/essentialConcepts#component-types) 
 component which inserts commas into a large integer when rendering. 
 
+The grouping follows the language the document is written in, not always a
+comma: a document with `lang="es"{:dn}` or `lang="de"{:dn}` renders
+`25236501.35` as `25.236.501,35`, and one with `lang="hi"{:dn}` groups in twos
+above the first thousand (`12,34,567`). Digits are only regrouped, never
+rounded or dropped, so `1000.50` keeps its trailing zero.
+
 <AttrPropDisplay name='intComma'/>
 
 ## Examples
--- a/packages/docs-nextra/pages/reference/pluralize.mdx
+++ b/packages/docs-nextra/pages/reference/pluralize.mdx
@@ -18,6 +18,12 @@ displayed anywhere a text value is accepted. Pluralization uses the
 [`compromise`](https://compromise.cool/) NLP library and handles common
 English irregulars (`mouse → mice`, `goose → geese`, `pony → ponies`).
 
+**English only.** That model has no equivalent for an arbitrary language, so in
+a document written in another language automatic pluralization is skipped and a
+warning is raised. The [`pluralForm`](#attribute-example-pluralform) attribute
+still works in every language — supply the plural yourself and
+`basedOnNumber` will choose between the two forms.
+
 
 <AttrPropDisplay name='pluralize'/>
 
--- a/packages/doenetml-worker-javascript/src/components/Boolean.js
+++ b/packages/doenetml-worker-javascript/src/components/Boolean.js
@@ -5,6 +5,11 @@ import {
     returnChildrenByCodeStateVariableDefinitions,
 } from "../utils/booleanLogic";
 import { returnSimplifyExpandOnCompareWarning } from "../utils/answer";
+import { booleanFromWord, booleanWord } from "../utils/booleanWords";
+import {
+    contentTranslator,
+    returnContentLocaleDependencies,
+} from "../utils/contentLocale";
 
 export default class BooleanComponent extends InlineComponent {
     static componentType = "boolean";
@@ -411,7 +416,7 @@ export default class BooleanComponent extends InlineComponent {
 
         stateVariableDefinitions.text = {
             description:
-                'The boolean rendered as a text string ("true" or "false").',
+                'The boolean rendered as a text string ("true" or "false", in the document\'s language).',
             public: true,
             shadowingInstructions: {
                 createComponentOfType: "text",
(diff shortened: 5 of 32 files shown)


dqnykamp:
The four groups of English this PR's sweep turned up outside #1519's list now have issues of their own, rather than living only in #1566's description:

- #1572 — the section auto-name and title frame, `<hint>`/`<solution>`, `<table>`/`<figure>`, `<paginator>`. All four are the same shape as `submitLabel` and reuse the `usedDefault` + `…PreLocalize` idiom this PR built.
- #1573 — chemistry, which is a translation-data and English-morphology problem rather than a string-literal one, and is scoped separately for that reason.

Both are filed as Phase 4, not Phase 5 — #1520 is already Phase 5 and is a different piece of work. Added to the roadmap in #861.

Issue #1555, opened 2026-07-26
Decide how schema-derived diagnostics should relate to core's


dqnykamp:
Splitting this out of the diagnostics-flow analysis done while scoping #1549, because it is the one part of that analysis which is **orthogonal to the parser error stream** and therefore does not need to be resolved before the parser translation work.

## The inconsistency

The editor shows diagnostics in two places, and they draw from different sets:

| Producer | Hover / squiggle | Diagnostics panel |
| --- | --- | --- |
| Parser `DastError` nodes | ✅ `extractDastErrors` | ✅ echoed back through the core |
| Core `DiagnosticRecord`s | ✅ when positioned and the toggle is on | ✅ always |
| **Schema violations** | ✅ | ❌ **never** |

`validateTextDocument` merges `extractDastErrors(...)`, `getSchemaViolations()` and the worker-pushed `additionalDiagnostics`. The panel calls `mergeDiagnosticsByType({initialDiagnostics, diagnostics})`, which sees only what the core produced. So a schema violation draws a squiggle and a hover but never appears in the panel or its counts.

## Why it is not simply a matter of feeding schema violations to the panel

The schema is an **imperfect static approximation of what core actually accepts**, and the gap is structural rather than a matter of missing entries.

Core's child-matching generally ignores composites and validates children *after* composites have been expanded into their replacements. `getSchemaViolations` runs over the raw DAST, pairing each element with its literal parent — it has no notion of expansion, so a composite sitting between a parent and the children it will produce reads as a violation even though core will accept the expanded result.

The current mitigation is ad hoc: static flags on components (`excludeFromSchema` and friends), a per-instance attribute allowlist rebuilt via resolver round-trips for `<module copy=…>` sites (#1154), alias redirection so `<row>` inside `<matrix>` resolves against the right schema, and parser-layer filtering so unquoted attribute values do not double-report (#1197). Each was added to remove a specific class of spurious diagnostic that confused authors.

**So promoting schema violations into the panel would amplify the false positives** into a more prominent surface, which is presumably why they are not there. But the current situation is an accident of implementation rather than a stated contract, and the author-visible symptom — "the squiggle says something the panel doesn't" — is real.

## The gap runs both ways

The framing above treats the schema as an approximation that **over**-reports. It also under-reports relative to core in one direction and, more awkwardly, **catches things core never will**.

[...]

Issue #1518, opened 2026-07-20
i18n Phase 3: stable diagnostic codes and localized messages


dqnykamp:
Part of #861 (internationalization umbrella). This is **Phase 3**: give every DoenetML warning/error a **stable diagnostic code** and localize the messages. The code system delivers value independent of translation — codes become documentation anchors and stable identifiers for tests and tooling.

## Progress

The infrastructure has landed. What remains is the message migration, which `lint:i18n` counts on every run.

| | |
| --- | --- |
| Diagnostic constructions | **201** |
| Coded so far | **143** (#1547, #1550, #1551, #1552) |
| Uncoded constructions | **58** |
| …but see below | the counter **undercounts** by roughly 20 |

- [x] Code registry, append-only, enforced by `diagnostic-codes.lock.json` + `lint:i18n` (#1547)
- [x] Coded form alongside the legacy string form, so the migration is incremental (#1547)
- [x] Wire format carrying `{code, args}` plus the English (#1547)
- [x] Lint burn-down of legacy call sites (#1547)
- [x] `buildIgnoredPhrase()` replaced by Fluent plural selectors (#1547)
- [x] LSP populates `Diagnostic.code` (#1547)
- [ ] **Migrate the remaining messages** — the bulk of what is left
- [ ] `codeDescription` links and per-code documentation pages → **#1548**
- [ ] Parser diagnostics and the language server's locale → **#1549**

Rust core panic strings are developer-facing and stay English, as originally scoped.

## What "remaining" consists of

**The burn-down undercounts, and this is the thing to know before treating it as a finish line.** `lint:i18n` counts diagnostic *constructions* — an object literal with a `type` and a `message`. Three helpers construct diagnostics on behalf of their callers (`pushWarning` in the prefigure utilities, `pushVariantInfo` in the variant code, and the annotation warning helper), so the counter sees **3 constructions** where there are in fact **30 call sites carrying 29 literal English strings, in about 25 distinct shapes** — in `Select.js`, `SelectFromSequence.js`, `SelectPrimeNumbers.js`, `variants.ts`, and seven files under `utils/prefigure/`.

So of the 58 uncoded constructions:

- **28 are core-internal invariant messages** (`EssentialValueWriter`, `StateVariableInitializer`, `UpdateExecutor` — "can't update state variable X of component Y, as it doesn't exist"). **Decided: developer-facing, so not translated**, on the same footing as the Rust panic strings. They still get codes, which are worth having for tooling and dedup.
[...]

PR #1569, opened 2026-07-27
Give the language server's schema checks diagnostic codes

### LINKED ISSUE #1567: Schema violations from the language server cannot be translated
The schema violations `@doenet/lsp-tools` raises are author-facing — they draw the squiggle and the hover in the editor — but they compose English at the call site, carry no diagnostic code, and no locale can reach them. #1518 coded the worker's and the parser's diagnostics; this family was never in its inventory.

## What it looks like

With the document locale set to `es`, `<abc></abc>` produces two hover entries:

```
Error
Tipo de componente no válido: <abc>          ← core's doenet-e0002, translated

Warning
Element <abc> is not a recognized Doenet element.   ← schema violation, English only
```

The first is translated because the core raises it with a code and `DocViewer` re-renders it in the reader's locale. The second cannot be, because there is nothing to render.

(That two different checks both report one mistake is a separate question, tracked in #1555. This issue is only about the second line being untranslatable.)

## The messages

All in `packages/lsp-tools/src/auto-completer/methods/get-schema-violations.ts` — seven sites, six distinct sentences:

```
Element `<x>` is not a recognized Doenet element.
Element `<x>` is not allowed at the root of the document.
Element `<x>` is not allowed inside of `<y>`.
Invalid attribute name='…'. Names must start with a letter.
Element `<x>` doesn't have an attribute called `y`.      (two sites)
…and the allowed-values message, which branches on whether the attribute takes a list
```

## The constraint that makes this non-trivial

`@doenet/lsp-tools` is bundled into the language server, which `@doenet/codemirror` embeds verbatim and starts as a blob worker. It has no locale and renders nothing, so a message catalog there is dead weight on the editor's critical path — about 30 KB of FTL plus the Fluent runtime. `packages/lsp/scripts/check-server-bundle.mjs` rejects a catalog in that bundle outright rather than budgeting for one (#1557), and #1553 put a test around the boundary. `@doenet/lsp-tools` currently has no dependency on `@doenet/i18n` at all, not even a dev one.

## Suggested approach

#1564 solved this exact shape for `@doenet/parser`, which is under the same constraint. `codedDastError` takes the English **alongside** the code rather than rendering it, so the producing package needs no runtime catalog:

- `@doenet/i18n` enters as a **devDependency**: `import type` in `src/` so every code is checked against the registry at compile time, and the real package only in a test.
- That test parses a corpus, then asserts every coded diagnostic renders through the catalog to exactly the string the producer wrote. A message edited on one side and not the other fails there, as does a missing or misnamed argument.
- `dependencies` stays `{}` and the bundle guard stays green.

The same three pieces should transfer directly to `get-schema-violations.ts`.

## Worth deciding at the same time

Whether the codes belong to the existing `doenet-w…` sequence or want a prefix of their own — these come from a different producer than every code issued so far, and a host filtering on them may care.

Part of #1518. Related to #1555, which asks a different question about the same diagnostics.

dqnykamp:
**#1566, which this was stacked on, has merged.** This is rebased onto `main`.

Closes #1567. Part of #861 — the tail of what #1518 tracked as Phase 3, closed once the codes were all in place.

## What changed

The schema violations `@doenet/lsp-tools` raises are what the editor draws a squiggle and a hover from, and they were the last author-facing diagnostic family composing their English at the call site with no code on them. Each now carries a stable code and the values that fill its message in, and the same sentences are in the catalogs for translators.

`codedLspDiagnostic` (`packages/lsp-tools/src/coded-lsp-diagnostic.ts`) is the direct counterpart of `codedDastError` from #1564, and makes the same trade for the same reason: the caller passes the English **alongside** the code rather than having it rendered, because this package is inside the language-server bundle that `@doenet/codemirror` embeds and starts as a blob worker. `@doenet/i18n` enters as a **devDependency** — `import type` in `src/`, the real package only in the round-trip test — so `dependencies` stays free of it and no catalog reaches the bundle. `check:size` reports **1045 KiB of 1128 KiB, no message catalogs**.

`test/coded-schema-violations.test.ts` is what stops the two copies drifting: it runs the checker over broken DoenetML **against the real DoenetML schema** and asserts every coded violation renders, through the catalog, to exactly the string the checker wrote. A message edited on one side and not the other fails there; so does an argument the catalog reads and the checker doesn't pass, which renders as `{$name}`.

## The two decisions the issue asked to make

**A prefix of their own, or the existing sequence?** The existing sequence. A code names a *situation*, not the producer that spotted it, and #1564 already put the parser's codes in the same ranges as the worker's; a per-producer prefix would fragment the ranges while answering a question (`who raised this?`) that no consumer asks. Five new codes, `doenet-w0112`–`doenet-w0116`.

**Six sentences, six codes?** Five. Two of them collapse:

- The allowed-values message becomes one code with a Fluent selector on `$isList`, following `name-attribute-invalid`'s `$reason` precedent in the same catalog — one situation, two sentences, because the reader has to be told whether they are choosing a whole list.
- The "Invalid attribute name='…'. Names must start with a letter." check turns out to be *the parser's own message*, already coded as `doenet-e0025` by #1564. It reuses that code rather than taking a second name for one mistake. That is also a small win for #1555: `dedupeLspDiagnostics` keys on code + arguments, so the two reports of that mistake now collapse into one hover entry even once they stop being the same string.

The two unknown-attribute reports — a name the schema has never heard of, and one that exists elsewhere but not here — differ only in which spelling they quote back, so they raise `doenet-w0115` from one call site with the spelling picked where they differ. Three local helpers name the ranges a violation can point at (`offsetRange`, `elementTagRange`, `attributeRange`), which the six call sites had been spelling out in full. The file also loses four imports (`RowCol`, `CompletionItem`, `CompletionItemKind`, `showCursor`) that nothing in it had referenced.

The allowed-values list is handed over as a list (`{ list, type: "unit" }`) rather than a pre-joined string, so a language that does not enumerate with a bare comma isn't stuck with English punctuation. The producer still joins it with `", "` for the English it writes, and `type: "unit"` is exactly that join in English — which is what the round-trip test checks.

## What did not change

Every message reads exactly as it did before, in every existing test. The inline snapshots in `doenet-auto-schema-check.test.ts` gained `code` and `data.args` and nothing else.

## Not in scope

Rendering these in the editor's hover in the reader's language — that is #1571, stacked on top of this. The parser's copies get translated today only because the worker re-surfaces them as runtime diagnostics and `DocViewer` re-renders those before the dedupe keeps them; schema violations have no worker counterpart, so translating them at display time needs the CodeMirror LSP plugin to render through a formatter, which is a change in a different package. This PR is what makes that possible — before it there was nothing to render *from*.

## Verification

- `lint:i18n`: 302 keys across 2 locales, 258 coded call sites covering **189/189** codes; 164/166 constructions migrated, 2 still holding a literal (unchanged — the deferred `<curve>` pair)
- `@doenet/lsp-tools` 635, including `coded-schema-violations` 9
- `@doenet/i18n` 157, `@doenet/lsp` 46 (1 skipped)
- `npm run build -w @doenet/lsp` then `npm run check:size -w @doenet/lsp`: 1045 KiB of 1128 KiB, no message catalogs
- prettier clean

### DIFF
--- a/.changeset/i18n-schema-violation-codes.md
+++ b/.changeset/i18n-schema-violation-codes.md
@@ -1,0 +1,15 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Give the language server's schema checks stable diagnostic codes, so the squiggles the editor draws under an unrecognized element, a misplaced one, an unknown attribute, or a value outside its enumeration can be translated and cited.
+
+These were the last author-facing diagnostics composing their English at the point they were raised, with no name a bug report could quote or a host could filter on. Each now carries a code and the values that fill its message in, alongside the English it has always shown, and the same sentences are in the message catalogs for translators.
+
+The check for a name that does not start with a letter shares its code with the parser's identical check rather than taking a second name for one mistake, which also lets the editor collapse the two reports into one entry.
+
+Every message reads exactly as it did before.
--- a/package-lock.json
+++ b/package-lock.json
@@ -26530,7 +26530,9 @@
                 "mdast-util-to-markdown": "^2.1.2",
                 "mdast-util-to-string": "^4.0.0"
             },
-            "devDependencies": {}
+            "devDependencies": {
+                "@doenet/i18n": "*"
+            }
         },
         "packages/memory-benchmark": {
             "name": "@doenet/memory-benchmark",
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -304,6 +304,34 @@ migrate and no code to name; they spread `diagnosticCodeFrom` to pass along
 whatever their source carried. `lint:i18n` counts those as migrated too, since
 there is nothing further to do to them.
 
+### Producers that cannot read the catalog
+
+`@doenet/parser` and `@doenet/lsp-tools` raise author-facing diagnostics and
+neither can render one. Both are bundled into the DoenetML language server,
+which `@doenet/codemirror` embeds verbatim and starts as a blob worker, so
+every byte sits on the editor's critical path before the first cursor-help
+request can be answered — and the server has no locale to render in anyway.
+Reaching the catalog from there pulls in `EN_CATALOG_SOURCE`, every English
+namespace joined, plus the Fluent runtime to read it;
+`packages/lsp/scripts/check-server-bundle.mjs` fails outright if either
+arrives, rather than budgeting for it (see [Bundling](#bundling)).
+
+So these two pass the English **alongside** the code instead of rendering it —
+`codedDastError` in `packages/parser/src/coded-dast-error.ts`,
+`codedLspDiagnostic` in `packages/lsp-tools/src/coded-lsp-diagnostic.ts`. The
+message lives in two places, and each package holds the two together with a
+test: `test/coded-dast-errors.test.ts` and `test/coded-schema-violations.test.ts`
+run their producer over a corpus of broken DoenetML and assert every coded
+diagnostic renders, through the catalog, to exactly the string the producer
+wrote. A message edited on one side and not the other fails there, as does a
+missing or misnamed argument — one the catalog reads and the producer doesn't
+pass renders as `{$name}`, which no hand-written English will match.
+
+`@doenet/i18n` is a **devDependency** in both: `import type` in `src/`, so
+every code is still checked against the registry at compile time, and the real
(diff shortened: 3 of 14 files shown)


dqnykamp:
The "Not in scope" section here is now addressed by #1571, stacked on top of this and #1570: `@doenet/codemirror` takes an optional `diagnosticPresentation` (a message formatter and the severity headings) that `EditorViewer` fills from the translator its Diagnostics tab already uses, so the hover over a schema violation reads in the reader's language. No catalogs enter that package — the seam is two plain values — and the bundle guard is unchanged at 1044 KiB.

PR #1578, opened 2026-07-27
fix: keep section child render indices aligned with active children

dqnykamp:
## Problem

Children of a `<section>` silently vanish from the rendered document when the section also contains a `<stylePalette>`, `<styleDefinition>`, or `<feedbackDefinition>`. Reported case:

```
<section>
  <stylePalette palette="grumpynarwhal" />
  <styleDefinition styleNumber="2" lineOpacity="1" />

  <graph name="g1">
    <vector name="v">(3,4)</vector>
  </graph>

  <graph copy="$g1">
    <circle filled center="(5,2)" />
  </graph>
</section>
```

The second graph never appears. Removing any one of the `<section>`, the `<stylePalette>`, or the `<styleDefinition>` brings it back.

## Cause

`SectioningComponent`'s `childIndicesToRender` is consumed by `returnActiveChildrenIndicesToRender` as **indices into `activeChildren`**, but it was computed from a `child` dependency that listed only five of the section's eight child groups:

```js
childGroups: ["anything", "variantControls", "titles", "setups", "cascadeMessages"]
```

`styleDefinitions`, `stylePalettes`, and `feedbackDefinitions` were missing. Each child is matched to exactly one group, so a child of one of those types shortens the dependency list by one and shifts every later index down by one — that many children then fall off the end and are never handed to the renderer.

In the example above the section has nine active children, but `childIndicesToRender` came out as `[0…6]`, dropping `activeChildren[7]` (the copied graph) and `[8]`. `initializeRenderedComponentInstruction` pushed `null` in the graph's place, and neither it nor its descendants ever reached the renderer:

```
before:  ["\n  ", null, "\n  ", null, "\n\n  ", graph, "\n\n  ", null,  null]
after:   ["\n  ", null, "\n  ", null, "\n\n  ", graph, "\n\n  ", graph, "\n"]
```

That also explains the three "delete any one and it works" cases. Removing the `<section>` moves the content to `<document>`, which extends `BaseComponent` and has no `childIndicesToRender` at all. Removing either style tag leaves a shift of only one, which swallowed just a trailing whitespace string.

This is a latent bug in the index computation rather than anything specific to style palettes; adding the `stylePalettes` group simply made a one-child shift reachable in a second way, so two configuration children could stack up and eat a real child.

## Fix

A shared `returnSectionChildDependencies()` helper now supplies both `child` dependencies used to split a section's children into the ones it renders (or hides) and the ones that merely configure it:

- `allChildren` uses `includeAllChildren: true`, whose indices are by construction the keys of `activeChildren`. The list can no longer drift as child groups are added.
- `configurationChildren` picks out the style/feedback children, which `nonConfigurationChildEntries()` filters back out while keeping the `[position, child]` pairs of everything else. They stay excluded from the rendered set and from `firstVisibleChild` exactly as before, while the surrounding indices stay aligned.

`childIndicesToRender` and `childrenToHide` both iterate `nonConfigurationChildEntries()`. `childrenToHide` reports component indices rather than positions, so it was never affected by the shift — but it enumerated the complement of the same three groups by hand, which is the same stale-list hazard one step removed: a newly added child group would silently stop being hidden. Its behavior is unchanged; it just no longer maintains its own list.

`startsWithIntroduction`/`endsWithConclusion` looks `childIndicesToRender` up in its own child list, so it takes the same complete `allChildren` dependency. It does not need `configurationChildren`, since those children are already excluded from `childIndicesToRender`.

The doc comment on the helper records why `includeAllChildren` is required rather than a group list, so the next person to touch these dependencies does not reintroduce the shift.

## Tests

In `sectioning.test.ts`:

- Two regression tests assert the outcome — which children actually reach the renderer, via `rendererState[section].childrenInstructions` — rather than the internal index list:
    - the reported case (`<stylePalette>` + `<styleDefinition>` + a graph copied by a second graph)
    - a broader case adding `<feedbackDefinition>` with three `<p>` children
- The existing "starts with introduction / ends with conclusion" test gains a section that carries a `<stylePalette>` and a `<styleDefinition>` alongside its `<introduction>` and `<conclusion>`. This pins the two positionally-indexed dependencies to each other: reverting either one back to a partial group list makes the test fail.

In `cascade.test.ts`:

- A section inside a `<cascade>` — the only place `hideChildren` is ever set — carries a `<stylePalette>` and a `<styleDefinition>`. The test asserts both which children are rendered and that `childrenToHide` covers the section's content but not its title or its configuration children.

The three new tests all fail on `main` and pass with the fix.

Also re-ran `sectioning`, `cascade`, `stylePalette`, `styledefinition`, `styleOverride`, `styleDescriptionLocale`, `document`, `feedback`, `problem`, `solution`, and `hint` locally — all green. `npm run build:schema -w packages/static-assets` leaves the generated schema unchanged.

## Not fixed here

`<setup>` and `<variantControl>` were already in the old dependency list, so they never triggered the index shift — but they are also not rendered, and they can still be picked as a list item's `firstVisibleChild`, which incorrectly turns on the list-item first-child alignment adjustment. That is a separate pre-existing issue and is left alone here.

`<cascade>` overrides `childrenToHide` with a deliberately different child set of its own, and is likewise left alone.

### DIFF
--- a/.changeset/section-child-render-indices.md
+++ b/.changeset/section-child-render-indices.md
@@ -8,4 +8,4 @@
 
 Sections: fix children silently disappearing when a section contains a `<stylePalette>`, `<styleDefinition>`, or `<feedbackDefinition>`.
 
-Each of those configuration children shifted the section's rendered-child indices by one, so that many children fell off the end of the section and were never rendered. A section holding both a `<stylePalette>` and a `<styleDefinition>` lost its last two children.
+Each of those configuration children shifted the section's rendered-child indices by one, dropping an equal number of children off the end of the section. A section holding both a `<stylePalette>` and a `<styleDefinition>` lost its last two children.
--- a/.changeset/section-child-render-indices.md
+++ b/.changeset/section-child-render-indices.md
@@ -1,0 +1,11 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+"@doenet/vscode-extension": patch
+"doenet-vscode-extension": patch
+---
+
+Sections: fix children silently disappearing when a section contains a `<stylePalette>`, `<styleDefinition>`, or `<feedbackDefinition>`.
+
+Each of those configuration children shifted the section's rendered-child indices by one, so that many children fell off the end of the section and were never rendered. A section holding both a `<stylePalette>` and a `<styleDefinition>` lost its last two children.
--- a/.changeset/section-child-render-indices.md
+++ b/.changeset/section-child-render-indices.md
@@ -8,4 +8,4 @@
 
 Sections: fix children silently disappearing when a section contains a `<stylePalette>`, `<styleDefinition>`, or `<feedbackDefinition>`.
 
-Each of those configuration children shifted the section's rendered-child indices by one, dropping an equal number of children off the end of the section. A section holding both a `<stylePalette>` and a `<styleDefinition>` lost its last two children.
+Each of those configuration children shifted the section's rendered-child indices by one, so content at the end of the section was silently dropped — one child for each configuration child present.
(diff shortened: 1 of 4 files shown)


dqnykamp:
Follow-up for the issue noted under "Not fixed here": #1579 is stacked on this branch and takes `<setup>`/`<variantControl>` out of the list item's `firstVisibleChild` selection. Merge this one first.

PR #1592, opened 2026-07-29
Let the schema drive autocomplete in VS Code

### LINKED ISSUE #1317: [Feature]: Add `doenetml` to VSX Registry  (registry for VSCode-compatible editors)
## Problem

There's not an easy way to get the `doenetml` extension on VSCode-compatible editors (for example, I'm on Codium).

## Solution

Publish extension to the [Open VSX Registry](https://open-vsx.org/). I haven't looked into it much, but I can't imagine this would be hard to do - we would just mirror the vscode publish workflow. Many extensions are on both Open VSX Registry and VSCode's marketplace.

See https://github.com/EclipseFdn/open-vsx.org/wiki/Publishing-Extensions.

dqnykamp:
## The problem

The VS Code extension has always had a language server behind it, and most of
what it offers did work — element names, close tags, `$ref` and `$ref.member`
chains, enumerated attribute values, child suggestions. Two gaps made it feel
like the schema wasn't there at all:

**Attribute suggestions never appeared.** Typing `<math exp` offered nothing,
while `<ma` offered `math`, `mathInput`, … The two reach the server by
different routes: `<` is a registered **trigger character**, so the server is
asked about it unconditionally, but an attribute name has no trigger character
and depends entirely on the editor deciding to auto-suggest. Where that is
routed inline instead of to the suggestion widget, element suggestions keep
working while attribute suggestions look like they have vanished — and the slot
they would have filled goes to whatever renders ghost text.

**Ctrl+Space produced nothing** anywhere the previous character wasn't `<`, so
VS Code fell back to word-based suggestions: words scraped out of the current
file, which read as invented tags and attributes.

## The fixes

**Attributes.** Doenet documents now ask for the suggestion widget by default
(`editor.quickSuggestions.other`), settled in the one place that knows the
schema is worth showing. An explicit user or workspace setting still wins, so
anyone who prefers inline suggestions in Doenet files keeps them.

**Ctrl+Space.** The server opens its element menu for an explicit invocation
even with no leading `<`, but it learns the invocation was explicit from
`context.explicit` — a field only the CodeMirror client sets. VS Code can't set
it: LSP has no such field, and `triggerKind` can't stand in, because VS Code
reports `Invoked` both for Ctrl+Space and for the quick suggestions that fire on
a typed letter. A server keying off `triggerKind` alone would drop the whole
element menu into the middle of a sentence.

So a `doenet.triggerSuggest` command bound to Ctrl+Space (and Ctrl+I / Cmd+I,
matching the built-in's own bindings) records the keystroke and hands off to
`editor.action.triggerSuggest`; completion middleware marks the request it
produces. The request is issued by hand rather than through `next` because the
client's `asCompletionParams` rebuilds the context from `triggerKind` and
`triggerCharacter` alone, dropping anything else set on the way in. The command
is hidden from the command palette (`"when": "false"`), where it would read as
a second, Doenet-flavored copy of a built-in the palette already lists.

Refreshes of an `isIncomplete` list inherit the flag, so the menu keeps
narrowing as you type (`Ctrl+Space` → `poin` → `point`, `pointList`,
`polyline`, …) instead of emptying on the first keystroke — within the document
the session belongs to, so a second editor's refreshes stay its own. Any other
request starts a fresh session, and a pending flag is matched on document URI
and version, so a keystroke that never produced a request can't leak into a
later quick suggestion. Going around `next` also means going around the
rejection handling it carries, so the hand-issued request routes failures back
through `client.handleFailedRequest`: a completion the next keystroke cancels
is ordinary, not something to report.

Word-based suggestions are off by default in Doenet documents, where they only
ever compete with the schema.

## Also here

- **The server attaches to Doenet documents on any filesystem**, not just
  `file:` and `untitled:`. It was silent on vscode.dev and github.dev, which
  serve files as `vscode-vfs:`, and in virtual workspaces generally.
- **The preview no longer throws when no editor is active** — which happens
  every time the last editor closes or focus leaves the editor area.
  `onDidChangeActiveTextEditor` hands you `undefined` there. It had been
  filling the web extension host log with `TypeError: Cannot read properties of
  undefined (reading 'document')`; it's caught per-listener, so it was noise
  rather than breakage.
- **The extension mirrors to Open VSX**, putting it within reach of
  VS Code-compatible editors such as VSCodium. There are `publish:openvsx`
  scripts mirroring the `vsce` ones, both publish channels mirror when an
  `OVSX_PAT` secret is present, and the manifests declare a valid SPDX license
  (`AGPL` isn't one). Absent the secret each channel warns and exits cleanly
  rather than failing the run — in the production job that keeps a release
  which has already reached npm and the Marketplace from going red over a
  mirror that can be re-run on its own. The `doenet` namespace and the
  `OVSX_PAT` repository secret both exist, so the first pre-release publish
  goes out with the next successful CI run on `main`. The namespace is
  unverified, which shows the publisher as unverified in the listing but does
  not affect installability; claiming it is a separate request to the Eclipse
  Foundation.
- **`engines.vscode` moves `^1.76.0` → `^1.85.0`**, the release where
  `editor.wordBasedSuggestions` became an enum, so the contributed defaults are
  valid against the schema they're declared under.
- **READMEs.** The marketplace README's feature list now names the Ctrl+Space
  gesture, and the package README's publishing section covers the Open VSX
  channel alongside the Marketplace one.

`doenet.triggerSuggest` is registered before the language client is built, so a
server that fails to start leaves Ctrl+Space opening the built-in widget rather
than reporting an unknown command.

Closes #1317.

## Verification

Beyond the unit tests, I drove a real VS Code build headlessly
(`vscode-test-web` + Playwright) against the built extension and read the
suggestion widget's actual contents at each cursor position. Flipping only the
contributed default confirms it is what governs the attribute case:

| contributed `[doenet]` default | `<ma` | `<math exp` | `<math exp` + Ctrl+Space |
|---|---|---|---|
| `other: "off"` | `math`, `mathInput`, … | *nothing* | `expand` |
| `other: "on"` | `math`, `mathInput`, … | `expand` | `expand` |

Ctrl+Space in the body went from `hello`, `text` (the file's own words) to
`title`, `setup`, `p`, `math`, `graph`, `answer`, … And `<gra`, `<point s`,
`displayMode="`, `</`, `$`, `$m.`, and `<` inside `<answer>` were re-checked to
confirm nothing that already worked regressed.

### DIFF
--- a/.changeset/vscode-explicit-completions.md
+++ b/.changeset/vscode-explicit-completions.md
@@ -21,6 +21,10 @@ them inline instead left element suggestions working while attribute
 suggestions appeared to be missing entirely. Doenet documents now ask for the
 widget by default.
 
+Both defaults are contributed under `[doenet]`, which raises the extension's
+minimum VS Code version to 1.85 — the release where
+`editor.wordBasedSuggestions` became an enum.
+
 The language server also attaches to Doenet documents on any filesystem rather
 than only `file:` and `untitled:` ones, so completions, diagnostics, hovers and
 formatting work on vscode.dev, github.dev, and in virtual workspaces.
--- a/.changeset/vscode-explicit-completions.md
+++ b/.changeset/vscode-explicit-completions.md
@@ -13,6 +13,14 @@ the web editor, and keeps narrowing it as you type. Word-based suggestions are
 off by default in Doenet documents, where they only ever compete with the
 schema; `"[doenet]": { "editor.wordBasedSuggestions": ... }` turns them back on.
 
+Attribute suggestions no longer depend on how the editor is configured to
+auto-suggest. Typing `<math exp` reaches the language server only if quick
+suggestions open the suggestion widget — unlike `<`, which is a trigger
+character the server is always asked about — so an editor configured to render
+them inline instead left element suggestions working while attribute
+suggestions appeared to be missing entirely. Doenet documents now ask for the
+widget by default.
+
 The language server also attaches to Doenet documents on any filesystem rather
 than only `file:` and `untitled:` ones, so completions, diagnostics, hovers and
 formatting work on vscode.dev, github.dev, and in virtual workspaces.
--- a/.changeset/vscode-explicit-completions.md
+++ b/.changeset/vscode-explicit-completions.md
@@ -13,13 +13,13 @@ the web editor, and keeps narrowing it as you type. Word-based suggestions are
 off by default in Doenet documents, where they only ever compete with the
 schema; `"[doenet]": { "editor.wordBasedSuggestions": ... }` turns them back on.
 
-Attribute suggestions no longer depend on how the editor is configured to
-auto-suggest. Typing `<math exp` reaches the language server only if quick
-suggestions open the suggestion widget — unlike `<`, which is a trigger
-character the server is always asked about — so an editor configured to render
-them inline instead left element suggestions working while attribute
-suggestions appeared to be missing entirely. Doenet documents now ask for the
-widget by default.
+Attribute suggestions no longer go missing. Typing `<math exp` reaches the
+language server only if quick suggestions open the suggestion widget — unlike
+`<`, which is a trigger character the server is always asked about — so an
+editor configured to render them inline instead left element suggestions
+working while attribute suggestions appeared to be missing entirely. Doenet
+documents now ask for the widget by default; an explicit
+`editor.quickSuggestions` setting still wins.
 
 Both defaults are contributed under `[doenet]`, which raises the extension's
 minimum VS Code version to 1.85 — the release where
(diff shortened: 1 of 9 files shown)


dqnykamp:
Added `"editor.quickSuggestions": { "other": "on" }` to the `[doenet]` contributed defaults (d8761bf).

Attribute suggestions reach the server only if quick suggestions open the suggestion widget. `<` is a registered trigger character, so the server is asked about it unconditionally — but `<math exp` has no trigger character, so it depends entirely on the editor deciding to auto-suggest. Where that is routed inline instead, the result is that element suggestions keep working while attribute suggestions look like they have vanished, and the slot they would have filled goes to whatever renders ghost text.

Verified against a real VS Code build, flipping only the contributed default:

| contributed `[doenet]` default | `<ma` | `<math exp` | `<math exp` + Ctrl+Space |
|---|---|---|---|
| `other: "off"` | `math`, `mathInput`, … | *nothing* | `expand` |
| `other: "on"` | `math`, `mathInput`, … | `expand` | `expand` |

So the contribution is read and it is what governs this. An explicit user or workspace setting still takes precedence, so anyone who prefers inline suggestions in Doenet files keeps them.

PR #1603, opened 2026-07-30
Load message catalogs for locales that are not inlined

### LINKED ISSUE #1611: Loaded (non-inlined) locales: uiLocale switch discards the reader's work, and a nested <document lang> never loads
Two defects in the on-demand catalog loading added by #1603/#1604. Both affect every language that is not inlined — today all eight of French, German, Italian, Dutch, Russian, Somali, Chinese and Hmong Njua. Neither is caught by CI, because the tests that would catch them use Spanish, which is still inlined and so never takes the loading path.

They surfaced while trying to stop inlining Spanish (making the rule uniformly "English is bundled, everything else loads"). With Spanish loaded rather than inlined, two long-standing tests in `packages/test-cypress/cypress/e2e/DocViewer/i18n.cy.js` fail. That work is parked until these are resolved, since landing it would extend both defects to Spanish.

## 1. Changing `uiLocale` rebuilds the core and discards the reader's work

`packages/doenetml/src/Viewer/DocViewer.tsx:2120-2123` states the intended contract:

```ts
// The content locale and its catalogs are fixed for a core's lifetime —
// they feed every string the core computes — so changing either rebuilds
// the core rather than updating in place, the way a changed `doenetML`
// does. (`uiLocale` is main-thread only and needs no rebuild.)
```

The statement immediately below breaks it:

```ts
const resourceKey = localeResourceKey(availableCatalogs);
if (lastLocaleResourceKey.current !== resourceKey) {
    lastLocaleResourceKey.current = resourceKey;
    changedState = true;
}
```

`availableCatalogs` is the **merged** map from `useLocaleCatalogs([effectiveUiLocale, effectiveDocumentLocale], …)`. When a reader switches `uiLocale` to a language that is not yet loaded, its *chrome* catalog arrives, the merged map gains a locale, `resourceKey` changes, `changedState` is set, and `coreId.current = nanoid()` rebuilds the core — even though nothing the core computes has changed.

Everything the student had typed goes with it:

```
retranslates in place when uiLocale changes:
  Timed out retrying after 30000ms
  + expected - actual
  +'wrong'            ← #ti_input is empty; the typed answer was discarded
```

The test's own comment is the specification: *"`uiLocale` is main-thread only: no core rebuild, so whatever the reader had typed survives the switch."*

**Suggested fix.** Key the rebuild on the catalogs the core actually consumes — the ones reachable from the content locale's negotiation chain — rather than on the merged map. A chrome-only catalog arriving would then not touch `resourceKey`. The existing behaviour for the *content* locale is correct and should be preserved: a content catalog landing after the core was created must still rebuild it.

## 2. A nested `<document lang>` never gets its catalog, and renders in English forever

`DocViewer.tsx:579` asks for exactly two tags:

```ts
const availableCatalogs = useLocaleCatalogs(
    [effectiveUiLocale, effectiveDocumentLocale],
    localeResources,
);
```

A language declared by a **nested** `<document lang="…">` is in neither. Its catalog is therefore never requested, and the nested subtree renders in English — not late, never:

```
labels a nested document with its own language:
  Timed out retrying after 30000ms
  -'thick dashed red line'
  +'línea discontinua gruesa roja'
```

This is not merely a main-thread rendering gap. The string under test is a style description **computed by the core**, and the nested language reaches the renderer as the `renderedLang` state variable added in 9fdb73f9 (#1594) — i.e. the main thread only learns the language *after* the core has already computed that subtree's prose in whatever catalogs the worker was given. So the worker needs the catalog in hand, and the loading has to be driven from something known earlier.

**Fix options.** Neither is obviously right, hence this issue rather than a PR:

| # | Approach | Notes |
|---|---|---|
| 1 | Have the core report the set of languages its documents declare; feed them into `useLocaleCatalogs`, let the catalog land and let the existing `resourceKey` path rebuild the core with it in hand | Uses machinery that already exists and is already documented as intended ("a catalog that finished loading after the core was created rebuilds it with that language on hand"). Costs a rebuild per newly discovered language. |
| 2 | Scan the source for declared `lang` attributes on the main thread before building the core, and preload those catalogs | No rebuild, and `@doenet/parser` is already imported on this path (`detectVersionFromDoenetML`). Needs to be right about where `lang` may legitimately appear. |
| 3 | Preload every language named anywhere in the source, cheaply, without parsing | Simplest, least precise. |

Option 1 composes with defect 1's fix: once `resourceKey` is keyed on content catalogs, a nested language arriving is exactly the case that *should* rebuild.

## Reproducing

Both are reachable today without any change, using any of the eight loaded languages — e.g. switching `uiLocale` to `"de"` at runtime with text typed into an input, or nesting `<document lang="de">` inside an English activity. To see them via the existing Spanish tests, drop `es` from `BUNDLED_TRANSLATIONS` in `packages/i18n/src/bundled.ts` and its `!` line from the glob in `packages/i18n/src/load.ts`, then run `npm run test:e2e-group5`. The rest of that group is unaffected: 247 of 249 tests pass, and the only two failures are these.

Worth noting from the same run: `follows the content's language, not the reader's` still passes but takes 17.2 s against 1–3 s for its neighbours — the load-then-rebuild cost, which is the behaviour all eight languages have today.

dqnykamp:
Only English and Spanish are carried inside the JavaScript. Every other locale now arrives on demand, so adding a language costs a directory under `packages/i18n/locales/` (plus the codegen `lint:i18n` already insists on) — no bundling decision, and no registry of languages to keep in step with it.

This is the groundwork for seeding translations in more languages — at roughly 16 KB gzipped for a complete translation, inlining them all would put the cost of every language on everyone who uses one, or none.

## How a catalog reaches the browser

`load.ts` finds the catalogs with `import.meta.glob`, which is what keeps the list of languages out of the source: a directory *is* the registration, the same way `SUPPORTED_LOCALES` derives its names from `Intl` rather than spelling them out.

The viewer does the loading itself, through a new `useLocaleCatalogs`, merging what it loads *under* the host's `localeResources` so a deployment correcting a shipped translation still wins. `localeResources` is therefore no longer the route to a language — it is the override — and the prop's documentation on `DoenetViewer`, `DoenetEditor` and `DocViewer` now says so. `documentLocale="de"` and `<document lang="de">` therefore work with nothing configured and nothing registered.

It is called wherever a language becomes known: `DoenetViewer` and `DoenetEditor` over the props, `DocViewer` over the languages it resolves from the source, and `EditorViewer` over the tag the viewer reports back — which the props may never have named. Requests are shared, so a language named in three places is fetched once.

`loadLocaleResources` returns `{}` rather than throwing for English, for a bundled locale, for a locale nothing offers a catalog for, and for a fetch that fails — a missing translation is not an error, and this runs during a paint. Tags negotiate as usual (`de-AT` loads `de`, keyed `de`), and requests are cached by what was asked for so N viewers on a page share one fetch. The namespaces are a parameter, since each costs a request or a chunk of its own; the viewer asks for all four because the one map it builds feeds both its own chrome and the core it hands `LocaleData.resources` to.

## The single-file builds need a different answer

`inlineDynamicImports` folds every dynamic import back into the one output file, so code-splitting cannot keep catalogs out of `@doenet/standalone` — and **being reachable is enough, whether or not anything calls it**. It defines `__DOENET_CODE_SPLIT_CATALOGS__` false to make the glob dead code, copies `locales/` into `dist/`, and installs `fetchLocaleLoaders` against that copy. Same arrangement as the core worker externalization, for the same reason.

The core worker gets the same define and needs no replacement of its own: the main thread hands it the one catalog it needs as `LocaleData.resources`. Without the define the glob would put every non-inlined catalog into that bundle — reachable, never read, and impossible to fetch back out of an IIFE started from a blob URL — which is how the need for it was caught.

Serving that copy is the one thing a host now has to do, so `@doenet/standalone`'s README says so — along with the origin fallback for an embed booted from a blob URL, and what happens when neither is served (the fetches fail quietly and the language falls back to English). `@doenet/doenetml-iframe`'s README no longer describes `localeResources` as the only route to a language beyond the bundled two.

| Build | Mechanism |
| --- | --- |
| `@doenet/doenetml` | `import.meta.glob` — one chunk per catalog |
| `@doenet/standalone` | `fetch` from `locales/`, served beside the bundle |
| `@doenet/doenetml-worker` | Neither — handed `LocaleData.resources` |
| `@doenet/doenetml-iframe` | Neither — what renders inside its iframe is a standalone bundle, which loads its own |

## Three defects loading exposed, fixed here

Closes #1611. All were invisible while every language anyone used was inlined, and all applied to every locale that is not.

**Changing `uiLocale` rebuilt the core and discarded the reader's work.** The chrome's catalogs and the content's share one map, and the core was rebuilt whenever anything in that map changed — so switching the reader's language pulled down a catalog the core never reads and took every typed answer with it. The comment three lines above that check already promised the opposite. A new `reachableCatalogLocales` answers which of the catalogs on hand a given set of locales can negotiate to, and the rebuild is gated on the content's. It filters to catalogs actually on hand, because `negotiateLocales` appends the default locale whether or not one arrived, and an answer about what has arrived must not claim English it never received.

The content's languages are the host's `documentLocale`, every language the source declares, and whatever core initialization resolved — the last of which is how the *outermost* `<document lang>` arriving through an external reference is still picked up (a nested one inside external content is not, and stays English). Only a catalog *arriving* counts, because what those locales can reach also shrinks: `effectiveDocumentLocale` names the previous language until the core has reported back, so an activity whose `<document lang>` overrides the host's tag — or a host that changes `documentLocale` mid-session — would otherwise move the gate on the strength of the very build the gate guards, and build the core a second time. The corollary is that withdrawing a catalog rebuilds nothing: a host that takes a locale back out of `localeResources` leaves the core rendering what it was built with. That is written on the prop.

**A nested `<document lang>` never had its catalog requested**, so its subtree rendered English however long you waited — it is in neither tag handed to `useLocaleCatalogs`, and the core settles it only while already computing that subtree. The languages are now read off the source, where the author wrote them, by a new `readDeclaredLangs`. `lang` is a primitive string attribute — a reference in it is rejected rather than resolved — so the literal text is every language the core could render in. It shares its attribute reading with `readDocumentLang`, which tightens that helper to what the core accepts: exactly one string child, where it used to join every text child and so could label a document with a language the core refuses to build.

The scan is skipped outright for source with no `lang` written in it, which is nearly all of it — this runs whenever the source changes, and in the editor that is as its author types.

**A viewer given `render={false}` started a core worker on every re-render and abandoned the one before it.** That branch returns before it records the source it was given, so it is reached again on every render until `render` turns true, and `attachNewCoreWorker` overwrites the refs — leaving the predecessor running with nothing pointing at it, never terminated, still reporting the document structure its replacement is also reporting. The leak predates this PR; what is new is how easy it is to reach, since a catalog landing re-renders the viewer and a viewer that is not rendering yet is exactly the one still in that window. It now reuses the worker it primed, and re-initializes it only when one of the initialization's inputs has changed — the same inputs the rebuild is gated on, compared for any change rather than for an arrival, there being no core yet whose work a needless rebuild would throw away. The reachable content catalogs move into a memo that gate and this both read.

Worth noting for review: gating the rebuild correctly also made the common case faster. `follows the content's language, not the reader's` in `e2e-group5` went from 17.2 s to 2.7 s, because the source scan prefetches the top-level document's catalog too rather than waiting for a core round trip and a rebuild.

## Guards

Every invariant here is the silent kind, so each is checked:

- `packages/standalone/scripts/check-bundle-size.mjs` fails if a served catalog turns up inside an emitted script, and if any locale directory did not reach `dist/locales/` — the build copies `locales/` whole, inlined locales included, so that half of the check has something to say even now, when nothing is served yet. It fingerprints each served locale by a string long enough, and different enough from *every inlined* catalog, to identify it: matching one of those would report a leak over strings that legitimately are in the bundle, and English is not the only inlined locale. The fingerprint is compared against the script with its character escapes decoded, since the minifier writes accents back out as `\xNN` and a raw comparison would pass over exactly the languages the check exists for.
- `lint:i18n` fails if the locales excluded from the glob stop matching `BUNDLED_LOCALES`. One way round costs a chunk and a Rollup warning on every build; the other makes a language silently unloadable.

## Verification

Checked against a throwaway locale carrying marker strings, across every bundling variant:

- `@doenet/doenetml` — markers in four per-namespace chunks, absent from `index.js`, `doenetml-inline-worker.js` and `doenetml-external-worker.js`
- `@doenet/doenetml-worker` — absent from both `index.js` and `index.esm.js` (it was present before the define)
- `@doenet/standalone` — absent from every emitted script, present in `dist/locales/`, within budget
- `@doenet/doenetml-iframe` — the component build code-splits, though nothing in it loads a catalog: what renders inside the iframe is a standalone bundle, and the IIFE shim delegates to it

Both new guards were negative-tested by breaking them — including against the real bundle, where an accented Spanish string that is genuinely inlined is invisible to a raw text search and is found once the escapes are decoded. `lint:i18n`, the LSP bundle guard (which this had to stay clear of), and the i18n suite all pass.

31 new unit tests cover the loader (2 of which stay skipped until a locale that is not inlined exists; `loadersFromModules` is driven directly against a synthetic module map for the same reason), 16 the bundle checks, and 16 more the first two fixes above — `reachableCatalogLocales` ignoring a catalog the given locales cannot reach, and `readDeclaredLangs` finding a nested language, keeping document order, and taking nothing from an attribute the core would reject.

Seven Cypress component tests. Four drive the whole round trip without waiting for a translation to exist: a stub loader is installed for a language nothing ships, through the same `setLocaleLoaders` seam `@doenet/standalone` uses. Two of them are the round trip itself — `<paginatorControls>` comes back in that language, which can only happen if `useLocaleCatalogs` loaded the catalog on the main thread and it reached the core as `LocaleData.resources`, while its untranslated half stays English; and the same document renders entirely in English with no loader installed. The other two pin the first two defects: the controls come back translated from inside a nested `<document lang>` that no prop names, and an answer typed before a `uiLocale` switch is still there after the reader's catalog has loaded.

Three more pin the third, counting `new Worker`. One re-renders a `render={false}` viewer three times; the other two follow it out of that window — `render` turning true, and a source edited while it was still false — where the worker that was primed has to be the one that goes on to render, carrying the edited source. All five of the defect tests fail against the code as it was before the fix each covers.


dqnykamp:
Follow-up #1604 stacks on this: the eight seeded language catalogs. It also re-exercised the guards here against real catalogs rather than a probe locale — 32 code-split chunks emitted, nothing in the entry chunk or the worker, and the standalone bundle check confirmed none of the eight leaked into the single-file bundle.

dqnykamp:
The `doenetml-iframe-cypress` failure on this PR is not caused by these changes. Tracked in #1608.

Short version: this PR branch's own history contains the proof. Run [30558745113](https://github.com/Doenet/DoenetML/actions/runs/30558745113) passed that job 20/20 in 5:01; run [30560716275](https://github.com/Doenet/DoenetML/actions/runs/30560716275), 25 minutes later, failed 11/20 — and the entire diff between the two commits is three comment lines inside `.ftl` catalogs, which Fluent strips at parse time. `@doenet/doenetml` is not even in that suite's runtime module graph (`src/utils.ts` imports it only in `typeof` positions, so esbuild elides it), so the i18n work here cannot reach it.

The underlying fragility reproduces off CI by pinning Cypress to 4 cores, and #1608 records the measurements — including two fixes ruled out (raising the timeout doesn't help; the base64 harness step is a one-time 67 ms) and the three harness costs worth attacking.

dqnykamp:
~~Two defects in this PR's loading machinery are now tracked in #1611.~~ **Both are now fixed in this PR** (`829e6fd40`), which closes #1611. Leaving the original findings below, since they are what the fix is for.

1. Switching `uiLocale` to a not-yet-loaded language rebuilt the core and discarded whatever the reader had typed. `DocViewer.tsx:2120-2123` promised the opposite ("`uiLocale` is main-thread only and needs no rebuild"); the `resourceKey` check below it was keyed on the merged catalog map, so a chrome-only catalog arriving tripped `changedState`. Now gated on the content's catalogs via `resourceKeyForLocales`.
2. A nested `<document lang="…">` was in neither tag passed to `useLocaleCatalogs` (`DocViewer.tsx:579`), so its catalog was never requested and the subtree rendered English permanently. The declared languages are now read off the source by `readDeclaredLangs`, before the core is built.

Both surfaced while making Spanish load like every other translation. That change has since landed in #1604, once these were fixed — with Spanish loaded, both defects reached Spanish too, which is what turned two long-standing e2e tests red and made them visible at all.

`e2e-group5` is now **249 of 249**, up from 247. `follows the content's language, not the reader's` also dropped from 17.2 s to 2.7 s: reading languages off the source prefetches the top-level document's catalog too, so every non-English document skips a core round trip and rebuild rather than only nested ones.

dqnykamp:
Two corrections to earlier comments on this PR, which have gone stale.

**The `doenetml-iframe-cypress` failure is resolved.** That job is green here now. The root cause was found and fixed on `main` by #1609: the iframe boot script loaded Comlink from `https://unpkg.com`, and a static `import` is resolved before any of a module's body runs, so a stalled CDN fetch left the whole boot script unexecuted — no handshake, the iframe parked on its loading placeholder, no timeout and no error. This branch is rebased onto that fix. It also accounts for what #1608 recorded and could not explain at the time: the blank iframe, the core worker never being requested, and a 90 s budget not rescuing it.

**"32 code-split chunks" is now 36.** That count was taken before Spanish stopped being inlined in #1604. Nine translations × four namespaces; English is the one language still carried in the entry chunk. The claim it was making — nothing in the entry chunk or either worker, and nothing leaking into the single-file bundle — still holds, and the standalone bundle guard now verifies the Spanish catalogs are served rather than carried too.

Current state of the stack: both PRs are green, and this one closes #1611.

Issue #1521, opened 2026-07-20
i18n: community translation platform setup (Weblate)


dqnykamp:
Part of #861 (internationalization umbrella). This sets up the **community translation workflow** — how volunteer translators contribute and how translations flow into releases. Can start as soon as Phase 1 ships the first real catalog; runs in parallel with later phases.

## Platform: Weblate (recommended)

- Hosted libre tier at hosted.weblate.org is gratis for open-source projects; Weblate itself is open source and self-hostable if hosting terms ever change.
- Native support for both catalog formats we use: **Fluent** (`.ftl`, including terms/attributes) for software strings and **XLIFF** for document translation (Phase 5).
- Machine-translation suggestions, review states, glossaries, and direct git integration.
- Runner-up: Crowdin (solid OSS tier but proprietary; occasional OSS-tier policy changes). Transifex is no longer meaningfully OSS-friendly.

## Repo layout & ownership

- `packages/i18n/locales/en/*.ftl` — **developer-owned source of truth.** New keys land in the same PR as the feature that uses them (CI-enforced).
- `packages/i18n/locales/<other>/*.ftl` — **Weblate-owned.** Weblate commits to a `weblate` branch and opens PRs, so translations get normal review and CI like any contribution.

## CI

- `lint:i18n`: every `t()` key (and every diagnostic code) exists in `en`; no orphaned keys.
- Fluent syntax check + placeholder/variable parity check across all locales (Weblate also enforces this; belt and suspenders).
- Pseudo-locale (`en-XA`) build with a Cypress smoke spec — catches hardcoded strings regressing in.
- Webhook: merges to `main` touching `locales/en/` trigger a Weblate refresh so new strings appear for translators automatically.

## Fallback chain

`negotiateLocales` (from `@doenet/i18n`, wrapping `@fluent/langneg`) builds an ordered bundle list, e.g. `es-MX → es → en`. The `en` bundle is always present (inlined into the package), so a missing translation renders English — never a raw key.

## New-string lifecycle

1. Feature PR adds the `en` key (CI-enforced).
2. Merge triggers Weblate sync; the string appears for translators.
3. Runtime falls back to English until the translation lands via a Weblate PR.
4. Diagnostic codes are append-only (CI check) so translated messages never silently detach from their codes.

## Size / dependencies

**S–M** (mostly configuration + CI wiring). Depends on Phase 0 (catalogs exist) and is most useful once Phase 1 ships the first `es` catalog.

[...]

PR #1604, opened 2026-07-30
Seed message catalogs for eight more languages

dqnykamp:
> #1603 has landed and this branch is rebased onto it, so the diff is now this PR's own. #1603 was the machinery that lets these eight languages reach the browser without adding ~130 KB gzipped to everyone's bundle.

French, German, Italian, Dutch, Russian, Somali, Chinese and Hmong Njua, across all four namespaces: the viewer chrome, the editor and language-server surfaces, the prose the core computes into a document, and the 214 warnings and errors. 559 keys × 8, less the 130 chemistry names Somali and Hmong Njua leave to English — see [the deliberate gaps](#two-deliberate-gaps).

`documentLocale="de"` and `<document lang="de">` now work with nothing configured, and all eight reach `<document lang>`'s autocomplete through the roster, each with its name in its own script (`русский`, `中文`, `Soomaali`) — derived from ICU at codegen time, not hand-written.

Two consequences follow from having nine translations rather than one, and both are in here: **Spanish stops being inlined**, and the bundled-translations machinery that held it is retired.

## These are unreviewed seeds, and they say so

Every file opens with a header saying it is machine-generated and pending review by a speaker (#1521). That is the honest state of them, and it is the reason this is safe to merge: **nothing fails silently.** A key a translation lacks renders in English, so a wrong string is wrong in one place and one key away from being fixed. `lint:i18n` reports any locale that falls short of the 559 on every run.

Spanish is one of them. Its catalogs arrived the same way in #1543 and no speaker has read them either, so all four now carry the identical header and the package README says plainly that English is the source of truth and that **none** of the nine translations has been reviewed. No locale is privileged, and nothing in the code, the tooling or the docs records a review status.

Those catalogs were also the only ones whose *own* commentary — header, `##` group headings, and the notes explaining a wording choice — was written in the language they translate into. They are in English now, like the other eight: a note nobody maintaining the file can read is a note nobody can check. The README states the convention.

Expect wording to be corrected. The point is to give #1521's translation platform something to correct.

## The style descriptions needed real grammar, not translation

`content.ftl` is the namespace that cannot be done word-for-word, and each language declares how it works:

| | agreement | word order |
| --- | --- | --- |
| German, Russian | three genders | adjectives before the noun |
| Dutch | *de*- vs *het*-words | adjectives before the noun |
| French, Italian | two genders | adjectives after the noun |
| Chinese, Somali, Hmong Njua | none | — |

Rendering the output rather than reading the catalogs is what found the bugs. All of them are fixed here except one, which no catalog can reach — see below.

Two were case errors on the border clause:

- German `mit einem dicker Rand` → **`mit einem dicken Rand`**
- Russian `с ромбами и толстая границей` → **`с ромбами и толстой границей`**

`mit` governs the dative and «с» the instrumental, but the adjective was arriving in the nominative. The fix uses the one lever the design offers — `noun-gender` returns a token whose meaning the *catalog* chooses, so `border` answers `datm` / `insf` and the adjectives carry the case. What that token cannot carry is two cases, so `borderStyleDescription` (a border's style with no preposition in front of it) comes out in that case too. `background` faces the identical fork and is resolved the other way, so `backgroundColor` reads right alone and `style-text` carries the nominative into its clause. Each catalog says so at the point it makes the choice, and cites #1606; fixing it properly means the code passing a case alongside the gender.

Five more were in `fillStyleDescription`, the one state variable that renders a fill pattern outside the "with dots" clause. It hands `style-fill` the gender `noun-gender` answers for `fill`, and five catalogs spent it on a noun that was not the one they had named:

- German `Punkten in rosafarbene` → **`rosafarbene Füllung mit Punkten`**
- Russian `точками цвета розовая` → **`розовая заливка с точками`**
- French `points de couleur noir` → **`remplissage noir avec des points`**
- Dutch `stippen in rode` → **`rode vulling met stippen`**
- Somali `dhibco midab casaan khafiif ah ah` → **`dhibco casaan khafiif ah`**

The first four now name the noun the colour agrees with — *Füllung*, *заливка*, *remplissage*, *vulling* — each of which is the gender `noun-gender` already answers for `fill`. German and Russian had been seeded from the Romance `{ pattern } de color { color }` framing while spelling their pattern words in the case the "with dots" clause governs; French had hung the colour off a feminine «de couleur» and Dutch off a bare `in`. Somali dropped the `midab … ah` framing, which doubled the particle for the colours that already carry one.

Also: German and Dutch fold the side count into a compound (`regelmäßiges 5-Eck`, `regelmatige 5-hoek`) and so have no noun tail, where French and Italian split it (`polygone régulier à 5 côtés`) the way Spanish does. Chinese joins modifiers to the noun with no space *and* reorders them — a `带…的` clause has to precede what it modifies — so its composition messages differ from the English ones in both respects.

One rendering is wrong and stays wrong: `describeClosedShape` appends the border clause with a literal space, the single join it does not hand to the catalog, so Chinese renders `带圆点的蓝色填充圆 和红色边框` with a space Chinese would not write. `zh/content.ftl` says so at `style-border-clause`. Closing it means changing `styleDescriptions.ts` to pass that join through the catalog like every other one, which is not this PR's diff — tracked as #1605, alongside #1606 for the related case-vs-gender limitation the German and Russian catalogs work around.

## The headers have to be true too

Each catalog opens with a note on the register it keeps, so that whoever corrects it knows what to match. Five of them described a register their file does not have: Italian and Somali label every control with the bare imperative, which is what those languages put on a button and not the impersonal form the header claimed; Russian and Dutch reach for an imperative in the click-to-open pair as well as in the two reload messages; German's reload messages are infinitives softened with `bitte`, not the imperative they were called. Each header now describes what its file does.

Three `content.ftl` headers described more than their file does. German and Russian said every description is a standalone phrase in the nominative and that `$gender` carries one token per gender — but both files have carried a fourth since they were seeded, `datm` and `insf`, which `noun-gender` explains a hundred lines below and the header contradicted. Chinese said its composition messages differ from English's only in spacing, when they differ in order too. Each header now names the exception and points at the explanation.

Two more were checked against something outside the file and did not hold. Chinese said no spaces are inserted around Latin text or numbers, where every placeable in the file is spaced off from the Chinese around it (`还剩 { $count } 次尝试`); what is actually tight is the unit against its number, `{ $percent }%`, which the file keeps throughout. Hmong Njua justified its tag with more than `Intl` delivers — see [the deliberate gaps](#two-deliberate-gaps) below.

A handful of words were wrong rather than merely unidiomatic: Italian named a circle `cerchio` in the descriptions and `circonferenza` in all ten diagnostics that mention one (now `circonferenza`, and feminine, which `noun-gender` had no entry for); Somali's parabola was `qalooc bir ah`, a curve made of metal, where its own diagnostics already say `parabola`; Hmong Njua called an interval `ceg`, which the same catalog uses for a polygon's side, where its diagnostics say `ntu`; German's Clear button said `Zurücksetzen`, which is Reset; Russian's feedback heading said `Отзыв`, which is a review someone leaves; and Chinese punctuated a numbered heading with the full stop `。` and hard-coded 页 into the second half of its page status, where the first half takes the author's own `pageLabel`.

## Two deliberate gaps

**Somali and Hmong Njua omit the 118 element names and 12 anion names.** Neither language has a settled chemical nomenclature I could seed from, and inventing one is worse than the English fallback. They sit at 429/559 and the lint says so, rather than reporting full coverage over invented terms.

**Arabic was prepared and then dropped.** Its grammar and plural rules work (`zero`/`one`/`two`/`few`/`many` all resolve correctly), but DoenetML has no right-to-left support: nothing emits `dir` beside the `lang` on the rendered wrapper, no stylesheet is written for RTL, and `createTranslator` defaults `useIsolating` off. An Arabic catalog would render correct words in a broken layout. Worth its own issue — it unlocks Hebrew, Persian and Urdu at the same time.

`hnj` rather than the macrolanguage `hmn`, deliberately — and for exactly two reasons, both checked against `Intl` rather than assumed. CLDR carries plural rules for `hnj` and none for `hmn`, so `Intl.PluralRules("hnj")` answers with Hmong's single category where `hmn` falls back to English and offers a `one` the language does not have; and `Intl.DisplayNames` names it "Hmong Njua" rather than plain "Hmong", which is what the roster shows an author. Number formatting and list joining fall back to `en-US` under *either* tag — `Intl` has no data for the pair — so they are a gap to fix in CLDR, not a reason to prefer the macrolanguage. `hnj/chrome.ftl` says so at the point it makes the choice.

## Spanish stops being inlined, and English is the only bundled language

Spanish was inlined back when it was the only translation there was. With nine of them, no single language is worth ~16 KB gzipped to every reader who never opens it, and none of the nine has a claim on the bundle the others lack. English stays, because it terminates every fallback chain and has to be reachable with no network and no module graph.

`@doenet/standalone` drops about 116 KB — it carried the Spanish catalogs in `doenet-standalone.js` and again in its core worker, roughly 58 KB apiece. Spanish takes the route the other eight already take: four code-split chunks from `@doenet/doenetml`, which now emits 36 of them rather than 32, and a directory under `locales/` for `@doenet/standalone`. Nothing needs configuring — `documentLocale="es"` and `<document lang="es">` work exactly as before. What changes is *when* the catalog arrives: a Spanish document paints in English until it loads and then re-renders, because a document's language is fixed for the lifetime of the core that renders it and is not known until the source has been parsed. That was already true of the other eight. A host that needs Spanish on the first frame passes `localeResources`, which still win over anything shipped.

The lazy-catalog glob now excludes `en` alone, and `lint:i18n` check 4c holds it to `BUNDLED_LOCALES`. The standalone bundle guard derived the same thing from the glob on its own, so it now actively verifies that the Spanish catalogs — and the eight new ones — are served rather than carried.

`DoenetViewer.checkWorkLocale` read its label with `.invoke("text").then(...)`, which does not retry, so it captured the first paint's English; it now settles on the document's language before reading. Every other `.invoke("text")` in these specs already gated on a retrying assertion.

Twenty-one unit tests, across five files, used bundled Spanish as a fixture. They now supply the catalog the way a host does — which is what they were covering anyway, and leaves them independent of what the bundle happens to carry. The worker's `test-core` helper reads this repository's catalogs for the locale it was asked for *and* for every language the source declares, which is the set the main thread loads and sends through `setLocaleData`, so the worker's locale suites keep working without each carrying the plumbing: 37 of those 111 tests fail if the helper stops reading catalogs off disk, 9 of them reached only through the declared-lang scan — a `<document lang>` naming a language no prop did. It negotiates by primary subtag, as `loadLocaleResources` does, so `es-MX` is still served by the `es` catalog.

That helper and `DocViewer` need the same answer to "which languages does this source declare?", so `readLangAttribute`, `readDeclaredLangs` and `declaredLangsInSource` now live in `@doenet/parser` and both call the last of them. It carries the cheap `lang=` pre-test that keeps a document declaring no language from paying for a parse on every keystroke, and the `try`/`catch` that keeps half-typed source from throwing out of a prefetch — which the test helper needs at least as much, since a great many worker tests feed the core deliberately malformed DoenetML.

## The bundled-translations machinery is retired

With Spanish loading like every other language, `BUNDLED_TRANSLATIONS` was an always-empty map, its `pick` only ever picked English, and the worker's `BUNDLED_WORKER_RESOURCES` was an empty object spread into every translator it built. What was left read as though something might be in it when nothing ever is, so `packages/i18n/src/bundled.ts` is gone.

`BUNDLED_LOCALES` becomes the constant `["en"]` in `catalogs.ts`, beside the English catalogs it now describes, and `bundledResources` becomes `englishResources(namespaces)` — the same combine, named for what it returns. The chrome keeps its map, because it negotiates against those keys and needs `en` as a candidate; the worker is handed nothing at all, English already being appended behind every chain by `createTranslator`.

Inlining a locale again is now a deliberate change with a case to make — add it to `BUNDLED_LOCALES`, exclude it from the lazy glob, give the chrome and the worker a way to reach it — rather than filling in a seam left standing open.

The prose the unbundling left stale went with it, in five passes: the i18n, standalone and iframe READMEs said English and Spanish were both carried; `supportedLocales`' generated header and `catalogUtils` pointed at `bundledResources` and `BUNDLED_TRANSLATIONS` for what the bundle holds; `load.ts` told you to register a new inlined locale in a module that no longer exists; the bundle guard said only half of it had anything to say "while every locale that exists is also inlined", which stopped being true the moment these eight landed; and a viewer spec and a Cypress config named Spanish as an inlined locale. The fourth reached the two claims that survived by being about something else: `englishResources` kept `bundledResources`' reason for taking a namespace list — that it "keeps the worker from carrying chrome strings it never draws" — when the worker no longer calls it and `createTranslator` appends all four English namespaces behind every chain anyway; and the i18n README said `createTranslatorFromLocaleData` merges host catalogs over bundled ones, which is exactly what stopped happening. While there, the README's claim that the worker ships `content` + `diagnostics` — twenty lines above the paragraph explaining why `WORKER_NAMESPACES` is `content` alone — was corrected too. The fifth found the rest in the tests themselves — two loader tests carried a note saying they could not run yet, the glob fixture beside them said it "holds nothing today", and a bundle-guard test was named for a state, every locale inlined, that no longer exists — and in two pending changesets that still offered `localeResources` as the way to "correct a bundled translation", of which there are now none.

## Tests that said `fr` and meant "nowhere"

`DoenetViewer.errorLocale.cy.tsx`, `DocViewer/i18n.cy.js` and `styleDescriptionLocale.test.ts` each used French as a stand-in for a locale no catalog answers for. French answers now, so all three say `qaa` instead — ISO 639-3 reserves `qaa`–`qtz` for local use, so no catalog can ever claim it and the assertion cannot rot again.

One goes the other way. "Follows the content's language, not the reader's" could only assert English before, since French had nothing to say; it now asserts `ligne épaisse discontinue rouge` for a `<document lang="fr">` read by a Spanish reader — French because the document is, and *not* Spanish even though the reader is. It was passing in CI against these catalogs by racing: the catalog for an authored `<document lang>` is not known until the source has been parsed, so it arrives after the first paint and the assertion was landing on the English that preceded it. That path — a catalog reaching the *worker* for a language only the document named — is now covered end to end.

`declaredLangsInSource` is covered in `documentLang.test.ts`: it agrees with `readDeclaredLangs` on source that parses, looks past a `lang=` that is not on a document, and answers rather than throwing for source that does not.

Three notes went the same way for the opposite reason — they used a language as a stand-in for something and it has since become real. The i18n README and `lint:i18n`'s roster check walked through "adding `locales/de/`", a directory that is now there; they say `pt`. `loadersFromModules`' fixture explained itself as avoiding a real locale while naming `de` and `fr` — what it avoids is *reading the shipped catalogs*, which is what would break the day a locale gained or lost a namespace. And `stubLoaders` said it stood in for catalogs the repository does not ship, when it installs itself over the loaders whatever the tag.

## Verification

Rendered, not just linted. Every locale was exercised end-to-end through the real `styleDescriptions` pipeline and `createDiagnosticFormatter`:

- Every message in every locale formatted under a real `FluentBundle`: no unresolved `{$name}`, no `{???}`, no runtime error, no newline leaking out of a multiline pattern
- Every argument English passes checked as used, every selector's variants checked against English's, and every DoenetML identifier (`through`, `numDimensions`, `selectFromSequence`, `<answer>`, `symbolicEquality`) checked as surviving verbatim in all eight
- Russian plural categories: `осталось 3 попытки` (few), not the `many` form
- `Intl.ListFormat` in the message's own language: `und` / `et` / `e` / `en` / `и` / `iyo` / `、…和` — `hnj` alone joins in English, ICU having no list data for it
- Style composition across all ten locales, which is how the agreement bugs were found

And the machinery from #1603 was re-checked against real catalogs rather than a probe: 36 code-split chunks (9 locales × 4 namespaces, 524 KB raw) emitted by `@doenet/doenetml`, **zero** catalog text in the entry chunks that carry the library (`doenetml-*.js`, the inline-worker build included) or in the core worker, `dist/locales/` populated with all ten for `@doenet/standalone`, and its bundle guard green — which now means it actively verified that none of the nine translations leaked into the single-file bundle.

`lint:i18n` passes: 559 keys across 10 locales, 364 call sites. The i18n suite is 199 green (the two loader tests that could not run while every translation was inlined now run unconditionally — the guard that skipped them is gone, and the registry test asserts there is a code-split locale left for them to reach for), alongside `@doenet/parser` (297), `@doenet/utils`' style descriptions (26), and the worker's locale suites (111).


dqnykamp:
> The `doenetml-iframe-cypress` failure on this PR is not caused by these changes. Tracked in #1608.

**Resolved — that job is green on this PR now.** The root cause was found and fixed on `main` by #1609: the iframe boot script loaded Comlink from `https://unpkg.com`, and because a static `import` is resolved before any of a module's body runs, a stalled CDN fetch left the whole boot script unexecuted — no handshake sent, the iframe parked on its loading placeholder, no timeout and no error. This branch has been rebased onto that fix.

That also accounts for the symptoms recorded in #1608 that I could not explain at the time: the blank iframe, the core worker never being requested, and a 90 s budget failing to rescue it.

The original evidence that the failure was not caused by these changes still stands, and is kept below for the record.

---

The `doenetml-iframe-cypress` failure on this PR is not caused by these changes. Tracked in #1608.

Short version: this PR branch's own history contains the proof. Run [30558745113](https://github.com/Doenet/DoenetML/actions/runs/30558745113) passed that job 20/20 in 5:01; run [30560716275](https://github.com/Doenet/DoenetML/actions/runs/30560716275), 25 minutes later, failed 11/20 — and the entire diff between the two commits is three comment lines inside `.ftl` catalogs, which Fluent strips at parse time. `@doenet/doenetml` is not even in that suite's runtime module graph (`src/utils.ts` imports it only in `typeof` positions, so esbuild elides it), so the i18n work here cannot reach it.

The underlying fragility reproduces off CI by pinning Cypress to 4 cores, and #1608 records the measurements — including two fixes ruled out (raising the timeout doesn't help; the base64 harness step is a one-time 67 ms) and the harness costs worth attacking.

PR #1619, opened 2026-07-31
Graph Accessibility and Visual Hierarchy Improvements

DoctorHaitch:
# Summary

Implemented a targeted light/dark mode styling update for JSXGraph components to establish a clearer visual hierarchy while improving overall graph readability and accessibility.

## Key Changes

- **Graph text remains unchanged**, continuing to use the original `canvasText` variable, which provides approximately **21:1 contrast** against the background. This preserves maximum readability for labels, annotations, and other primary textual content. Includes Graph Navigation controls.

- **Axes, tick marks, and tick labels** were explicitly targeted and updated to use colors meeting **WCAG 2.2 AAA contrast requirements (~7:1)**. This increases the prominence of structural graph elements while maintaining a clear distinction from primary graph text.

- **Grid lines** were updated to approximately **4.6:1 contrast**, creating a more subtle visual presence that supports graph interpretation without competing for attention.

- Equivalent contrast levels were implemented for both **light and dark themes**, ensuring consistent visual emphasis and readability regardless of color mode.

### Resulting Visual Hierarchy

1. **Graph text and Graph Navigation** (highest emphasis, ~21:1 contrast)
2. **Axes and tick labels** (strong secondary emphasis, ~7:1 contrast)
3. **Grid lines** (supporting visual guidance, ~4.6:1 contrast)

These changes significantly improve graph readability, accessibility, and overall visual clarity by reducing clutter while preserving strong navigational and interpretive cues.

### Commented out Empty Rulesets

Commented out several empty rulesets within the stylesheet. This allows them to be removed during CSS minification, resulting in a modest reduction in the delivered stylesheet size and eliminating unnecessary declarations from the final build.

### DIFF
--- a/.changeset/graph-visual-hierarchy-introduced.md
+++ b/.changeset/graph-visual-hierarchy-introduced.md
@@ -1,0 +1,18 @@
+---
+"@doenet/doenetml": patch
+"@doenet/standalone": patch
+"@doenet/doenetml-iframe": patch
+---
+
+Graphs: introduce a visual hierarchy for grid lines, axes, and labels.
+
+Graph rendering no longer paints all graph elements with `--canvasText`.
+Grid lines now use `--graphGrid`, while axes and tick labels use
+`--graphAxes`; graph navigation controls continue to use `--canvasText`.
+
+This improves visual distinction between graph structure and primary
+content in both light and dark themes while preserving readability and
+accessibility.
+
+Added Cypress regression coverage to verify the hierarchy remains applied
+to axes, tick labels, grid lines, and navigation controls in both themes.
--- a/packages/doenetml/src/DoenetML.css
+++ b/packages/doenetml/src/DoenetML.css
@@ -267,13 +267,15 @@ div.jxgbox input[type="checkbox"] {
 }
 
 /* CSS attributes will (permantely) overwrite attributes set in JSXGraph */
+/*
 .JXGimage {
-    /* opacity: 1.0; */
+    opacity: 1.0;
 }
 
 .JXGimageHighlight {
-    /* opacity: 0.6; */
+    opacity: 0.6; 
 }
+*/
 
 .jxgbox :focus {
     outline-width: 0.5px;
--- a/packages/doenetml/src/DoenetML.css
+++ b/packages/doenetml/src/DoenetML.css
@@ -84,9 +84,8 @@ html {
     --indicatorHoverBlue: var(--lightBlue);
     --buttonHoverBlue: var(--lightBlue);
     --doenetTagColor: #3e844a;
-    --graphElements: #757575;
-    --graphTicks: #595959;
-    --graphText: var(--canvasText);
+    --graphGrid: #757575;
+    --graphAxes: #555555;
 }
 
 [data-theme="dark"] {
@@ -118,9 +117,8 @@ html {
     --indicatorHoverBlue: #5b8cb5;
     --buttonHoverBlue: hsl(209, 54%, 82%);
     --doenetTagColor: #5cb870;
-    --graphElements: #8d8d8d;
-    --graphTicks: #959595;
-    --graphText: var(--canvasText);
+    --graphGrid: #7e7e7e;
+    --graphAxes: #aeaeae;
 }
 .doenet-viewer {
     font-family: "Open Sans" !important;
@@ -267,17 +265,15 @@ div.jxgbox input[type="checkbox"] {
 }
 
 /* CSS attributes will (permantely) overwrite attributes set in JSXGraph */
-/*
+
 .JXGimage {
-    opacity: 1.0;
+    /* opacity: 1.0; */
 }
 
 .JXGimageHighlight {
-    opacity: 0.6; 
+    /* opacity: 0.6; */
 }
-*/
-
-.jxgbox :focus {
+*/ .jxgbox :focus {
     outline-width: 0.5px;
     outline-style: dotted;
 }
@@ -301,7 +297,6 @@ div.jxgbox input[type="checkbox"] {
     color: #666;
 }
 
-/*
 .JXG_navigation_button_left {
 }
 
@@ -334,7 +329,6 @@ div.jxgbox input[type="checkbox"] {
 
 .JXG_navigation_button_screenshot {
 }
-*/
 
 .JXG_navigation_button:hover {
     border-radius: 2px;
@@ -469,21 +463,6 @@ div.jxgbox input[type="checkbox"] {
  */
 .doenet-viewer .doenet-tag {
     color: var(--doenetTagColor);
-}
-
-/* Set default graph contents to ~4.6:1 contrast ratio */
-.jxgbox * {
-    --canvasText: var(--graphElements);
-}
-
-/* Set text components back to Canvas text color. */
-.jxgbox .JXGtext,
-.jxgbox .JXG_navigation .JXG_navigation_button {
-    --canvasText: var(--graphText);
-}
-
-.jxgbox .JXGtext[id*="ticks"] {
-    --canvasText: var(--graphTicks);
 }
 
 /**************************************************
--- a/packages/doenetml/src/DoenetML.css
+++ b/packages/doenetml/src/DoenetML.css
@@ -84,6 +84,9 @@ html {
     --indicatorHoverBlue: var(--lightBlue);
     --buttonHoverBlue: var(--lightBlue);
     --doenetTagColor: #3e844a;
+    --graphElements: #757575;
+    --graphTicks: #595959;
+    --graphText: var(--canvasText);
 }
 
 [data-theme="dark"] {
@@ -115,6 +118,9 @@ html {
     --indicatorHoverBlue: #5b8cb5;
     --buttonHoverBlue: hsl(209, 54%, 82%);
     --doenetTagColor: #5cb870;
+    --graphElements: #8d8d8d;
+    --graphTicks: #959595;
+    --graphText: var(--canvasText);
 }
 .doenet-viewer {
     font-family: "Open Sans" !important;
@@ -293,6 +299,7 @@ div.jxgbox input[type="checkbox"] {
     color: #666;
 }
 
+/*
 .JXG_navigation_button_left {
 }
 
@@ -325,6 +332,7 @@ div.jxgbox input[type="checkbox"] {
 
 .JXG_navigation_button_screenshot {
 }
+*/
 
 .JXG_navigation_button:hover {
     border-radius: 2px;
@@ -459,6 +467,21 @@ div.jxgbox input[type="checkbox"] {
  */
 .doenet-viewer .doenet-tag {
     color: var(--doenetTagColor);
+}
+
+/* Set default graph contents to ~4.6:1 contrast ratio */
+.jxgbox * {
+    --canvasText: var(--graphElements);
+}
+
+/* Set text components back to Canvas text color. */
+.jxgbox .JXGtext,
+.jxgbox .JXG_navigation .JXG_navigation_button {
+    --canvasText: var(--graphText);
+}
+
+.jxgbox .JXGtext[id*="ticks"] {
+    --canvasText: var(--graphTicks);
 }
 
 /**************************************************
--- a/packages/doenetml/src/DoenetML.css
+++ b/packages/doenetml/src/DoenetML.css
@@ -273,7 +273,8 @@ div.jxgbox input[type="checkbox"] {
 .JXGimageHighlight {
     /* opacity: 0.6; */
 }
-*/ .jxgbox :focus {
+
+.jxgbox :focus {
     outline-width: 0.5px;
     outline-style: dotted;
 }
--- a/packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
+++ b/packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
@@ -133,7 +133,7 @@ export function createYAxis({
             position,
             offset,
             anchorx,
-            strokeColor: "var(--canvasText)",
+            strokeColor: "var(--graphAxes)",
             highlight: false,
         };
         if (SVs.yLabelHasLatex) {
@@ -142,7 +142,7 @@ export function createYAxis({
     }
     previousYaxisWithLabelRef.current = Boolean(SVs.yLabel);
 
-    yaxisOptions.strokeColor = "var(--canvasText)";
+    yaxisOptions.strokeColor = "var(--graphAxes)";
     yaxisOptions.highlight = false;
 
     yaxisOptions.ticks = {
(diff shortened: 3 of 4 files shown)


dqnykamp:
Since I'm not great at reasoning with CSS, I asked Claude to review this and identify the errors cause the test failures. I'm next going to ask Claude to comment on this PR with its findings. I hope the comment ends up being understandable.

dqnykamp (commented):
## Review

The three CI failures are caused by this PR, not flake, and two of them are accessibility regressions against guards that were added deliberately (#396/#397). The mechanism (`.jxgbox * { --canvasText: … }`) is too blunt for the effect described in the PR body.

### 1. What the PR does

Adds three theme variables (`--graphElements`, `--graphTicks`, `--graphText`) and then, at `DoenetML.css:475-487`, re-scopes `--canvasText` for everything inside a graph, with two exceptions re-raising it back for text and the navigation bar. It also comments out two blocks of empty JSXGraph rulesets.

Because DoenetML passes the literal string `var(--canvasText)` into JSXGraph stroke attributes and into many component CSS files, redefining that variable on `.jxgbox *` changes far more than axes/ticks/grid.

### 2. Failing CI — all three are this PR

| Spec | Assertion | Actual |
|---|---|---|
| `accessibility/darkModeVisibilityRegressions.cy.js:91` | graph focus outline must resolve to white in dark mode (#396) | `rgb(141,141,141)` = `--graphElements` |
| `tagSpecific/mathinput.cy.js:1096` | in-graph mathInput text must be light in dark mode | `rgb(141,141,141)` |
| `tagSpecific/textinput.cy.js:97` | in-graph textInput border must match a plain disabled input | `rgb(117,117,117)` vs expected `rgb(0,0,0)` |

Root cause for each:

- `DoenetML.css:238-241` — `.jxgbox :focus, .jxgbox :focus-visible { outline: 2px solid var(--canvasText) }`. That rule exists specifically because JSXGraph's own outline is invisible on a dark canvas; the focused element matches `.jxgbox *`, so the outline is now gray. Worse, a gray ring around a now-gray axis or handle can drop well under 3:1 against *adjacent* colors (SC 1.4.11 / 2.4.11 measure against what is next to the indicator, not just the canvas).
- `mathInput.css:48-52` — `.mathInputGraphWrapper { color: var(--canvasText) }` sits *inside* the `.JXGtext` rendNode, so the `.JXGtext` exception never reaches it. This is the general flaw in the approach: any nested element re-matches `.jxgbox *`.
- `textInput.tsx:222,237-238` — border/stroke from `var(--canvasText)`, so an in-graph text input no longer matches the styling of an identical input outside a graph.

Same class of collateral damage, not covered by tests: `polygon.tsx:442` and `polyline.tsx:395` (vertex `fillcolor`), `booleanInput.tsx:138-139` and `booleanInput.css:51,85`, and JSXGraph smart labels — `smartlabel.js` sets `cssClass: 'smart-label-pure …'` which **replaces** `JXGtext`, so the dark-mode fix at `DoenetML.css:438-440` is silently undone too.

### 3. The tick-label rule is dead code

`.jxgbox .JXGtext[id*="ticks"]` (`DoenetML.css:485`) matches nothing. JSXGraph builds tick ids as `board.setId(this, 'Ti')` (`ticks.js:159`) → `attr.id = this.id + ld.i + "Label" + counter` (`ticks.js:1561`) → DOM id `container.id + "_" + el.id` (`renderer/abstract.js`), e.g. `g_gTi43Label1`. Attribute-value matching is case-sensitive, so `"ticks"` never appears. Net effect: tick labels keep full `--canvasText` (21:1), never the intended 7:1.

It is also fragile in the other direction — an author writing `<point name="ticksA">` would produce an id containing `ticks` and get their label recolored. Substring matching on generated ids should not be load-bearing.

### 4. Accessibility, measured

Computed against the actual canvases (`--canvas`: `#fff` / `#121212`, applied to the `.jxgbox` div via `GraphFrame.tsx:66`):

| Element | Light | Dark | Note |
|---|---|---|---|
| `--graphElements` #757575 / #8d8d8d | 4.61:1 | 5.64:1 | ≥3:1, passes SC 1.4.11 for lines |
| `--graphTicks` #595959 / #959595 | 7.0:1 | **6.25:1** | dark misses the AAA 7:1 claimed (needs ≈ #9c9c9c) |
| Grid lines (same color, `strokeOpacity: 0.5`) | **1.94:1** | **2.32:1** | was 3.95:1 / 5.28:1 |
| Focus outline | 4.61:1 | 5.64:1 | was 21:1 / 18.7:1; breaks the #396 guard |
| In-graph mathInput text | 4.61:1 | 5.64:1 | scrapes past AA 4.5:1 for user-entered text |

So: no hard WCAG *minimum* is violated for lines, but the described hierarchy is not what ships. Axes/tick marks land at 4.6:1 (not 7:1), tick labels stay at 21:1 (not 7:1), and grid lines land near 2:1 (not 4.6:1) because `jsxgraph.ts:159` applies `strokeOpacity: 0.5` on top of whatever color is chosen. Note also that grid lines *are* the major ticks (`majorHeight = -1`, `jsxgraph.ts:170-178`) — one JSXGraph object, one stroke color — so grid and tick marks cannot be given different contrast levels by color alone. That constraint needs a different implementation, not a different variable.

### 5. Suggested approach

Drop the `.jxgbox *` override and color the graph elements at the point of use instead — `utils/jsxgraph.ts` already sets every axis/tick/grid color explicitly:

```js
yaxisOptions.strokeColor = "var(--graphAxis)";
yaxisOptions.ticks.strokeColor = "var(--graphGrid)";        // grid + tick marks, opacity 0.5
yaxisOptions.ticks.label.strokeColor = "var(--graphTickLabel)";
```

That gets exactly the three targets the PR body describes, touches nothing else, and leaves `--canvasText` meaning "canvas text" everywhere. If grid lines need to be more subtle than tick marks, raise `strokeOpacity` toward 1 and encode the subtlety in `--graphGrid` itself, so the effective ratio is the one you picked.

### 6. Smaller points

- **Missing changeset.** AGENTS.md §Changesets: user-facing changes need one (patch, per convention while < 1.0). This is user-facing.
- **Commented-out empty rulesets buy nothing.** Verified: `esbuild --minify` already drops `.a {}` entirely, and Vite minifies CSS with esbuild by default. The blocks also mirror upstream JSXGraph's stylesheet, so keeping them intact keeps future upstream diffs readable. I would revert both hunks.
- **`--graphText: var(--canvasText)`** is a pure alias defined next to `--canvasText` in both themes; if the `.jxgbox *` override goes away, it can go too.
- **Trailing whitespace** after `opacity: 0.6;` (`DoenetML.css:276`).
- **Test coverage.** Whatever lands should add a spec under `test-cypress/cypress/e2e/accessibility/` asserting the intended hierarchy (axis stroke, tick-label color, grid color, both themes) — right now nothing would catch the dead `[id*="ticks"]` selector. The three existing failures should be fixed by changing the CSS, not by relaxing the assertions.

dqnykamp:
OK, that was a long comment. Does it more-or-less make sense? I imagine implementing the suggestions are more than you bargained for. Let me know how you want to proceed. I'm happy to help out.

DoctorHaitch:
I actually think the suggested fix is the right approach. At the time, I just didn't know where those attributes were defined, so having the AI point out the exact file that needs to be updated for strokeColors was helpful.

We already added the correct token colors; this change would simply provide a cleaner implementation for applying them. It shouldn't be difficult to implement. I'll try it locally later and update this PR if everything works as expected.

I also suspect the AI is miscalculating some of the resulting cascades and contrast ratios, but that doesn't materially affect this change. Those values should be reflected more accurately once we follow its recommendation and apply the specific tokens directly instead of relying on the universal token that was previously being used.

DoctorHaitch:
I may need help on the final point: Test Coverage, but I should be able to handle everything else.

dqnykamp:
Sounds good. Let me know if you need assistance. Or mark ready for review and request a review from me when it's ready.

DoctorHaitch:
@dqnykamp I figured out how to get Cypress running in my dev container and added both regression coverage and a changeset.

I also updated the implementation to introduce dedicated graph hierarchy tokens in the emitted JSXGraph elements rather than reusing `--canvasText` everywhere. The original approach appeared to work in my local testing because I was also applying renderer-specific CSS overrides for components such as `mathInput` and `textInput`, which masked the cypress testing issues above. Those overrides were not present in the standalone/test environment, so some targets that should not change in the core DoenetML were changing, which wasn't an issue for my implementation as I was already changing them elsewhere.

The new implementation emits:

- `--graphGrid` for grid lines
- `--graphAxes` for axes and tick labels
- `--canvasText` for navigation controls and primary text

and the new Cypress regression test verifies that hierarchy in both light and dark mode so we catch future selector or token regressions.

dqnykamp:
## Re-review

I fetched the branch and compared against the earlier review. **The substantive issues are addressed.** All 24 CI checks pass on `d80efaa`, including the three specs that previously failed.

The diff is now only 4 files, and the blunt override is gone:

| Prior finding | Status |
|---|---|
| `.jxgbox * { --canvasText: … }` re-scoping | **Removed entirely** — no `jxgbox *` rule remains anywhere |
| Focus outline regression (#396) | Fixed by consequence — `.jxgbox :focus` still resolves `--canvasText` |
| in-graph mathInput / textInput regressions | Fixed by consequence — those files are untouched |
| Dead `.JXGtext[id*="ticks"]` selector | Gone; tick labels now colored at the point of use (`jsxgraph.ts:153,244`) |
| Uncovered collateral (polygon/polyline/booleanInput, smart labels) | No longer possible — nothing outside `jsxgraph.ts` changed |
| Commented-out empty rulesets | Reverted |
| Trailing whitespace | Gone |
| Missing changeset | Added; package list (`doenetml`, `standalone`, `doenetml-iframe`) matches repo conventions |
| Test coverage | Added `accessibility/graphVisualHierarchy.cy.js` |
| Stray alias `--graphText` | Gone; only `--graphGrid` / `--graphAxes` remain |

The implementation is exactly the suggested approach — colors applied in `utils/jsxgraph.ts` at the point of use, with `--canvasText` left meaning "canvas text" everywhere.

### Two remaining points (neither blocking)

**1. Grid/tick contrast is now lower than on `main`, because `strokeOpacity: 0.5` still stacks on top of the new token** (`jsxgraph.ts:158,249`). Measured against `--canvas`:

| | light | dark |
|---|---|---|
| `--graphAxes` (axes, tick labels) | 7.46:1 | 8.44:1 — matches the stated hierarchy |
| `--graphGrid` nominal | 4.61:1 | 4.61:1 |
| `--graphGrid` **as rendered** (×0.5 opacity) | **1.94:1** | **2.05:1** |

Previously that was ~3.95:1 / ~5.3:1. And since `majorHeight = -1` only for `grid="medium"/"dense"`, with the default `grid="none"` the *real tick marks* — one JSXGraph object sharing `ticks.strokeColor` — also render at ~1.94:1. Tick marks are meaningful structure, so 3:1 (SC 1.4.11) is the number I would want there. Fix is the same as before: raise `strokeOpacity` toward 1 and encode the subtlety in `--graphGrid` itself, so the effective ratio is the one you picked.

**2. The new spec tests token assignment, not hierarchy.** It asserts on the literal attribute strings (`stroke="var(--graphAxes)"`, `path[stroke="var(--graphGrid)"]`). Those strings are theme-independent, so the dark-mode `it()` runs the same assertions as the light one and adds no coverage beyond `[data-theme="dark"]` existing — while the file header claims it enforces `graphGrid -> graphAxes -> canvasText`. Reading resolved colors via `getComputedStyle` (as `darkModeVisibilityRegressions.cy.js` does) and asserting the ordering would make the test match its own docstring, and would have caught the opacity issue above. Also, `.JXG_navigation_button` uses `.each()`, which passes vacuously if those buttons ever stop rendering.

Trivia: the changeset has no trailing newline, and there is a stray added blank line at `DoenetML.css:268` left over from reverting the comment-out hunks.

dqnykamp:
### Follow-up: the grid token only reaches one of the two grid code paths

Expanding on point 1 above. `<graph>` has two distinct ways to draw a grid, and this PR only styles one of them.

**Path A — `grid="medium"` / `grid="dense"`.** No separate grid object exists; the axis ticks *are* the grid (`majorHeight = -1`, `jsxgraph.ts:171-178, 265-272`). These pick up the new `--graphGrid`, but `strokeOpacity: 0.5` (`jsxgraph.ts:158, 249`) still multiplies it:

| | nominal | **as rendered** |
|---|---|---|
| light `#757575` on `#fff` | 4.61:1 | **1.94:1** |
| dark `#7e7e7e` on `#121212` | 4.61:1 | **2.05:1** |

Note this also drags down the *real tick marks* under `grid="none"` (the default), since they share `ticks.strokeColor`.

**Path B — a numeric grid, e.g. `grid="1 1"`.** `SVs.grid` becomes an array, which takes a completely different branch: `JSXGraphRenderer.tsx:63-77` sets only `JXG.Options.grid.gridX/gridY` and passes `grid: true` to `initBoard`, and `useGridAndAxesSync.ts:51-54` creates the object with `board.create("grid", [], { gridX, gridY })` — no `strokeColor`, no `strokeOpacity`. Nothing in `packages/doenetml/src` ever sets `JXG.Options.grid.strokeColor`. So it falls back to JSXGraph's built-in defaults (`strokeColor: '#c0c0c0'`, `strokeWidth: 1`, `strokeOpacity: 0.5`):

| | effective color | vs canvas |
|---|---|---|
| light (`#fff`) | `#c0c0c0` @0.5 → `#e0e0e0` | **1.32:1** |
| dark (`#121212`) | `#c0c0c0` @0.5 → `#696969` | **3.41:1** |

That is a fixed gray that **ignores the theme entirely** — brighter than the canvas in dark mode, nearly invisible in light mode, i.e. inverted relative to the intent. It also means a numeric grid and `grid="medium"` render as visibly different greys on the same page.

So the changeset's "grid lines now use `--graphGrid`" holds only for the keyword grid modes.

### Suggested fix, covering both

1. **Path B** — pass the token when creating the grid object in `useGridAndAxesSync.ts`:

```js
board.create("grid", [], {
    gridX: SVs.grid[0],
    gridY: SVs.grid[1],
    strokeColor: "var(--graphGrid)",
    strokeOpacity: 1,
});
```

2. **Path A** — drop `strokeOpacity: 0.5` to `1` in both `createXAxis` and `createYAxis`, so the rendered ratio is the one the token encodes.

3. Then pick `--graphGrid` for the contrast you actually want, since opacity is no longer distorting it. For ≥3:1 (SC 1.4.11 territory, and a sensible floor for tick marks, which share this color): `#959595` on the light canvas and `#606060` on the dark canvas both land at 3.0:1. Those stay clearly subordinate to `--graphAxes` at 7.46:1 / 8.44:1, so the intended hierarchy survives.

With both paths using the same token at full opacity, the two grid forms finally match each other and follow the theme.

If it is useful, extending the new `graphVisualHierarchy.cy.js` with a `grid="1 1"` case — and asserting resolved `getComputedStyle` colors rather than the literal `var(...)` attribute strings — would lock both paths down.

PR #1622, opened 2026-08-01
Try out the new math-expressions

siefkenj:
# Switch to the Rust/WASM math-expressions engine

> ## Merge order — steps 1 and 2 are done; this PR is step 3
>
> 1. ~~Merge [Doenet/math-expressions#84](https://github.com/Doenet/math-expressions/pull/84).~~ **Done.**
> 2. ~~Publish `math-expressions@3.x` to npm.~~ **Done** — `math-expressions@3.0.0-alpha.1` is on the registry and is `latest`.
> 3. Merge this. **Nothing left to edit first.** Step 6 of `MATH_EXPRESSIONS_RUST_MIGRATION_PLAN.md` is complete: the submodule is gone, `packages/math` depends on the published package, and the range a consumer installs (`^3.0.0-alpha.1`) is already in the tree.
>
> The one thing to know about that range: it is named in **`packages/doenetml/vite.config.ts`**, as `publishRanges` handed to `createPackageJsonTransformer`, and *not* in `packages/doenetml/package.json`, which keeps saying `file:../math` along with the other six manifests. That was measured, not preferred — declaring the registry range in the manifest makes npm install the published package *inside* `packages/doenetml`, shadowing the seam there while every sibling still resolves it, and `tsc` then fails on `initMathWasm`.
>
> ⚠️ **The dot in `-alpha.1` is load-bearing, and `^3.0.0` would not match it.** npm's caret range excludes prereleases, so the range must name one; and semver compares an undotted prerelease tail as text, so `^3.0.0-alpha1` would match neither `alpha10` nor `alpha11`. When upstream publishes a non-prerelease `3.0.0`, change the one constant in `packages/doenetml/vite.config.ts` to `^3.0.0`. Nothing in either build checks this; it is a convention, deliberately.

Replaces the legacy `math-expressions@2.x` JavaScript library with the Rust core compiled to WASM, via the `math-expressions-js-compat` drop-in from [Doenet/math-expressions#84](https://github.com/Doenet/math-expressions/pull/84), now installed from npm as `math-expressions@3.0.0-alpha.1`. **Review #84 first** — this PR is its downstream half. The switch is permanent: Rust is the only supported configuration, and to A/B against the old engine you check out a commit before this branch. **The mechanism is an alias, not a codemod** — `@doenet/math` (`packages/math`) backs the bare specifier, every consuming `package.json` says `"math-expressions": "file:../math"`, and no file that imports `math-expressions` changes.

**Building needs no toolchain it did not need before.** The published tarball ships both wasm-bindgen targets, so nothing here compiles the engine: `packages/math/scripts/build-wasm.mjs` reads the `.wasm` out of `node_modules` and base64-inlines it. `.github/actions/setup-math-wasm`, the `wasm-toolchain` devcontainer feature, every `submodules: recursive` checkout and the `git submodule update` in `postCreateCommand.sh` are gone. `.devcontainer/` is now byte-for-byte `main`'s; `.github/` is too apart from three riders listed below. `npm run build` still reaches `wasm-pack` for `packages/doenetml-worker-rust` — DoenetML's own core — exactly as `main` always has, and `wasm-pack` brings its own target and bindgen. A side effect worth having: the bytes a bundle inlines are now the ones the lockfile pins, so a local build and a CI build carry an identical engine. They did not while it was compiled here.

**Consumers of the published `@doenet/doenetml` now install one more package.** The engine is a peer dependency rather than a private copy inside each `@doenet/*` library, so an application embedding several of them resolves one engine rather than one per library: `npm install math-expressions@^3.0.0-alpha.1`, and nothing else — no code and no initialization step. `packages/doenetml/src/utils/mathWasm.ts` is where the two worlds meet: in this repository the peer is `@doenet/math` and its `initMathWasm` starts the inlined core; for a consumer the peer is the published package, which has no such export, so the fallback instantiates its `--target web` glue and hands it to upstream's `setWasmModule`. The trade-off for a consumer is that the binary is fetched as an ordinary bundler asset rather than inlined — a host that cannot make a same-origin request for it (a `srcdoc` or blob-URL document) should use `@doenet/standalone`, which carries everything in one file.

**Status**: CI is green at `ea6629e7`, the last commit that changes any code a build compiles — 23 passing, 1 skipped, 0 failed, including every Cypress group. The review-cycle commits after it change comments and documentation, drop one stray blank line, and add one vitest file: `packages/doenetml/src/utils/mathWasm.test.ts`, which covers the published-consumer fallback described above and was the one path in this diff with no test at all (5 tests; `packages/doenetml` 237/237, `packages/math` 31/31, `npm run typecheck` clean). #84 is green at the commit the published tarball was cut from, and its `package publishability` job packs the tarball, installs it into a throwaway project outside the workspace and drives both consumer paths.

**User-visible behavior changes** are in `.changeset/rust-math-expressions-engine.md` — read it rather than a summary here. `.changeset/piecewise-domain-without-intervals.md` covers an unrelated pre-existing crash fixed in passing.

**Where to start reviewing**: `packages/math/` (the seam — `engine-rust.ts` is a straight re-export, `wasm-loader.ts` carries the one ordering constraint), then `packages/doenetml/src/utils/mathWasm.ts` and `packages/doenetml/vite.config.ts` (the published-consumer path), then `MATH_EXPRESSIONS_ENGINE_NOTES.md`, then the changesets. The worker-javascript component and test churn is mostly the NaN-boundary convention and unpadded-delimiter expectations the notes describe.

**In this diff but not the engine switch**, kept and called out: a `<video>` source-change fix; the memory-bench instruments and two Cypress specs from the memory-reduction workstream; a bash-3.2 rewrite of `.husky/format-staged.sh`; the piecewise-domain crash fix; and three in `.github/` — a `tsc --noEmit` gate over every workspace (`scripts/typecheck.mjs` plus a `Type check` step in `ci.yml`, with today's unclean packages listed and their error counts pinned), a `pull-requests: read` scope on the `Detect Changed Paths` job without which `paths-filter` 403s on every PR, and a correction to `.github/skills/changesets/SKILL.md` about what `transform-package-json.ts` emits. The case for splitting each into its own PR is in `MATH_EXPRESSIONS_ENGINE_NOTES.md`, "What is still riding along, and should not be".

**The residual risk, plainly**: no differential grading harness exists, which is the primary risk of an engine swap — the suites are green, and the review still found **eighteen** wrong-answer-on-grading defects that no existing test named. All eighteen are fixed. The confidence argument is structural rather than exhaustive: most of them shared one cause — `evaluate_to_constant` answering `null`, which coerces to `0` in JavaScript, where the legacy library answered `NaN` — and that sentinel was fixed at its source rather than one call site at a time, which is what makes the unswept reads safe by default. The eighteen, the sweeps and the guard-audit method are in `MATH_EXPRESSIONS_ENGINE_NOTES.md`, "Known risks and open product decisions".

**Four documents carry the detail**: `MATH_EXPRESSIONS_ENGINE_NOTES.md` (architecture, the conventions the switch established, known risks, follow-up PRs, what is riding along), `MATH_EXPRESSIONS_RUST_MIGRATION_PLAN.md` (the two-stage plan and the Step 6 record), `MATH_EXPRESSIONS_UPSTREAM_REQUESTS.md` (engine-level divergences still open upstream), and, in #84, `active-plans/PR84_REVIEW_KNOWN_ISSUES.md`. The pass-by-pass review history is in the git log (`Review cycle N:` commits) and in this description's edit history (GitHub web UI only).

🤖 This description was written by an agent during review; the findings ledger lives in the documents above.


dqnykamp:
Right now, merges to main auto-publish to the `dev` npm tag and are queued to be published to the `latest` tag on next 0.7.x release. I think the backward-incompatibility of a new math-expressions should not get folded into the 0.7 series. I haven't thought about how to maintain a separate 0.8 line. For now, we should create a separate 0.8 branch and this PR should target that. We'll have to periodically merge in the main branch to keep it in sync.

I'll have to learn the best ways to maintain multiple lines. If you have suggestions, let me know!

dqnykamp:
Or maybe, we release one more `0.7` version, create a `0.7` for any backport fixes, and publish `0.8` dev versions from now on off of `main`.

PR #1629, opened 2026-08-01
Minor bug fix -- adds consistent application of .trim() in breakstrings.js

### LINKED ISSUE #1617: [Bug]: `tupleList` Equality Depends on Position of Closing `</tupleList>` Tag
### Your DoenetML

```xml
<graph>
  <vector name="v1" />
</graph>
<boolean>
  <tupleList unordered name="blah">
    ($v1.head.x,$v1.head.y) (4,5,6) (1,2)
  </tupleList>
  = <tupleList name="reblah">(1,0) (4,5,6) (1,2)</tupleList>
</boolean>
```

### What happened?

# Summary

A `<tupleList>` can evaluate differently depending on whether its closing `</tupleList>` tag appears on the same line as the final tuple or on its own line. This causes equality comparisons between tuple lists to fail unexpectedly.

This is particularly problematic because CodeMirror's **Format as DoenetML** feature formats longer tuple lists with the closing tag on its own line, producing the broken behavior automatically.

---

## Affected DoenetML

The following evaluates the boolean expression as **false**:

```doenetml
<graph>
  <vector name="v1" />
</graph>
<boolean>
  <tupleList unordered name="blah">
    ($v1.head.x,$v1.head.y) (4,5,6) (1,2)
  </tupleList>
  = <tupleList name="reblah">(1,0) (4,5,6) (1,2)</tupleList>
</boolean>
```

---

## Actual Behavior

The boolean evaluates to:

```text
false
```

even though the tuple lists should be equivalent.

The failure is triggered when the closing `</tupleList>` tag appears on its own line.

Notably, the same issue occurs if the second `<tupleList>` is reformatted similarly. The problem is not caused by inconsistent formatting between the two tuple lists, but by the closing tag being placed on a separate line.

---

## Expected Behavior

Whitespace and line breaks immediately preceding `</tupleList>` should not affect the contents of a tuple list.

The expression should evaluate identically regardless of whether the closing tag appears:

```doenetml
(1,2)</tupleList>
```

or

```doenetml
(1,2)
</tupleList>
```

In both cases, the boolean should evaluate to:

```text
true
```

---

## Working Version

When the closing tag is moved onto the same line as the final tuple, the comparison behaves as expected:

```doenetml
<graph>
  <vector name="v1" />
</graph>
<boolean>
  <tupleList unordered name="blah">
    ($v1.head.x,$v1.head.y) (4,5,6) (1,2)</tupleList>
  = <tupleList name="reblah">(1,0) (4,5,6) (1,2)</tupleList>
</boolean>
```

---

## Steps to Reproduce

1. Create a DoenetML document containing:

   ```doenetml
   <graph>
     <vector name="v1" />
   </graph>
   <boolean>
     <tupleList unordered name="blah">
       ($v1.head.x,$v1.head.y) (4,5,6) (1,2)
     </tupleList>
     = <tupleList name="reblah">(1,0) (4,5,6) (1,2)</tupleList>
   </boolean>
   ```

2. Observe that the boolean evaluates to `false`.

3. Move the closing `</tupleList>` tag to the end of the final tuple:

   ```doenetml
   <tupleList unordered name="blah">
     ($v1.head.x,$v1.head.y) (4,5,6) (1,2)</tupleList>
   ```

4. Observe that the boolean now evaluates to `true`.

---

## Impact

This bug causes tuple-list equality checks to depend on formatting rather than content.

Because CodeMirror's **Format as DoenetML** moves closing tags onto their own lines for longer tuple lists, the built-in formatter can transform a working document into one that evaluates incorrectly. This makes the issue difficult to diagnose and can lead to incorrect assessment behavior or debugging confusion for authors.

---

## Suspected Cause

The parser appears to be treating the trailing newline and/or whitespace before `</tupleList>` as semantically significant content. As a result, the internal representation of the tuple list differs depending on whether the closing tag immediately follows the final tuple or appears on a separate line.

Tuple-list parsing should ignore trailing whitespace before the closing tag so that formatting changes do not affect semantic behavior.

### What version of DoenetML are you using?

0.7.21

### What browsers are you seeing the problem on?

Firefox

DoctorHaitch:
# Summary

This PR fixes inconsistent whitespace handling when parsing tuple/list component strings in `breakstrings.js`.

Specifically, `.trim()` was added to several component string parsing paths in:
`packages/doenetml-worker-javascript/src/components/commonsugar/breakstrings.js`

This brings the implementation into alignment across related parsing functions, ensuring component strings are treated consistently regardless of surrounding whitespace.

## Motivation

[Issue #1617](https://github.com/Doenet/DoenetML/issues/1617) exposed cases where tuple/list parsing behavior differed depending on which helper function processed the input. Some code paths trimmed surrounding whitespace while others did not, leading to inconsistent parsing results and unexpected behavior.

By normalizing component strings with `.trim()` throughout the relevant parsing logic, tuple/list handling becomes more predictable and consistent.

Fixes #1617.

## Changes

- Added `.trim()` to several component string parsing operations in `breakstrings.js`
- Aligned whitespace handling across tuple/list parsing functions
- Removed inconsistencies between equivalent parsing paths
- Improved robustness when tuple/list components contain leading or trailing whitespace

## Impact

This is a low-risk change that affects preprocessing of component strings prior to parsing.

Expected benefits include:

- Consistent tuple/list parsing behavior
- Correct handling of leading/trailing whitespace across all parsing paths
- Resolution of Issue #1617
- Prevention of related whitespace-induced parsing bugs that could arise from inconsistent implementation details

## Testing

Verified that tuple/list expressions containing leading and trailing whitespace are parsed consistently across the affected code paths and produce the same results as equivalent trimmed inputs.

### DIFF
--- a/packages/doenetml-worker-javascript/src/components/commonsugar/breakstrings.js
+++ b/packages/doenetml-worker-javascript/src/components/commonsugar/breakstrings.js
@@ -221,7 +221,7 @@ export function breakEmbeddedStringsIntoParensPieces({
             continue;
         }
 
-        let s = component;
+        let s = component.trim();
 
         let beginInd = 0;
 
@@ -292,7 +292,7 @@ export function breakIntoPiecesBySpacesOutsideParens({ componentList }) {
             continue;
         }
 
-        let s = component;
+        let s = component.trim();
 
         let beginInd = 0;
 


dqnykamp (changes requested):
Great, can you add a regression test that fails without this change and passes with it?

For example, in `boolean.test.ts`, you could use this DoenetML
```
<boolean name="b">
  <tupleList>
    (1,0) (1,2)
  </tupleList>
  = <tupleList>(1,0) (1,2)</tupleList>
</boolean>
```
and assert that the `value` state variable of `b` is true.

PR #1654, opened 2026-08-05
Seed catalogs for eight indigenous languages of the Americas

dqnykamp:
Add message catalogs for Haitian Creole, Quechua, Guarani, Aymara, Nahuatl, Kʼicheʼ, Mapudungun and Ojibwe.

Each covers all four namespaces — the viewer chrome, the editor and language-server surfaces, the prose the core computes into a document, and the warnings and errors. `documentLocale="qu"` and `<document lang="ht">` work with nothing configured, and all eight reach `<document lang>`'s autocomplete.

These are **unreviewed machine-generated seeds**, and every file says so in its header. Nothing falls back silently: a key a translation is missing renders in English, which is what makes seeding safe.

These eight are the roster's first languages **indigenous to the Americas**. The gap they close was not a long tail — it was a whole continent that a 124-locale roster with every European national and regional language had nothing at all from.

## The fallback bug this uncovered

This is the first batch to ship codes that stand for **more than one individual language**, and doing so found a real bug rather than merely needing a note. `qu`, `ay`, `gn` and `oj` are ISO 639-3 macrolanguages and `nah` is an ISO 639-3 collection code, each covering a family of individual languages with codes of their own. **CLDR's likely-subtags folds exactly one member of a macrolanguage to it and leaves the rest unresolvable.** So `quz` reached `qu` on ICU data alone and `quh` did not; `ojg` reached `oj` and `ojb` did not; `gug` reached `gn` and `gui` did not. A Bolivian Quechua reader arriving under `quh` was getting English with a `qu` catalog sitting on disk, and nothing said why.

`MACROLANGUAGE_MEMBERS` in `negotiate.ts` closes it, keyed on **published membership** — ISO 639-3's for the four macrolanguages, ISO 639-5's grouping of the Nahuan languages for `nah` — which is what makes it checkable rather than a matter of taste, and what distinguishes it from the `nn` and `fat` cases the alias map deliberately leaves out: neither of those is a member of `nb` or `ak`. The member CLDR already folds is listed anyway, so each list reads as the whole of a group rather than the leftovers of one, and so a change in ICU data cannot silently drop a code out of coverage. `ciw` is the one worth naming: Chippewa is the variety the Fiero orthography `locales/oj` is written in was devised for, so it is the member the catalog answers best and the one a reader is likeliest to arrive under. Algonquin is the nearest miss and the other one worth naming: it is often called a dialect of Ojibwe and is mutually intelligible with the Ontario varieties, but ISO 639-3 gives it `alq` outside the `oj` macrolanguage, so it falls to English like Fante does — the rule doing its job rather than a gap in the list, since the moment membership becomes a judgement about how close two varieties sound nothing in the map is checkable any more.

## What the eight are not

They are not one linguistic story, and the interesting thing is how little they share beyond a hemisphere.

**Ojibwe is the batch's real `$gender` catalog, and the third mechanism that argument has carried.** Ojibwe has grammatical gender and it is not masculine and feminine — it is **animate against inanimate** — and the words that describe a thing are *verbs* agreeing with it. So `noun-gender` answers `anim` or `inan` and every colour and width selects on it: «miskozi» for an animate red thing, «miskwaa» for an inanimate one. That is the `locales/sw` precedent exactly, and it is why the argument was named for a position rather than for a case. Which shapes are which is not derivable from anything — a star is animate in Ojibwe and a rock is not — so the assignment in `noun-gender` is a guess the header explicitly asks a speaker to check, because every adjective downstream follows it.

**Ojibwe is also the only one of the eight that keeps its counted selects.** It marks the inanimate plural and the verb agrees with it, so «bezhig» and several really are two sentences. The other seven drop those selects — Creole has no plural suffix, Quechua and Aymara and Guarani and Mapudungun drop theirs after a numeral, Nahuatl and Kʼicheʼ do not pluralize inanimates at all — and each site says which. Ojibwe drops a handful of its own, all in `diagnostics.ftl` and all the same shape: a message that is nothing but the negated frame «gaawiin … -sinoon» and a placeable has no noun for «-an» to land on and no second form to select, so those sites are written as one string with a comment saying so. That the files differ here is a fact about the languages, not about how carefully they were written.

**Six of the eight put their adjectives in front of the noun, which is English's order.** After six Romance catalogs in #1653 that all had to invert it, that is the useful contrast: the `$part` argument's shape was never about Romance. In all six the side count is a prenominal modifier, so the head carries it and `noun-regular-polygon`'s `[tail]` branch renders empty. `styleDescriptions.test.ts` pins that as an **identity** rather than as a difference, which is what would catch someone "correcting" one of these catalogs by moving its adjectives behind the noun on the assumption that a non-European language must want them there.

**Haitian Creole is the roster's most analytic catalog, and it reaches `[noun-tail]` anyway.** Not one message in it selects on `$gender` or `$role`: no gender, no case, no adjective agreement, no plural suffix. And it still splits `noun-regular-polygon` the way `locales/es` does — head «poligòn regilye», tail «ki gen 5 kote» — purely so the side count does not sit between the noun and the words describing it. So that branch is now exercised by a language with **no agreement to protect**, which shows the `$part` split to be about word order as much as about agreement.

**The affix rule bites in four new shapes.** Quechua and Aymara mark every case relation with a suffix, and the constraint turns out not to bite at most sites for a reason worth recording: the word a suffix lands on is one the catalog writes — «manyayuq» for the border, «p'anqamanta» for the page count — never the value. Where that was not true, `style-filled`'s comitative «-wan»/«-mpi» on `{ $pattern }`, the catalogs name the value instead, after the Andean words for a woven design («pallaynin», «saltanakapa»). Guarani is sharper: **nasal harmony decides which allomorph a suffix takes** («-kuéra» against «-nguéra») from the value's last syllable, so it is not a fixed ending welded to an unknown word but a *choice* made from one. Nahuatl is sharper again in the other direction: a noun loses its absolutive «-tl»/«-tli»/«-li» the moment it is compounded, and which ending it had is a fact about the noun, so no compound can be assembled around a placeable and every message writes two words where the language would write one.

**Kʼicheʼ is the roster's first ergative–absolutive language, and it brings the first *prefix* that cannot be welded.** The possessive marker is «u-» before a consonant and «r-» before a vowel — its shape decided by the **first sound of the word after it**. Every entry in the README's table for this constraint is a suffix decided by the word before; this is the same problem seen from the other end, and the nearest thing to it in the roster is `locales/lb`'s Eifeler Regel, which is also decided by what follows but as a deletion rather than a choice. Nowhere do these files put «u-»/«r-» on a value: where the English possessed one, they write the free relational «rech».

**Four of the eight cannot fit English's colour table, and they fail to fit it four different ways.** This is the batch's most substantive limitation and each header says so rather than leaving a reader to think the entries are merely wrong:

- Ojibwe is the worst fit in the whole roster: «ozhaawashko-» covers blue *and* green, and «ozaawi-» covers yellow *and* orange *and* brown, so **five of English's twelve keys are served by two Ojibwe stems**. The catalog narrows them with the modifiers the language itself uses («ozhaawashko-gizhigaa», of the sky) and says plainly that no amount of correcting individual entries fixes the shape of the table.
- Guarani «hovy» spans blue and green with «hovyũ» at the darker end; the catalog assigns `.blue` and `.green` across that pair, which is what Paraguayan school materials do and is not a translation of either English word.
- Aymara and Quechua divide the same range at *different* points from each other — «larama»/«ch'uxña» against «anqas»/«q'umir» — two neighbouring languages, two partitions, neither of them English's.
- Mapudungun is the one that does line up: it distinguishes «kallfü» from «karü», where the other four do not.

**Two catalogs are honest about leaning on Spanish, and say where the boundary is.** Guarani is written in the *jopara* register the Ministerio de Educación's own materials use — Spanish loans where Guarani has no term, rather than coinages — and Mapudungun's seam runs between its colours and its geometry: the colour words, width words and everyday verbs are Mapudungun, and the geometry nouns are largely Spanish, because Mapudungun-medium education teaches mathematics out of Spanish materials and has no settled term for a ray or a parabola. Where that boundary should fall is a language-planning judgement rather than a mistake, and both headers say so.

**Aymara marks how the speaker knows what they are saying**, on every assertion. The direct-knowledge «-wa» is written throughout, which is a deliberate reading of what these strings are: the software counted the attempts, graded the response and read the source, so direct knowledge is the honest marking and a reportative would be the wrong claim rather than a more cautious one. That decision is the same in all four files, so it is recorded once at the top rather than at fifty sites.

## Chemistry

**This is the first batch with no complete catalog in it**, and the split is worth keeping because only part of it is about the languages. Seven of the eight are the colonial-medium case in three school systems: Quechua, Aymara, Guarani, Nahuatl, Kʼicheʼ and Mapudungun meet the periodic table in Spanish and Haitian Creole in French. Bilingual education in all seven reaches real classrooms — Guarani furthest of any, being co-official in Paraguay — but it is the primary grades and the language classroom, and secondary chemistry is not, so in all seven the fallback *is* the curriculum. Ojibwe is the eighth and different: its schooling is immersion and revitalization and stops below those grades, and it is also where coining would be hardest rather than merely unreviewed, since a Nahuatl coinage is a name and an Ojibwe one is a whole descriptive verb — 118 of them would be 118 sentences invented by a machine.

## Scripts and tags

`oj` adds a script asymmetry of the kind `bs` and `pa` already have: Ojibwe is written both in the Latin orthography this catalog uses and in syllabics, so a reader under `oj-Cans` reaches it and gets Latin. `nah` is the first locale in the roster CLDR has **no name for in either language**, so `<document lang>`'s autocomplete offers "nah" and nothing else — the documented fallback for a tag ICU does not know, and nothing here hand-writes around it, the same rule that leaves `co` reading "Corsican" twice.

None of the eight is right-to-left, so `direction.ts` and `direction.test.ts` are untouched.

## What is deliberately not here

Kalaallisut, Inuktitut, Cree, Cherokee and Navajo were in the original plan and are not in this PR. For those five a seed would have been largely invented vocabulary — polysynthetic word-building I cannot verify, tone and diacritics at a high error rate, and for the three syllabary languages text a reader cannot check against the English beside it. That last is precisely the reason this README already gives for *omitting* Khmer, Lao and Sinhala chemistry rather than guessing it, and the same reasoning applies to a whole catalog. They want a corpus or a speaker, not another machine pass.

## Tests

- `packages/i18n/test/negotiate.test.ts`: both halves of the fold — the member CLDR handles and several it does not — across all five wider codes, `ciw` and the Nahuan codes among them, the individual-language three-letter codes that need no alias, the region and script tags including `oj-Cans`, and two negative controls: no fold is invented for `quc` or `arn`, which are individual languages and members of nothing, and `alq` is held outside `oj`.
- `packages/utils/test/styleDescriptions.test.ts`: Ojibwe animacy across four nouns, plus the case that would fail if `$gender` were ignored (the adjectives alone, noun withheld) and the two-animacies-in-one-sentence case where the fill agrees with the shape and the border with «jiigaatigwaan»; the six prenominal catalogs asserted as identities; Haitian Creole and Guarani reaching `[noun-tail]`; and all eight added to the two-positions suite on the identity side, since not one of them forks on `$role`.
- `packages/doenetml-worker-javascript/.../styleDescriptionLocale.test.ts`: Ojibwe and Haitian Creole through the whole worker path — the two ends of the batch's range, one agreeing two animacies in a sentence and the other inflecting nothing at all.

## Docs

`packages/i18n/README.md` gains the eight codes in its roster, the chemistry paragraphs for this batch, and a naming/script section covering the macrolanguage bug, the written standard each catalog picked, `oj-Cans` and `nah`'s missing CLDR name.

## Regenerated

`supportedLocales.ts` (132 locales) and `doenet-schema.json`, via codegen → build i18n → build:schema → build static-assets.


dqnykamp:
Follow-up for the five held back from this batch: #1655 (Kalaallisut, Inuktitut, Cree, Cherokee, Navajo).

Issue #1684, opened 2026-08-10
Removed pegboard's pegs reappear when the graph's bounding box changes


dqnykamp:
`pegboard.tsx` registers a `boundingbox` listener on the JSXGraph board but never removes it. As a result, a pegboard that has been removed from the document leaves a live listener behind, and that listener recreates the pegs the next time the graph's bounding box changes. The recreated pegs belong to no mounted component, so nothing can ever remove them.

## To reproduce

```doenet
<booleanInput name="show" prefill="true" />
<graph name="g" xmin="-5" xmax="5" ymin="-5" ymax="5">
  <conditionalContent condition="$show"><pegboard/></conditionalContent>
</graph>
<p>Change xmax: <mathInput name="xmaxInput" bindValueTo="$g.xmax" /></p>
```

1. The graph shows a 9×9 grid of pegs.
2. Uncheck the boolean input. The pegs disappear — correct.
3. Change `xmax` from 5 to 9 (or pan/zoom the graph by any other means).
4. **The pegs come back**, laid out for the new bounding box (13×9), even though the `<pegboard>` is no longer in the document. Unchecking/rechecking the boolean input does not clear them; only reloading does.

Confirmed in headless Chrome against a local build of `main` at `6c812b598`, by counting `elType === "point"` objects on the board: 81 initially, 0 after removal, **117 after the bounding-box change**.

## Cause

In `packages/doenetml/src/Viewer/renderers/pegboard.tsx`, `createPegboardJXG()` ends with

```ts
board.on("boundingbox", () => {
    // ...recompute lattice indices...
    if (/* indices changed */) {
        recalculatePegboard(minXind, maxXind, minYind, maxYind);
    }
});
```

but the unmount cleanup only removes the points:

```ts
useEffect(() => {
    //On unmount
    return () => {
        deletePegboardJXG();   // removes points, sets pegboardJXG.current = null
    };
}, []);
```

There is no matching `board.off("boundingbox", listener)`. So after unmount the listener is still attached to a board that is still alive, and on the next bounding-box change it calls `recalculatePegboard`, which opens with

```ts
if (pegboardJXG.current === null) {
    return createPegboardJXG();
}
```

`pegboardJXG.current` is `null` (cleanup set it), so the pegs are rebuilt from scratch — and `createPegboardJXG()` registers *yet another* listener on the way out.

Two related consequences of the same root cause:

[...]

PR #1693, opened 2026-08-11
refactor: take a field's function as a child, and give it a variables attribute

dqnykamp:
Follow-up to #1682, acting on review feedback. `<slopeField>` and `<vectorField>` are unpublished, so nothing here keeps a shim for the old spelling.

```doenet
<mathInput name="v1">q</mathInput>
<mathInput name="v2">r</mathInput>
<graph>
  <vectorField variables="$v1 $v2">(r, -q)</vectorField>
</graph>
```

## The `function` attribute is gone; the function is a child

A field's function is written inside the component, either as a bare expression or as a `<function>` child — which may be a reference to one declared elsewhere. The state variables read the `functions` child group.

The two spellings were doing the same work. The bare expression was already being turned into a `<function>`, and `function="..."` created one too, so an author picked between forms that could not differ — and when both were given, one silently won. That warning (`doenet-w0123`) is retired along with the attribute.

## New `variables` attribute

`variables` names the inputs of a bare expression, in order, and defaults to `x y`, so `<slopeField variables="s t">s - t</slopeField>` reads an equation in the letters its author wrote it in without reaching for an explicit `<function variables="s t">`.

It is the same `_variableNameList` a `<function>` takes, and it is **moved onto the wrapping `<function>` rather than read**, so the names may themselves be references — the snippet above lets a student say which axis `q` names and which `r` does, and the field is redrawn as they type. The expression there is written in fixed letters on purpose: `($v2, -$v1)` would have renamed the inputs and the expression together, and drawn the same field whatever the student typed.

An explicit `<function>` child names its own inputs and is used exactly as written, so `variables` has nothing to say about it. Writing both **warns** (`doenet-w0124`), as does writing `variables` with no expression at all; the attribute otherwise looks exactly like the one a `<function>` does obey while doing nothing.

## The sugar moved to the DAST normalizer

`packages/parser/src/dast-normalize/component-sugar/field.ts`, joining `fractionInput`, `answer` and the rest.

Doing it in the worker meant working against components that had already been built: `variables` was a serialized `_variableNameList` whose references had deliberately not resolved, so it had to be *moved* onto the wrapper rather than copied, or its components would exist twice under one index. That cost two additions to shared machinery — `allComponentAttributes`/`removeAttributes` in `applySugar`, and `skipIfAllWrappingComponentType` plus wrapping attributes in the child branch of `returnWrapNonLabelsDescriptionsSugarFunction`.

In the parser the same rewrite is a tree edit. Nothing has an index yet, so moving an attribute is `delete` and assign, exactly as `descriptionAttributeSugar` already does. Both shared-machinery changes are reverted, and both files are net deletions: `sugar.ts` loses `componentAttributeNames`, added there in #1682 for the now-retired warning, and `label.ts` loses `createWrappingComponentAttributes`, added there in #1682 for this one caller — the field's sugar was the only thing that ever passed it, and the wrapping component's attributes are empty for everyone else.

## Fields no longer take a `<label>`, or anything that styles one

`GraphicalComponent` hands every subclass the `labels` child group, the `labelIsName`, `applyStyleToLabel` and `maskLabel` attributes and the `label` state variables unconditionally. A field covers the whole visible region, so its renderer passes `withLabel: false` and draws none however one is written — the `<label>` was accepted, read into a state variable, and dropped, and the two styling attributes were public, sent to the renderer, and offered in the editor's completions with nothing to act on.

`static includeLabels` drops the surface whole. The group and the `label` state variable cannot be separated in any case, since `label` depends on the group. The flag is tested against `false` rather than for truthiness because these methods are borrowed outside the hierarchy — `Function` calls `GraphicalComponent.returnStateVariableDefinitions.call(this)`, where the flag is absent and labels are wanted.

On the renderer side, `GraphicalSVs` splits into the half every graphical component has — layer, visibility, style — and the label half only a component that can carry one receives. `FieldSVs` extends the first, naming what a field is sent rather than subtracting what it is not; a subtraction would have had to be revisited each time a label state variable was added, and would have silently claimed the new one until it was.

`<pegboard>` is in the same position and is left alone here; a follow-up PR covers it.

## Fixed: a one-input function's domain

A field always has both an `x` and a `y` to hand its function, but the function need not take two inputs — a `<function>` child naming a single variable is now the ordinary way to write a field of `f(x)`, and a referenced one-input function reached the same path before this PR.

`createFunctionFromDefinition` gives a one-input function the signature `(x, overrideDomain)`. Passing the lattice's `y` positionally therefore did not supply a second input; it suppressed the domain check, and a field of a domain-restricted `f(x)` was drawn right across the interval its author had excluded — everywhere except the row `y = 0`. The rehydration now consults `numInputs` first. It also moves from `useFieldCurve.ts` to `fieldGeometry.ts`, which is pure and directly unit-testable; every definition that omits `numInputs` (interpolated, bezier, piecewise, a function operator) describes one input, so the absent case takes the one-input path.

## Notes for review

- **Diagnostic codes are append-only.** #1682's removal of `doenet-w0123` deleted it from the lock, which `lint:i18n` rejects. It is now retired in place — kept in `DIAGNOSTIC_CODES` with its catalog entry and listed in `RETIRED_DIAGNOSTIC_CODES`, which was empty until now — and the new warning takes `w0124` rather than reusing the freed number.
- The schema diff is confined to the two elements: `labelIsName`, `applyStyleToLabel`, `maskLabel` and the three label children come off, `function` becomes `variables`, and the five function child types move to the head of the child list as the new `functions` group. Nothing else changes. Verified by diffing the parsed schemas rather than the JSON text.
- The sugar finds `variables` however it was capitalized. Element names are case-sensitive — `<slopefield>` is an invalid component type, so the names this sugar matches on can be written out as they stand — but attribute names are not, so `VARIABLES="s t"` is the same attribute a component would be built from. Left behind, it would name nothing while the wrapper took the default `x y`, and the expression the author did write would be NaN at every lattice point: a blank field reported as having a function. `deprecations.ts` looks its own attributes up the same way.
- `<slopeField>$F</slopeField>` is still wrapped. At DAST level a macro naming a `<function>` is indistinguishable from one naming a `<math>` — neither has resolved. That is right either way: a `<function>` holding a referenced function passes its inputs through positionally, so the referenced function keeps its own variable names.

## Testing

- `packages/parser` — the full suite. `normalize-dast.test.ts` gains 9 tests asserting exact XML round-trips: the wrap, `variables` moved with `$v1 $v2` intact, `VARIABLES` moved despite its capitalization, an explicit `<function>` child left alone, whitespace-only, the two warning cases with their `reason` argument, and two on where the wrapper lands — a `<label>` written after the expression, and one written before it across line breaks, each staying on the side it was written on. `coded-dast-errors.test.ts` gains a case for `doenet-w0124` in both its reasons; that file asserts every code the parser names is exercised, which is a guard scoping a run to one spec walks straight past.
- `packages/doenetml-worker-javascript/src/test/tagSpecific/slopeField.test.ts` — 15 tests. The dynamic-variables one is written to discriminate: the expression is fixed at `q - r` while `variables="$v1 $v2"` comes from the inputs, so swapping the inputs flips `f(3,5)` from `-2` to `2`. Writing the expression as `$v1 - $v2` would have been symmetric and passed vacuously. Two more cover what only the worker can show: that a one-input `<function>` child keeps its `domain`, and that the `no-expression` warning arrives as a diagnostic even when it is the only child the component has.
- `packages/doenetml/src/Viewer/renderers/utils/fieldGeometry.test.ts` — 4 tests on the rehydration, including one that calls a domain-restricted `f(x)` at five lattice rows; passing `y` positionally makes four of them succeed where they should be NaN.
- `packages/doenetml-worker-javascript` — the full suite (3514 tests), as a regression net around the `GraphicalComponent` change, which every graphical component inherits. It earned its keep: the first version of the flag stripped `<function>` of its label and broke a point-constraint test.
- `packages/test-cypress/cypress/e2e/tagSpecific/slopefield.cy.js` — 3 tests, run against a full build. The third exercises `variables` end to end, which is the path that depends on the attribute surviving the move.
- Full suites for `@doenet/i18n`, `@doenet/doenetml` and `@doenet/static-assets`; `lint:i18n`, `check:docs-coverage`, prettier and `tsc` pass; schema regenerated.

Reference pages and the changeset are updated throughout; two things on those pages are worth naming. The `variables` example that lets a student name the inputs draws a fixed expression, for the reason just given, so swapping the two `<mathInput>`s visibly flips the field. And the fine-lattice example on each page now sets `lineWidth="2"`, with a sentence saying why: a field is drawn in its component's selected style, whose line is 4 pixels wide by default, and a lattice of hundreds of marks at that weight is heavier than an author expects. The remaining examples keep the default, so the pages still show what an author gets without asking for anything.

#1682's own changeset is corrected too: it is still pending release, so it would otherwise have documented the `function` attribute in the same release notes that announce its removal, for an attribute no release ever had.


dqnykamp:
Follow-up #1694 is stacked on this branch: it removes the `<label>` from `<pegboard>` using the `includeLabels` flag added here, and fixes #1684. Merge this one first.

PR #1704, opened 2026-08-13
Publish vscode extension once a day

siefkenj:
This PR modifies the vscode extension publish script so that it only runs once a day and only pushes a new one if the build changes.


siefkenj:
@dqnykamp This is ready for review

dqnykamp:
I had claude review this PR. Is the following review useful?

.github/workflows/vscode-prerelease.yml
```diff
@@ -0,0 +1,329 @@
...
+                      echo "Nothing was built or published." >> "$GITHUB_STEP_SUMMARY"
+                      echo "proceed=false" >> "$GITHUB_OUTPUT"
+                      exit 0
+                  fi
+
+                  # Never publish code CI has not blessed.  A pending or failed
+                  # run is a skip rather than a failure: the next scheduled run
+                  # picks the commit up once CI is green, and a red cron job
+                  # every morning would be noise, not signal.
+                  if ! TARGET_SHA="${HEAD_SHA}" node .github/scripts/verify-ci.mjs; then
```
dqnykamp, line 128: **A commit skipped for non-green CI is skipped permanently**, which contradicts the comment just above.

The comment says the next scheduled run picks the commit up once CI is green, but the baseline is derived from wall-clock time, not from "last published."

Scenario: `C1` lands on main day 1 at 07:00. The 07:17 run finds CI still in progress, `verify-ci.mjs` fails, the step writes `proceed=false`. No further commits land. On day 2 at 07:17, `--before="24 hours ago"` is day 1 07:17, so `BASE_SHA` resolves to `C1` (committed 07:00, before the cutoff) — which is also `HEAD_SHA`. The run reports "No commits in the last 24 hours — nothing to do" and `C1` is never published, even though CI went green five minutes after the day-1 run.

The same permanent loss happens if the run itself fails (build error at the current commit) or is cancelled.

.github/workflows/vscode-prerelease.yml
```diff
@@ -0,0 +1,329 @@
...
+                      ''|*[!0-9]*)
+                          echo "Error: window_hours must be a positive integer, got '${WINDOW_HOURS}'" >&2
+                          exit 1
+                          ;;
+                  esac
+
+                  HEAD_SHA=$(git rev-parse HEAD)
+                  # The newest commit on main that is already older than the
+                  # window — i.e. what main looked like ${WINDOW_HOURS} ago.
+                  BASE_SHA=$(git rev-list -n 1 --first-parent --before="${WINDOW_HOURS} hours ago" HEAD || true)
```
dqnykamp, line 109: **Cron jitter lets commits fall into the baseline without ever having been the "current" build.**

This anchors the window to *this* run's start time, but the previous run's coverage ended at *its own* start time. GitHub scheduled runs routinely drift by tens of minutes, sometimes over an hour.

Scenario: day 1's run starts 07:17 with `HEAD = C0`; `C1` lands day 1 at 07:30; day 2's run is delayed to 08:30, so the cutoff is day 1 08:30 and `BASE_SHA = C1`. `C1` was never a "current" build and is now the baseline — if later commits don't further change the extension, `C1`'s extension change is never published.

Anchoring the baseline on the last commit this workflow actually evaluated/published (a tag, a cached artifact, or the previously published version's SHA) rather than on the clock would close both this and the CI-skip case above.

.github/scripts/hash-vscode-extension.mjs
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Hash one file's contents, with package.json versions blanked out.
+ *
+ * @param {string} file
+ * @returns {string}
+ */
+function hashFile(file) {
+    let contents = fs.readFileSync(file);
+    if (path.basename(file) === "package.json") {
```
dqnykamp, line 90: **Blanking `version` here does not stop version bumps from forcing a publish.**

The header claims a version bump alone is not a reason to publish, but `packages/doenetml/vite.config.ts:55` bakes `DOENETML_VERSION: JSON.stringify(version)` (from `packages/doenetml/package.json`) into the bundle, and `build:preview-window` depends on `../doenetml:build`.

I confirmed the literal `0.7.24` is present in `packages/vscode-extension/extension/build/preview-window/assets/EditorViewer-BZ8KbrSc.js` in the current build. So merging the changesets "Version Packages" PR — which changes nothing but version numbers — flips the fingerprint and triggers a Marketplace + Open VSX pre-release. That is exactly the spurious release this PR exists to prevent.

.github/scripts/diff-extension-manifests.mjs
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Collapse a content hash in a built filename, so that the same chunk before
+ * and after a rebuild lines up: `codeEditor-ObpSVeQN.js` -> `codeEditor-*.js`.
+ *
+ * @param {string} filePath
+ * @returns {string}
+ */
+function collapseContentHash(filePath) {
+    return filePath.replace(/-[A-Za-z0-9_-]{8,}(\.[^./]+)$/, "-*$1");
```
dqnykamp, line 41: **This regex is greedy enough to merge genuinely distinct chunks.**

`[A-Za-z0-9_-]{8,}` includes `-`, and the match starts at the *first* hyphen, so everything from that hyphen to the extension is swallowed.

In the real build tree, `content-Ca3obMaT-Ca3obMaT.js`, `content-GsVI9JDC-GsVI9JDC.js`, `content-CrLFa4AO-CrLFa4AO.js`, and dozens of other distinct chunks all collapse to the single key `assets/content-*.js` — as do all the `editor-*`, `chrome-*`, and `diagnostics-*` chunks. A rebuild touching 50 content chunks reports "1 modified", not the "97 files modified" the header promises.

Worse, a *newly added* chunk whose prefix already exists only changes the Set size, so it is reported under `modified`, never `added`.

It also collapses names carrying no hash at all: `open-sans-v18-latin-regular.woff` becomes `open-*.woff`, and `config/language-configuration.json` becomes `config/language-*.json`.

Anchoring on the *last* hyphen and disallowing `-` inside the hash class would fix it.

.github/workflows/vscode-prerelease.yml
```diff
@@ -0,0 +1,329 @@
...
+                  else
+                      # The decision above is a plain byte comparison, so any
+                      # difference at all publishes.  The summary below only
+                      # formats that difference for a reader: Vite gives chunks
+                      # content-hashed names, so a one-line source fix renames
+                      # most of the import graph and a raw diff is ~200 lines of
+                      # near-identical renames.
+                      echo "The built extension differs from the baseline:"
+                      diff -u "$BASE" "$CURRENT" || true
+                      node "$RUNNER_TEMP/diff-extension-manifests.mjs" \
```
dqnykamp, line 252: **The "cannot cause a change to be missed" invariant only holds if the formatter's exit status is ignored.**

This step runs under `set -euo pipefail` and this line has no `|| true`, unlike the `diff -u` immediately above it. If the formatter throws — malformed manifest line, or ENOSPC/EPIPE writing to the 1 MiB-capped step summary — the `compare` step fails, `steps.compare.outputs.publish` is never written, and the workflow aborts *before* the publish steps.

So a real extension change goes unpublished, which directly contradicts the script header's claim that nothing here can cause a change to be missed.

.github/workflows/vscode-prerelease.yml
```diff
@@ -0,0 +1,329 @@
...
+              id: decide
+              env:
+                  GITHUB_TOKEN: ${{ github.token }}
+                  REPO: ${{ github.repository }}
+                  WINDOW_HOURS: ${{ inputs.window_hours || '24' }}
+              run: |
+                  set -euo pipefail
+
+                  case "${WINDOW_HOURS}" in
+                      ''|*[!0-9]*)
```
dqnykamp, line 100: **`window_hours` validation accepts `0` despite the error message saying "positive integer".**

`''|*[!0-9]*)` rejects empty and non-digit input, but `0` is all digits. A manual dispatch with `window_hours: 0` yields `--before="0 hours ago"`, which returns `HEAD` itself, so `BASE_SHA == HEAD_SHA` and the run silently reports "No commits in the last 0 hours — nothing to do" instead of the intended validation error.

Adding a `0)` arm to the `case` would catch it.

.github/workflows/vscode-prerelease.yml
```diff
@@ -0,0 +1,329 @@
...
+                      echo "::warning::VSCE_PAT repository secret is not set — the VS Code extension pre-release was not published."
+                      echo "Add VSCE_PAT as a repository-level secret (Settings → Secrets → Actions) to enable pre-release publishing."
+                      exit 0
+                  fi
+                  npm run publish:prerelease -w @doenet/vscode-extension
+              env:
+                  VSCE_PAT: ${{ secrets.VSCE_PAT }}
+
+            - name: Publish the Open VSX pre-release
+              if: steps.compare.outputs.publish == 'true'
```
dqnykamp, line 320: **"Each registry publishes independently" overstates what this gate does.**

The header comment says a missing token for one registry leaves the other alone. That holds for the missing-token case, which is handled in-shell — but any *other* `vsce publish` failure (expired PAT, Marketplace 5xx, rate limit, or `version already exists` on a re-run) fails the previous step, and this one is gated only on `steps.compare.outputs.publish == 'true'`, so it is skipped. Open VSX then silently falls behind the Marketplace.

`if: always() && steps.compare.outputs.publish == 'true'` would match the stated intent.

.github/scripts/hash-vscode-extension.mjs
```diff
@@ -0,0 +1,122 @@
...
+const manifest = lines.join("\n") + "\n";
+fs.writeFileSync(manifestOut, manifest);
+
+const aggregate = createHash("sha256").update(manifest).digest("hex");
+
+console.log(`Hashed ${lines.length} files under ${extensionDir}`);
+console.log(`Manifest: ${manifestOut}`);
+console.log(`Aggregate hash: ${aggregate}`);
+
+if (process.env.GITHUB_OUTPUT) {
```
dqnykamp, line 120: **The documented `hash` step output is unreachable.**

The header advertises that the script sets a `hash` step output when running under GitHub Actions, but neither `Fingerprint the baseline build` (workflow line 163) nor `Fingerprint the current build` (line 200) declares an `id:`, so no expression can ever read `steps.*.outputs.hash`.

Either give those steps ids or drop this block along with the doc line.

dqnykamp (commented):
Review of the once-a-day pre-release workflow. Nothing here blocks a merge on its own, but two findings go to whether the workflow does what its comments promise — a commit can be skipped permanently, and a version-only bump still forces a publish.

**Checked and believe fine:** `steps.baseline-build.outcome == 'success'` correctly distinguishes a real success from a `continue-on-error` failure and from `skipped`; `git clean -xfdq -e node_modules -e target` preserves `packages/doenetml-worker-rust/target` while wiping `extension/build`; `$RUNNER_TEMP` is outside the workspace so the stashed helper scripts survive the clean; `inputs.dry_run`/`inputs.force` evaluate to `''` on `schedule`, so scheduled runs publish and manual runs default to dry (and `set -u` is satisfied since both are declared in `env:`); the version bump happens after both fingerprints and `check-version-consistency.ts` only compares the two package.json files the workflow writes, so `vscode:prepublish` passes; the `node_modules` hash exclusion is harmless because both publish scripts use `--no-dependencies`.

PR #1705, opened 2026-08-13
Graphs: Improve grid visual hierarchy, contrast accessibility, and grid synchronization across rendering modes

DoctorHaitch:
# Graphs: Improve grid visual hierarchy, contrast accessibility, and grid synchronization across rendering modes

## Summary

This PR improves the visual hierarchy, accessibility, and rendering consistency of graph grids across all supported grid modes.

`grid="medium"` and `grid="dense"` derive their spacing from the axis major-tick interval, while defined-step grids such as `grid="1 1"` continue to use their authored x/y spacing. Grid rendering has also been consolidated through a single rendering and synchronization path, ensuring that grid presentation remains accurate when switching grid modes or when effective axis tick spacing changes.

A primary goal of this change is to reduce visual clutter and prevent structural graph elements from competing with meaningful authored content. Grid lines now use dedicated theme tokens with lower visual emphasis than axes, tick marks, axis labels, and authored labels or annotations. This creates a clearer distinction between background reference geometry and information-bearing content, improving graph readability while maintaining WCAG 2.1 AA accessibility targets. Additionally, axis labels and lines use a higher contrast than grids, but lower contrast than default styling for labels and other user-authored content or navigation controls.

Additionally, this change introduces Cypress regression coverage that validates browser-resolved rendered colors and contrast ratios across medium, dense, and defined-step grids in both light and dark themes.

The results of this change can be previewed here:

lightMode darkMode

---

#### Global Changes

- Removed "opacity" fixes that could cause inconsistent contrast computations or rendering.
- Added fixed color tokens with desired contrast ratios.
- Changed universal use of `--canvasText` token to targeted use of new tokens.

#### Medium Grids (`grid="medium"`)

- Grid lines now align with the graph's major tick spacing.
- Grid spacing updates automatically when effective axis tick intervals change.
- Improves consistency between axis structure and grid presentation.

#### Dense Grids (`grid="dense"`)

- Major grid lines align with major tick spacing.
- Minor grid lines are rendered between major intervals. Full lines were replaced with a dotted grid.
- Minor axis tick marks remain visible when ticks are enabled.
- Provides additional reference guidance while preserving visual hierarchy.

#### Defined-Step Grids (`grid="1 1"`, etc.)

- Continue to use explicitly authored x/y spacing.
- No behavioral changes to authored grid density.
- Benefit from the same rendering and synchronization infrastructure as other grid modes.

---

## Rendering Architecture Improvements

Grid generation and updates now flow through a single rendering path.

### Benefits

- Consistent behavior across all grid modes.
- Elimination of mode-specific synchronization differences.
- Reliable updates when:
  - Switching between grid modes.
  - Axis scales change.
  - Tick spacing changes.
  - Theme changes occur.

### Result

The rendered grid always reflects the current graph configuration and effective axis spacing, regardless of grid type.

---

## Visual Hierarchy Improvements

The primary objective of these changes is to establish a clearer visual hierarchy within graphs.

Grid lines are structural reference elements and should remain visually subordinate to higher-value content, including:

- Plotted data
- Authored labels
- Annotations
- Other author-defined graph elements

Previous grid styling could cause structural components to compete visually with meaningful authored content. By adjusting spacing behavior and introducing dedicated contrast-calibrated tokens, the graph presents less visual noise while retaining clear orientation cues.

### Expected Outcomes

- Reduced visual clutter.
- Improved distinction between structure and content.
- Less confusion between grid geometry and authored labels.
- Better emphasis on meaningful graph information.
- More consistent reading experience across light and dark themes.

---

### Grid Lines

Grid lines now use dedicated light- and dark-theme color tokens rendered at full opacity.

Goals:

- Approximately **3:1 contrast** against the graph canvas.
- Sufficient visibility for orientation and reference.
- Lower emphasis than axes and authored content.

### Axes, Tick Marks, and Axis Labels

Axes, tick marks, and axis labels now use separate color tokens.

Goals:

- Slightly greater than **4.5:1 contrast** against the graph canvas.
- Stronger visual prominence than grid lines.
- Improved readability and accessibility.

### Dense Grid Tick Behavior

When ticks are enabled:

- Dense grids continue to display minor axis tick marks.
- Minor ticks remain visually associated with dense-grid subdivisions.
- Structural guidance is preserved without elevating grid prominence.

### Accessibility Compliance

The resulting contrast relationships fall within the expectations of **WCAG 2.1 AA** while maintaining a clear visual hierarchy between supporting structure and meaningful graph content.

---

### Cypress Regression Coverage Added

New Cypress regression coverage validates the graph visual hierarchy across:

#### Grid Modes

- `grid="medium"`
- `grid="dense"`
- Defined-step grids such as `grid="1 1"`

#### Themes

- Light theme
- Dark theme

### What Is Verified

The tests intentionally validate the browser's rendered output rather than implementation details.

For each graph configuration and theme, the suite:

- Creates representative medium, dense, and defined-step graphs with axes, labels, and plotted content.
- Reads graph styling from the browser's computed styles.
- Resolves the active canvas, grid, and axis colors from the actual rendered theme variables.
- Locates rendered SVG grid and axis elements based on their computed colors rather than serialized SVG attributes.
- Verifies that grid lines, axes, and axis labels resolve to the expected rendered colors.
- Confirms that axis labels inherit the same visual treatment as axes.
- Computes effective grid contrast using the rendered grid color, applied opacity, and graph canvas color.
- Computes axis and label contrast against the graph canvas.
- Runs the same assertions in both light and dark themes.

### Contrast Validation

The regression tests explicitly verify the visual hierarchy targets introduced by this change:

- Grid lines target approximately **3:1 contrast** against the graph canvas, consistent with WCAG's non-text contrast guidance.
- Axes and axis labels exceed **4.5:1 contrast** against the graph canvas.
- Grid lines remain visible while maintaining lower visual prominence than axes and labels.
- The distinction between structural graph elements and meaningful authored content is preserved across themes.

### Why This Approach Matters

The tests intentionally use the browser's resolved styles because JSXGraph may serialize SVG styling as either literal `var(...)` references or fully resolved colors. By validating computed styles and calculating contrast from the rendered result, the coverage verifies the user-visible experience rather than implementation-specific details.

This provides regression protection for:

- Theme token changes.
- CSS variable resolution behavior.
- Grid opacity changes.
- Accessibility contrast requirements.
- Graph visual hierarchy expectations.

### User Impact

These tests ensure that future styling updates cannot unintentionally:

- Increase grid prominence to the point where it competes with axes, labels, or authored content.
- Reduce grid visibility below the intended accessibility threshold.
- Introduce inconsistent rendering between light and dark themes.
- Break the intended distinction between structural graph elements and meaningful graph information.

---

### Before

- Grid spacing behavior differed across grid modes.
- Grid updates could become inconsistent when axis tick spacing changed.
- Structural elements could compete visually with authored content.
- The distinction between grid geometry and labels was less clear.
- Accessibility expectations were not comprehensively validated through regression testing.

### After

- Medium and dense grids consistently follow axis major tick spacing.
- Dense grids provide additional reference guidance through minor grid lines.
- Defined-step grids retain authored spacing behavior.
- Grid rendering remains synchronized through a unified rendering path.
- Visual clutter is reduced.
- Structural graph elements are more clearly differentiated from authored labels and annotations.
- Axes and labels maintain stronger visual prominence than grids.
- Automated regression coverage protects contrast, accessibility, and rendering behavior across themes.


dqnykamp:
I asked claude to review this PR. I think it will pile on next. I hope it is helpful!

dqnykamp (commented):
## Automated code review

Reviewed the diff at `5f4c1dd` against the JSXGraph 1.12.2 sources in `node_modules` to confirm the runtime semantics. 11 findings, inline below.

The headline issue is that the grid change-detection in `useGridAndAxesSync.ts` doesn't actually work: the `majorStep` array is handed to JSXGraph *and* kept as the memo, and JSXGraph mutates it in place. The effect has no dependency array, so the grid is destroyed and rebuilt on every render — including every frame of a point drag. The sibling memo on `JXG.Options.grid.gridX/gridY` has the same effect for a different reason: it's a JSXGraph global, so two graphs with different grid modes on one page invalidate each other. Since `board.grids` is never spliced on removal, both of those leak two curve entries per rebuild.

Two findings undercut the PR's own goals: the grid's `majorStep` is snapshotted before `setMinorTicks` has run on the first board-enabled pass (the desync this PR is fixing), and dropping `highlightStrokeOpacity: 1` from the axis tick-label options restores JSXGraph's 0.667 dimming default — hovering a tick label now fades it below the 4.5:1 contrast target, on top of the color having moved to the lower-contrast `--graphAxes` token.

The new Cypress spec's dark-mode block passes identically in light mode — it reads the token from the same element it asserts against, so a total dark-theme regression wouldn't fail it.

packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts
```diff
@@ -34,26 +34,86 @@ export default function useGridAndAxesSync({
...
+        if (
+            Array.isArray(SVs.grid) ||
+            SVs.grid === "dense" ||
+            SVs.grid === "medium"
+        ) {
+            const fixedGrid = Array.isArray(SVs.grid);
+            const gridX = fixedGrid ? SVs.grid[0] : null;
+            const gridY = fixedGrid ? SVs.grid[1] : null;
+            const gridMode = fixedGrid ? "fixed" : SVs.grid;
+            const majorStep = fixedGrid
```
dqnykamp, line 46: **HIGH** — the `majorStep` array is aliased into JSXGraph's `visProp` and mutated in place, so this memo self-invalidates on every render.

The same array object is passed as the `majorStep` attribute *and* stored as `grid.doenetGridMajorStep` (line 98). JSXGraph stores arrays by reference (`Type.mergeAttr` flat-copies arrays, `type.js:1354`; `GeometryElement._set` skips arrays because `Type.isObject` excludes them, `element.js:1130`), and `grid.js:493-497` mutates it:

```js
if (majorStep[0] === 'auto') { … majorStep[0] = delta; }
```

The board is created with `axis: false`, so `board.defaultAxes` is never set and `parentAxes` stays empty — the `'auto'` branch is guaranteed.

Scenario: `<graph grid="medium" displayYAxis="false" />`. `yaxisRef.current` is null, so `majorStep[1] === "auto"`; the first board update rewrites it to a number; the next effect run computes `"auto"` again and `doenetGridMajorStep[1] !== majorStep[1]` → `gridParamsChanged` → the grid is destroyed and recreated. The effect has no dependency array, so this repeats on **every render**, including every frame of a point drag.

dqnykamp, line 46: **MEDIUM** — `majorStep` is snapshotted before `setMinorTicks` has run.

`getDistanceMajorTicks()` reads `minorticks` (`ticks.js:829`: `minDist *= (ev_mt + 1)`), and `setMinorTicks` writes it. `useAxisTickSpacingSync` is declared *after* `useGridAndAxesSync` and is still gated on `!boardJustInitializedRef.current`, while `useGridAndAxesSync` was changed here to `Boolean(board)`.

So on the first board-enabled pass the grid is created from a `minorticks` value `setMinorTicks` (`jsxgraph.ts:100-146`) has never touched. If no further render of the graph occurs, the grid spacing stays permanently different from the axis major ticks — the exact desync this PR sets out to fix.

Passing `[xaxisRef.current, yaxisRef.current]` as the grid's parents with `majorStep: 'auto'` would let JSXGraph track this itself.

packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts
```diff
@@ -34,26 +34,86 @@ export default function useGridAndAxesSync({
...
+                      xaxisRef.current?.defaultTicks.getDistanceMajorTicks() ??
+                          "auto",
+                      yaxisRef.current?.defaultTicks.getDistanceMajorTicks() ??
+                          "auto",
+                  ];
+            const currentGrid = (board as any).doenetGrid;
             const gridParamsChanged =
-                JXG.Options.grid.gridX !== SVs.grid[0] ||
-                JXG.Options.grid.gridY !== SVs.grid[1];
+                JXG.Options.grid.gridX !== gridX ||
```
dqnykamp, line 56: **MEDIUM-HIGH** — `JXG.Options.grid.gridX/gridY` is a JSXGraph **global** being used as a per-board memo.

Two graphs on the same page with different grid modes clobber each other's value. With `<graph grid="1 1" />` and `<graph grid="medium" />`, graph A writes `1` and graph B writes `null` on alternating effect runs, so `gridParamsChanged` is true for both on every render → both grids are torn down and rebuilt on every render (again, every frame during a drag).

Before this PR the `medium`/`dense` path never touched `JXG.Options.grid`, so this combination wasn't affected. The memo belongs on the grid object, next to `doenetGridMode`/`doenetGridMajorStep`.

packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts
```diff
@@ -34,26 +34,86 @@ export default function useGridAndAxesSync({
...
             if (gridParamsChanged) {
-                JXG.Options.grid.gridX = SVs.grid[0];
-                JXG.Options.grid.gridY = SVs.grid[1];
-                if (board.grids.length > 0) {
-                    board.removeObject(board.grids[0]);
+                JXG.Options.grid.gridX = gridX;
+                JXG.Options.grid.gridY = gridY;
+                if (currentGrid) {
+                    board.removeObject(
+                        [currentGrid, currentGrid.minorGrid].filter(Boolean),
```
dqnykamp, line 66: **MEDIUM** — `board.grids` is never cleaned when the doenet grid is removed (same at line 111).

`grid.js:1039-1040` pushes both the major and the minor curve into `board.grids`, and `board.removeObject` never splices them out. Every recreation appends two more stale entries; combined with the two memo bugs above this grows without bound for the lifetime of the board.

It also breaks the fallback at lines 70 and 116: `board.grids.slice(0, 2)` removes the two *oldest* (already-removed) entries and then `board.grids = []` discards the rest, so `board.removeGrids()` and any future `board.grids` consumer see a wrong list.

packages/doenetml/src/Viewer/renderers/utils/useGridAndAxesSync.ts
```diff
@@ -34,26 +34,86 @@ export default function useGridAndAxesSync({
...
-                board.create("grid", [], {
-                    gridX: SVs.grid[0],
-                    gridY: SVs.grid[1],
+            if (!(board as any).doenetGrid) {
+                const isDenseGrid = SVs.grid === "dense";
+                const grid = board.create("grid", [], {
+                    gridX,
+                    gridY,
+                    majorStep,
+                    minorElements: isDenseGrid ? 4 : 0,
```
dqnykamp, line 80: **MEDIUM** — `minorElements: isDenseGrid ? 4 : 0` hardcodes a subdivision count that `setMinorTicks` may contradict.

`setMinorTicks` (`jsxgraph.ts:100-146`) chooses 3 or 4 minor ticks depending on the major-tick mantissa. For a 2·10^k major interval it picks 3, so the axis draws minor tick marks every `majorStep/4` while the dense minor grid draws dotted lines every `majorStep/5`.

Scenario: `<graph grid="dense" xMin="-10" xMax="10" />` at a width where the major interval settles on 2 — the dotted minor grid doesn't line up with the minor tick marks, contradicting the PR's stated goal that "minor ticks remain visually associated with dense-grid subdivisions".

packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
```diff
@@ -947,7 +946,6 @@ export function syncLineStrokeStyle(
     }
     if (vp.strokeopacity !== lineOpacity) {
```
dqnykamp, line 947: **MEDIUM** — `syncLineStrokeStyle` no longer keeps `highlightstrokeopacity` in sync, but creation still sets it to half.

`buildLineLikeAttributes` (`buildGraphicalAttributes.ts:89`) creates line-family objects with `highlightStrokeOpacity: SVs.selectedStyle.lineOpacity * 0.5`. With this assignment removed, a later `lineOpacity` change leaves the highlight value stale.

Scenario: a `<line>` created under a style with `lineOpacity` 0.3 (highlight 0.15) whose `styleNumber` is then changed to a style with `lineOpacity` 1 — the line renders at full opacity but hovering it drops it to 0.15, i.e. it nearly vanishes on hover.

This change also looks unrelated to the PR's grid/contrast purpose.

packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
```diff
@@ -244,20 +249,18 @@ export function createYAxis({
...
     yaxisOptions.ticks = {
         ticksDistance: 2,
         label: {
             offset: [12, -2],
             layer: 2,
-            strokeColor: "var(--canvasText)",
-            highlightStrokeColor: "var(--canvasText)",
-            highlightStrokeOpacity: 1,
+            strokeColor: "var(--graphAxes)",
+            highlightStrokeColor: "var(--graphAxes)",
```
dqnykamp, line 261: **LOW-MEDIUM** — dropping `highlightStrokeOpacity: 1` from the axis tick-label options (same at line 350) restores JSXGraph's dimming default.

JSXGraph's `text` default is `highlightStrokeOpacity: 0.666666` (`options.js:9286`), and `label.tsx:104-107` documents explicitly that this default must be overridden (jsxgraph issue #777).

After this change, hovering an x- or y-axis tick label fades it — and since the axis tick color was simultaneously moved from `--canvasText` to the lower-contrast `--graphAxes`, the faded label falls below the 4.5:1 target this PR is trying to establish.

packages/doenetml/src/Viewer/renderers/utils/jsxgraph.ts
```diff
@@ -270,12 +273,12 @@ export function createYAxis({
             yaxisOptions.ticks.scaleSymbol = scaleSymbol;
         }
     }
-    // Only control real ticks (height != -1), not grid lines (height = -1)
+    // Grid lines are rendered separately; these values control axis ticks.
     if (SVs.grid === "dense") {
-        yaxisOptions.ticks.majorHeight = -1;
-        yaxisOptions.ticks.minorHeight = -1;
+        yaxisOptions.ticks.majorHeight = 0;
+        yaxisOptions.ticks.minorHeight = 0;
```
dqnykamp, line 279: **LOW** — `createYAxis`/`createXAxis` disagree with `applyAxisTickHeights` for `grid === "dense"` (same at line 371).

Creation sets `minorHeight = 0` unconditionally, while `applyAxisTickHeights` sets `displayXAxisTicks ? 10 : 0`. Because `applyAxisTickHeights` is called at `useGridAndAxesSync.ts:121`, i.e. *before* the axis reconciliation that creates the axis, toggling `displayXAxis` from false to true on a `grid="dense"` graph produces an axis with no minor tick marks until some later render happens to re-run the effect.

The `medium` branch already uses `SVs.displayXAxisTicks ? 10 : 0` — `dense` should match.

packages/test-cypress/cypress/e2e/accessibility/graphVisualHierarchy.cy.js
```diff
@@ -0,0 +1,172 @@
...
+
+function composite(foreground, background, opacity) {
+    return parseRgb(foreground).map(
+        (component, index) =>
+            component * opacity + parseRgb(background)[index] * (1 - opacity),
+    );
+}
+
+describe("Graph visual hierarchy", { tags: ["@group5"] }, () => {
+    beforeEach(() => {
```
dqnykamp, line 75: **MEDIUM** — the spec posts its message without waiting for the harness.

`window.onmessage` is assigned during `CypressTest`'s render (`packages/test-cypress/src/CypressTest.tsx:173`), so it doesn't exist until React mounts. Every other spec in this directory guards first (see `darkModeRendererAccessibility.cy.js:28`):

```js
cy.get("#testRunner_toggleControls").should("exist");
```

Without it the `postMessage` can be dropped and the test flakes on `cy.get("#medium")` timing out.

packages/test-cypress/cypress/e2e/accessibility/graphVisualHierarchy.cy.js
```diff
@@ -0,0 +1,172 @@
...
+                ),
+            ).to.be.closeTo(3.1, 0.1);
+            expect(contrastRatio(axesColor, canvasColor)).to.be.greaterThan(
+                4.5,
+            );
+        });
+    }
+
+    for (const darkMode of ["light", "dark"]) {
+        it(`uses rendered colors with the required contrast in ${darkMode} mode`, () => {
```
dqnykamp, line 164: **MEDIUM** — the dark-mode case never verifies the theme actually switched, so it passes identically in light mode.

Every assertion compares the rendered stroke against `--graphGrid`/`--graphAxes` read from the same element, and computes contrast against that element's own background. Light-theme values satisfy the assertions just as dark-theme values do (#939393 on white → 3.07; #767676 on white → 4.54).

If dark-mode theming regressed completely — token overrides dropped, `data-theme` never applied — both `it` blocks would still pass.

Suggest adding `cy.get('[data-theme="dark"]').should("exist")` and asserting the expected per-theme token values.

packages/doenetml/src/Viewer/renderers/utils/labelMaskStyle.ts
```diff
@@ -6,9 +6,10 @@
  * element.
  */
 export interface LabelLikeJXG {
-    visProp: Record<string, any>;
+    visProp: any;
```
dqnykamp, line 9: **LOW** — unrelated type weakening plus dead code.

`visProp: Record<string, any>` → `any` and `update: () => void` → `Function` remove the little checking this interface provided. `hoverTargetJXG` was also widened to `| null` with a silent early return at line 150 — a caller that passes null now loses its hover highlight with no signal — and that makes the second `if (!hoverTargetJXG) return;` at line 161 unreachable.

No current caller (`point.tsx:249`, `line.tsx:215`, …) can pass null, and the commit message ("One more attempt … If this one fails I'll revert it back") suggests this was trial-and-error against a build error rather than a deliberate contract change.

DoctorHaitch:
Cool, I'll start addressing Claude's comments as soon as I get a chance once the semester starts!

PR #1769, opened 2026-08-29
Seed catalogs for eleven languages of Oceania

dqnykamp:
Seeds eleven unreviewed machine-generated catalogs for the languages of
Oceania, taking the roster from 245 locales to 256: Marshallese (`mh`),
Chuukese (`chk`), Pohnpeian (`pon`), Kosraean (`kos`) and Gilbertese (`gil`) in
Micronesia; Niuean (`niu`), Tokelauan (`tkl`), Tuvaluan (`tvl`), Rarotongan
(`rar`) and Wallisian (`wls`) in Polynesia; and Bislama (`bi`), the
English-lexified creole of Vanuatu.

It is the first batch assembled around **an ocean** rather than a family, a
script or a state, and the first whose members span nine countries and
territories with no shared administration between them.

## Four catalogs were seeded and deliberately dropped

Nauruan (`na`), Yapese (`yap`), Palauan (`pau`) and Rotuman (`rtm`) were seeded
as part of this batch and are **not** shipping in it. They reached only their
style tables: their `chrome.ftl`, `editor.ftl` and `diagnostics.ftl` stayed
byte-identical to English, translating 0 of 67, 0 of 64 and 0 of 220 messages
respectively. Shipping them would have put over 5,000 lines of English under
four locale directories, where the existing English fallback already puts
identical text on the screen.

This is a deliberate decision rather than a defect. The four are deferred to a
later batch with better lexical sources; every key, placeable and variant key
was already in place in them, so what they need is words rather than structure.
The README's confidence-scale section records that the tier existed and why it
was held back.

## The batch that needed nothing from the maps

`MACROLANGUAGE_MEMBERS` and `LANGUAGE_ALIASES` are **unchanged**, and after two
batches that each turned on them that is worth asserting rather than passing
over. The Caucasus batch had to keep `ckb` *out* of a list it had never been in;
the Uralic north had to take `koi` and `mrj` *out* of lists they were already
in. Here not one of the eleven is a macrolanguage, a member of one, or
previously folded onto a wider code — so no existing reader is sent anywhere
new. `negotiate.test.ts` pins that as a property: each of the eleven resolves to
itself against the full roster and to English against a roster of English alone,
which is the half that would fail if some entry were quietly folding one of
these tags.

Two of them (`mh`, `bi`) have ISO 639-1 codes, so the alpha-3 doors (`mah`,
`bis`) are pinned too — that folding is ICU's rather than this repository's.

## Not one member has CLDR plural data

The first batch of which that is true, and it inverts the Uralic north's
finding: there four Sami catalogs wrote a `[two]` branch because their own CLDR
rules select it and a fifth could not, and here the fifth case is the whole
batch. `Intl.PluralRules` resolves all eleven tags against the *runtime's*
default locale, so any category branch in these files would be text selected by
English's rules on English's terms.

The grammar agrees with the constraint — a noun in these languages is not
marked for number after a numeral — so a single unselected form is the right
translation as well as the safe one. `chrome.test.ts` holds all eleven to it,
and holds the other half too: each still renders its count, and each still
selects the explicit `[0]` branch, which is matched against the number rather
than against a category.

Pohnpeian and Gilbertese hit the same wall from another direction. Both count
with **numeral classifiers**, and in both the classifier is a suffix on the
numeral word — «uoua», «ruuw», never «2-ua» — so a count arriving as
`{ $count }` gives it nothing to attach to. That is the README's existing "an
affix cannot be welded to a placeable" rule reached through a classifier rather
than a case ending, and `chk` records it a third time.

## A declared confidence scale

The eleven catalogs are **not equally good, and each one says where it sits.**
Published lexical material for these languages runs from a modern dictionary
with a searchable web edition to a 1907 missionary wordlist that exists nowhere
as text, and the catalogs track that range instead of papering over it:

- **Translated** — `pon`, `mh`, `chk`, `niu`, `tkl`, `tvl`, `rar`, `wls`, `bi`
  write their own vocabulary throughout, on a published dictionary each header
  names.
- **Framed** — `kos` and `gil` write the catalog's frame in the language
  («Tia ku in…», «Wangin…», «ke sripen») around English technical nouns, and
  say so. `locales/kos` argues the case for leaving loans visible: `b c d g h j
  q v x z` are not Kosraean letters, so a loan should look like one rather than
  be respelled by invented loan phonology.

No word went in by guess, and `styleDescriptions.test.ts` pins the words that
are the language beside the ones that are not, so replacing a loan is a visible
diff rather than a silent improvement.

## A colour system that is not English's

`gil` is the batch's one colour-system mismatch: «mawaawa» covers green and blue
both, so it takes one key while the other keeps a loan, with the header saying
why rather than splitting the word silently.

## Word order, and a disagreement inside a subfamily

Ten put the adjectives behind the noun and one in front — `bi`, which follows
`locales/tpi`. A region is not a word order and here neither is a family. Two
catalogs write a linker out themselves: `chk`'s relative «mi» and `gil`'s «ae».

The sharper split is the side count: eight of the eleven — `chk`, `pon`, `kos`,
`niu`, `tvl`, `rar`, `wls`, `bi` — make it a following clause and reach
`[noun-tail]`, the widest use that variant has had in one batch. The other
three, `mh`/`gil`/`tkl`, fold it into the head by their own grammar. `tkl` is
the interesting one: it is Tuvaluan's closest relative in the batch and `tvl`
went the other way — a disagreement *inside* a subfamily, which no earlier
batch's had, and both files state their choice so a reviewer can see it is
deliberate.

A catalog reaching `[noun-tail]` has to place the tail the same way in
`style-with-noun` and `style-filled-with-noun`, or one polygon is assembled two
ways in one language depending only on whether it is filled. `locales/pon` and
`locales/kos` each said in their headers that the tail closes the phrase and
then placed it right after the noun in the filled message; both are corrected,
and `styleDescriptions.test.ts` now renders the filled phrase for all eleven so
the two messages cannot drift apart again.

## Pairs that are deliberately not copies

Five relationships here could each have produced a respelling of an existing
catalog, and all five are argued in a header instead. `niu` and `wls` both sit
beside `locales/to`: Niuean has no glottal-stop letter where Tongan writes the
fakauʻa, and Wallisian writes /ŋ/ as a single `g`, so "correcting" `g` to `ng`
would convert that file into Tongan. `rar` sits beside `mi` and `ty` with a
ten-row correspondence table and the instruction not to edit it back toward
Māori. `bi` sits beside `tpi` with fifteen rows and the rule stated outright
that resemblance is not evidence. And `tkl` and `tvl` each warn that the two
are expected to look alike and that **their agreement is not evidence either is
right**, since one process produced both — the self-suspicion the Uralic north
batch had to discover by auditing itself, built in here from the start.

`tkl` also names its own largest risk: Samoan `ʻ` corresponds to Tokelauan `k`
where it continues Proto-Polynesian \*k and to nothing where it does not, and
the seed cannot always tell which.

## Also here

- A `LOCALE_NAME_FALLBACKS` entry for `wls`, the only tag of the eleven CLDR
  has no name for in any language, its own included. «Fakaʻuvea» is copied
  letter for letter from the catalog's headers. ICU knows the other ten, down
  to `tkl` and `niu` — the best-named batch the roster has had.
- `map`, the ISO 639-5 collection over all Austronesian languages, left to miss
  for `smi`'s reason one family up: it maximizes to nothing at all. And
  recorded rather than fixed, `und-WF` maximizes to **French**, which is why
  `wls`'s loans are French-mediated where `to`'s are English-mediated — `rar`
  is the mirror image, taking English loans where `ty` takes French.
- Twelve near misses, sharper than the north's because the Pacific's language
  boundaries do not follow its political ones. `kpg` and `nkr` are *Polynesian*
  languages spoken inside the Federated States of Micronesia, so neither the
  country's catalogs nor the family's is right. `fud` (East Futunan) is this
  batch's `fkv` and sharper: it is spoken in the **same territory** as `wls`.
- The first catalogs to carry `math-embedded-input-blank`, its ordinal and
  `math-embedded-input-shape-unsuitable`, so each sits at 445/575 rather than
  442.
- Regenerated locale roster and schema, so `<document lang>` autocompletes all
  eleven.
- A README section covering the above.

## Honesty

Every string is machine-generated and unread by a speaker, and each catalog
says so in its own header. `chk` names «pwóón» — its word for *answer*, the
highest-frequency word in the catalog — as one it could not confirm, and `gil`
names its colour line as its least certain. Each catalog lists its own
coinages, and correcting any of it needs no permission.

## Testing

`lint:i18n` passes across 256 locales, every new catalog at 445/575.
`@doenet/i18n`, `@doenet/utils` and `@doenet/lsp-tools` all pass, including the
negotiation, plural-category and word-order assertions.

https://claude.ai/code/session_01EqQm2Ww4ATCb8fCud1kcqn


dqnykamp:
Note for anyone picking up a later Pacific batch: the four catalogs this PR
seeded and then deferred — `na` (Nauruan), `yap` (Yapese), `pau` (Palauan) and
`rtm` (Rotuman) — are not lost. Their finished files, including a pass that
recovered basic colour, width and shape vocabulary from Josephs, Churchward and
Jensen, are in this branch's history at **`2fdf23224`**:

```
git show 2fdf23224:packages/i18n/locales/pau/content.ftl
```

They are worth starting from rather than reseeding. Three things want attention
first, though:

- **Palauan's sourcing is unconfirmed.** The header claims each word is quoted
  from a tekinged entry. The words are genuine output of the research pass, but
  the verification pass that was supposed to check them never actually
  completed, so that claim has not been made good. Re-run it before trusting
  the header.
- **Nauruan's orthography question is open.** The header calls the tilde vowels
  and «ñ» the 1938 reform, while published accounts of that reform say it did
  the reverse. It governs the spelling of every «ng» in the catalog, and both
  colour words that went in use «ng».
- **Yapese is missing its underlined series** ḏ ḻ ṉ ṟ in nine words, because
  the web edition of Jensen used as a source prints none of them. Systematic
  and findable with the print edition.

The reason these four were deferred rather than shipped is in the PR
description above: each had translated only `content.ftl`, leaving chrome,
editor and diagnostics byte-identical to English, so the fallback would have
rendered the same screen.

Issue #1948, opened 2026-09-19
sort with sortByProp stops re-sorting once a value it sorted by changes


dqnykamp:
`<sort sortByProp="x">` sorts correctly when the document is built and then stops responding to the values it sorted by. Moving one of the points leaves the list in an order that matches neither the old values nor the new ones.

```xml
<graph name="g">
  <point name="A">(5,0)</point>
  <point name="B">(2,0)</point>
  <point name="C">(9,0)</point>
</graph>
<sort name="s" sortByProp="x">$A $B $C</sort>
<p>$s</p>
```

Drag A to 100, then C to -50:

| | `<sort>` | `<sort sortByComponent="1">` | `<sort sortByProp="x">` |
|---|---|---|---|
| initially | ( 2, 0 ), ( 5, 0 ), ( 9, 0 ) | ( 2, 0 ), ( 5, 0 ), ( 9, 0 ) | ( 2, 0 ), ( 5, 0 ), ( 9, 0 ) |
| A → 100 | ( 2, 0 ), ( 9, 0 ), ( 100, 0 ) | ( 2, 0 ), ( 9, 0 ), ( 100, 0 ) | **( 100, 0 ), ( 2, 0 ), ( 9, 0 )** |
| C → -50 | ( -50, 0 ), ( 2, 0 ), ( 100, 0 ) | ( -50, 0 ), ( 2, 0 ), ( 100, 0 ) | **( -50, 0 ), ( 100, 0 ), ( 2, 0 )** |

The two ways of sorting points by their first coordinate that do not name a prop both keep working. Only `sortByProp` goes wrong, and it goes wrong in a way that is worse than being stale: after the first move the order is the document order, and after the second it is neither the old sorted order nor the new one.

It is specifically a prop that *changes* that is not tracked. The existing `sort by prop` test sorts by `numDimensions` and updates correctly when a `<mathInput>` changes a point's coordinates, so the dependency works in that shape — the suspicion is the prop values that feed `listValues` when the prop belongs to a component whose own state variable changed, rather than the child being replaced.

Reproduces on `main` (checked at ac3b66624, with no other changes in the tree). Found while testing #1947's first PR, which does not touch this path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1947, opened 2026-09-19
Composites recreate all their replacements when they could rearrange or withhold the ones they have


dqnykamp:
Several composites throw away and rebuild all of their replacements in situations where the components they need already exist. The cost is real: `<sort>` over a 40-entry list adds ~75–90 ms to every drag that changes the order, against ~2 ms for the sibling operators that update in place, and whatever reads its output re-resolves 40 fresh component indices on top of that.

The individual cases look unrelated, but they share one cause. The core's vocabulary for replacement change is positional and suffix-shaped:

- `replacementsToWithhold` is a **count from the end** (`adjustReplacementsToWithhold` slices `replacements.length - replacementsToWithhold`), so only a suffix can be withheld;
- `add` replaces a **contiguous run** with freshly serialized components;
- `delete` removes a **contiguous run**.

There is no way for a composite to say *"keep these components I already have, in this order."* Every composite that needs to says it with a full rebuild instead, and two of them have a TODO on the line where they give up:

```js
// delete previous replacements and create new ones
// TODO: could we find a way to withhold old ones?
// Either change order of replacements or allow to withhold later replacements
```
— `ConditionalContent.js`

```js
// check if changed type
// or have excluded elements
// TODO: don't completely recreate if have excluded elements
```
— `Sequence.js`, and identically in `RepeatForSequence.js`

## Survey

Every composite defining `calculateReplacementChanges`, by what it does when its inputs change.

**Recreate everything, but the components it needs are a permutation of the ones it has**

| composite | trigger |
|---|---|
| `<sort>` | any change in sorted order (`// for now, just recreated`) |
| `<shuffle>` | any change in shuffled order; same code shape as `<sort>` |
| `<repeat>` | any position where `sourcesChildIndices` differs — so reordering the source list rebuilds every iteration |
| `<numberList>` `<mathList>` `<textList>` `<booleanList>` `<pointList>` `<vectorList>` `<intervalList>` | any position where the copied source list differs |

**Recreate everything because it cannot withhold anything but a suffix**

| composite | trigger |
|---|---|
| `<conditionalContent>` | any change of selected case, unless there is exactly one case (the single case is the only shape where "withhold everything" is a suffix) |
| `<sequence>` `<repeatForSequence>` | any change at all once `exclude` is non-empty |
| `<collect>` | a collected component disappearing — deletes and recreates everything after it, "since deleting replacement will shift the remaining replacements and change resulting names" |

[...]

Issue #1944, opened 2026-09-19
Saved state is keyed by a component index that varies between builds, so a reload can restore values into the wrong components


dqnykamp:
Saved state is keyed by a component index that is not stable between builds of the same document, so restoring it can put a reader's values into the wrong components. The visible symptom is a point coming back with its coordinates swapped, but the cause is general and the symptom is silent corruption, not an error.

Every figure below was measured on `main` (9331461a0) through `createTestCore`, rather than inferred from the code.

## The sharpest reproduction

Saving is not involved. Take **one fixed saved-state string** and load the same document from it twenty times:

```js
const DOC = `<graph><point name="P" x="0" y="1" /></graph>`;
const SAVED = JSON.stringify({
    "4": { expressionWithCodes: 3,  __def_primitive_ignore_0: "3" },
    "5": { expressionWithCodes: -5, __def_primitive_ignore_0: "-5" },
});
```

> `{"(3,-5)": 14, "(-5,3)": 6}`

The same bytes restore to two different points. In author-visible terms: `<point x="0" y="1"/>` in a `<graph>`, dragged to (3, −5), saved and reloaded, comes back at (−5, 3) roughly a third of the time. Ten identical drag → save → reload runs gave 6 transposed and 4 correct.

Because it is a race, **a single observation tells you almost nothing**. Three separate investigations of this bug each proposed a confident mechanism from one run — "it depends on the dragged value", "it depends on whether the point is constrained", "it depends on the constraint target" — and all three were wrong.

## Cause

1. `DastElement.attributes` is a `HashMap<String, DastAttribute>` (`dast_structure.rs:86`, and the same at `:284` and `:493`). Rust's `HashMap` deliberately randomizes iteration order.
2. **Measured at the JS boundary:** instrumenting `ComponentBuilder` to log the attribute key order it receives, ten builds of the identical document gave five different orders — `["name","x","y"]` ×5, `["x","y","name"]` ×2, and one each of `["name","y","x"]`, `["x","name","y"]`, `["y","name","x"]`.
3. `ComponentBuilder.ts:486` — `for (const attrName in serializedComponent.attributes)` — creates one component per attribute in whatever order it got, so `x` and `y` are assigned different `componentIdx` values from build to build.
4. `stateId` defaults to that index: `const stateId = serializedComponent.stateId ?? componentIdx.toString()` (`ComponentBuilder.ts:671`).
[...]

PR #1949, opened 2026-09-19
Rearrange sorted replacements rather than rebuilding them

dqnykamp:
First step of #1947. A composite could tell the core to add, delete or withhold replacements, but never to keep the ones it has in a different order. `<sort>` needs exactly that — any value moving past another reorders its output — so it took the only route available and rebuilt every replacement, leaving everything downstream to resolve fresh component indices.

## What this adds

A `rearrangeReplacements` change type. Its payload is the new arrangement of the composite's existing active replacements: entry `k` says which of them belongs at position `k` now. The handler reorders `component.replacements`, reports the new ordered content to the resolver so an index into the composite still lands where the author meant, adds the usual blockers, and passes the same arrangement down to shadows, which hold their replacement arrays one-for-one.

`<sort>` takes that route whenever the components it is copying are the same ones in a new order, and recreates as before when the children themselves changed. A component appearing more than once — `<sort>$a $a</sort>` — is matched off in order, so the arrangement stays a permutation.

`calculateReplacementTypesFromChanges` in `utils/copy.js` permutes its running list of replacement types the same way, so the function stays complete over the change types a composite can emit. Nothing reaches that branch yet: `<sort>` is the only source of the change, and it does not run its changes through `verifyReplacementsMatchSpecifiedType`. A mixed-type sort keeps its types today through the shadow recursion in the handler, which is covered by a test.

## What it deliberately does not do

The arrangement has to be a permutation of *all* the active replacements. Withholding an arbitrary subset is the other half of #1947 and needs a deeper change: `replacementsToWithhold` is a count from the end of the array, so a non-suffix subset cannot be expressed at all. `<conditionalContent>`, and `<sequence>` / `<repeatForSequence>` with excludes, keep recreating until that lands. The payload is shaped as an arrangement rather than a swap so that work does not need a second format.

## Effect

Median cost of one drag, over 25 drags of one point in a document with 40 draggable
points, measured under vitest in node (several times slower than the browser worker —
the ratios are the point). `before` is `main` at ac3b66624; `after` is this branch.

| document | before | after |
|---|---|---|
| 40 points, a `<numberList>` of their `x`, no `<sort>` | 2.8 ms | 2.7 ms |
| `<sort>` of the 40 `x` values, nothing reading it | 119 ms | 65 ms |
| `<sort>` of the 40 points themselves, nothing reading it | 169 ms | 46 ms |
| `<sort>` of the 40 `x` values, read by a `<searchSorted>` of 40 targets | 225 ms | 144 ms |

Every drag above moves the point past others, so the sort reorders each time; a drag
that leaves the order unchanged returns early in `calculateReplacementChanges` before
either path and costs the same.

Comparing the two documents that sort the same 40 numbers, more is saved in absolute
terms when something reads the sorted list — 81 ms against 54 ms — since that is the
part that no longer re-resolves: instrumenting `processNewDefiningChildren` counts 42 calls
per drag in that document before and 19 after. What remains is not the rebuilding. The
rearrangement itself is 0.6–1.5 ms per drag, and `processNewDefiningChildren`, where a
parent re-derives all of its children, costs 30–60 ms per drag on *both* paths. That is
the next lever — noted on #1947.

The replacement components are what survive: traced across four reorders of
`<sort>5 $m 1</sort>`, the sort's own replacements permute within a fixed set of three
indices, where on `main` all three are replaced every time. A composite that copies the
output, such as `<p>$s</p>` or `$s[2]`, still builds replacements of its own and still
rebuilds some of them on each reorder — it does less of it, not none.

## Testing

`npm run test -w @doenet/doenetml-worker-javascript -- src/test/tagSpecific/sort.test.ts` and the neighbouring composite suites (`shuffle`, `conditionalcontent`, `collect`, `repeat`, `sequence`, `numberlist`, `listoperators`, `lists`, `src/test/copying`) pass.

New tests in `sort.test.ts` cover a value moving past another (asserting the replacement components survive, by their indices), a repeated child, a changed set of children still recreating, a copy of the sort following along, an index into the sort resolving to the new position, a prop reference following the reorder, and sorted points staying live afterwards. The degenerate cases have their own tests: a sorted list that empties and refills, children whose values do not parse as numbers, and a sort whose replacements are of more than one component type.

While writing those I found a pre-existing bug, unrelated to this change and reproducing on `main`: `<sort sortByProp="x">` does not re-sort when a point moves, though `<sort>` and `<sort sortByComponent="1">` both do. Filed separately as #1948.

## Known issue: saved state can be restored to the wrong position

Not yet fixed. Reproduced independently twice: against a control that turns only the
new path off, and against `main` at ac3b66624 directly.

A `<sort>` replacement's `stateId` is its `componentIdx` (`ComponentBuilder.ts` falls
back to `componentIdx.toString()` when a serialized replacement carries no explicit
`stateId`), and saved reader state is keyed by `stateId`. On the old path the
replacements were deleted on every reorder, so `DeletionEngine` dropped their saved
entries and the rebuilt ones started clean. Keeping the components alive keeps those
entries alive too — under a `componentIdx` that a fresh load reassigns in sorted-position
order, while at save time it was in original-creation order.

So where a replacement owns its value essentially, and the reader writes into a sorted
position in a way that reorders the list, a reload puts that value on the wrong
replacement:

```xml
<sort name="s">5 3 1</sort>
<p name="pList">$s</p>
<mathInput name="mi" bindValueTo="$s[1]" />
```

Typing `7` gives `3, 5, 7`; reloading from the saved state gives `7, 5, 7`. With
`bindValueTo="$s[2]"` and `0`, `0, 1, 5` reloads as `0, 0, 5`. A third: with
`<sort name="s">2 4 6 8</sort>` and `bindValueTo="$s[4]"`, typing `1` gives
`1, 2, 4, 6` live and `1, 2, 4, 1` on reload. All three restore correctly on `main`.
The saved payload shows the extra key directly — four entries on this branch where
`main` saves three, the fourth being the orphaned replacement.

It needs the replacements to hold the value themselves: with named `<number>` or
`<math>` children the write lands on the child and both paths reload correctly, as do a
reorder driven by an input outside the `<sort>` and a drag of a sorted point shown in a
`<graph>`, which records nothing on the replacement. A fix belongs in how a rearranged
replacement's persistence key is chosen — deriving it from the source being copied
rather than from creation order — and is not attempted here.

Refs #1947.


dqnykamp:
Converted to a draft and parked, on the user's decision, until saved-state keying is settled (#1944).

The reason is the "Known issue" section below: keeping replacements alive across a reorder makes saved reader state restore onto the wrong replacement. It was found by the second review cycle and independently reproduced by the third, with a second reproducer built from scratch, each run against a control with the three changed files checked out from `upstream/main` in place. The mechanism and the saved-payload evidence are recorded on #1944, and what it implies for the rest of the plan is on #1947.

The code here is reviewed and the branch is in a state to be picked up: three review cycles plus a closing read-only pass, covering correctness and edge cases, behavior delta, and claims against code. The timing claims were re-measured from scratch in cycle 3 and were understated rather than overstated.

Nothing further should be done on this branch until a replacement identity that survives reuse exists. At that point the regression should be re-tested rather than assumed fixed — the reproducers are in the "Known issue" section.

dqnykamp:
Rebased onto `542bb3e76` and taken out of draft. CI is green apart from `Test Worker JS (24.x, group4)`, which fails these four and nothing else:

```
sort.test.ts > a value typed into a sorted position survives a reload
    expected [ '7', '5', '7' ] to deeply equal [ '3', '5', '7' ]
sort.test.ts > a value typed into the second sorted position survives a reload
    expected [ '0', '0', '5' ] to deeply equal [ '0', '1', '5' ]
sort.test.ts > a value typed into the last sorted position survives a reload
    expected [ '1', '2', '4', '1' ] to deeply equal [ '1', '2', '4', '6' ]
statePersistenceContents.test.ts > reloads a `<sort>` showing what the reader was looking at
    expected { shown: '7', value: '7', … } to deeply equal { shown: '3', value: '3', … }
```

All four are the shadow-state duplicate that #1974 removes — they pass on `main`, fail here, and pass again with #1974 merged in locally. **Not to be merged before #1974.** Nothing else in the run is red: 79 files pass in that job, and every Cypress group, `Test Main`, the schema check and the docs build are green.

